PR #2, opened 2023-04-17
Make sure ID is ASCII

duglin:
Please only review the 2nd commit.

ping @jskeet

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -146,7 +146,7 @@ TODO: Add `format`
 - Type: String   # SHOULD this be a URI-Reference?
 - Description: An immutable unique identifier of the entity.
 - Constraints:
-  - MUST be a non-empty string
+  - MUST be a non-empty string consisting of visible US-ASCII octets (33-126).
   - MUST be immutable
   - MUST be case-insensitive unique within the scope of the entity's parent.
     In the case of the `id` for the Registry itself, the uniqueness scope will


core/spec.md
```diff
@@ -146,7 +146,8 @@ TODO: Add `format`
 - Type: String   # SHOULD this be a URI-Reference?
 - Description: An immutable unique identifier of the entity.
 - Constraints:
-  - MUST be a non-empty string
+  - MUST be a non-empty string that is compliant with the definition of
+    an HTTP header [`field-name`][field-name]`.
```
jskeet, line 150: I believe this prohibits the use of `/` and `:` within an ID (as well as various other characters, of course). I can see those being pretty common within IDs... do we definitely want to prohibit them?

duglin: oh wait - you're right, I messed up. I need to worry about valid http header values, not names. Doi! Will fix

core/spec.md
```diff
@@ -146,7 +146,7 @@ TODO: Add `format`
 - Type: String   # SHOULD this be a URI-Reference?
 - Description: An immutable unique identifier of the entity.
 - Constraints:
-  - MUST be a non-empty string
+  - MUST be a non-empty string consisting of visible US-ASCII octets (33-126).
```
jskeet, line 149: Assume we genuinely want all of those characters to be valid, that looks fine to me. I'm not in a position to judge whether we *do* want that character set :)

duglin: I'm inclined to start with being less restrictive and wait until someone can identify a reason for us to ban more chars.

duglin:
Approved on the 4/27 call

PR #30, opened 2023-05-02
Add format attributes, expand in inlining and more clean-up

duglin:
- Moved inlining and filtering to their own sections
- Add a `format` attribute
- started the pattern of specifying the exact attributes on each resource instead of just doing "spec by example"
- added `specVersion` attribute
- a lot of general clean-up (and add more clarity) as we weren't as "spec-ish" as we need to be


duglin:
Approved on the 4/4 call

PR #34, opened 2023-05-04
Remove 'core' from endpoint/message/schema specs

duglin:
I left the "CE Registry" stuff in the "endpoint" spec for now until we figure out where to put it. The main purpose of this PR is to get rid of the "core" stuff in each spec so people don't get confused.


duglin:
Approved on the 4/4 call

PR #45, opened 2023-07-11
make filtering an entity clearer

duglin:


core/spec.md
```diff
@@ -4062,6 +4062,11 @@ MUST apply for an entity to be considered a match of the filter expression:
   (`true` or `false`). When `VALUE` is not present then the attribute is a
   match is its value is `true`
 
+If the request references an entity (not a collection), and the `EXPRESSION`
```
jskeet, line 4065: It's a shame that 412 (precondition failed) is so narrowly defined. I'd personally prefer the filter to be seen as a precondition, so that clients could distinguish between "resource doesn't exist" (404) and "resource exists but doesn't match the filter" (412). But a strict reading of the HTTP spec doesn't meet that :(

duglin: On the last call we discussed several options, including creating a new query parameter to mean a "precondition" type of thing.  I've been thinking about this and I really think creating a new query parameter just for this case (instead of reusing `filter`) is overkill and a bit too pedantic for our needs. I think it'll just annoy our users.

I was checking the 404 description and rfc2616 starts with:
```
The server has not found anything matching the Request-URI.
```
TBH I think that covers what's happening here. The root entity of the GET doesn't "match" the "Request-URI" since the "Request-URI" includes the query parameters. So I'd like to propose that we keep the PR as is.

jskeet: Yup - any comments I have on xRegistry are just suggestions. I'd have to get fully back into the context to say whether or not I completely agree with the reasoning here, but it certainly doesn't seem egregious :)

duglin: I'll take "not egregious" for now :-)
Thanks so much for all the reviews.

duglin:
approved on the 7/27 call

PR #46, opened 2023-07-11
limit attr names

duglin:


jskeet (approved):
Just to check, `_xyz` is intended okay?

duglin:
> Just to check, `_xyz` is intended okay?

yep - unless someone things we need to ban it for some reason.

PR #50, opened 2023-07-17
Allow for a client to specify version of xReg spec

duglin:


duglin:
Approved on the 7/20 call

PR #52, opened 2023-07-26
Week of July 25 edits

duglin:
- Updated the PUT/POST stuff for Resource/Versions
- s/tags/labels/
- serialize labels as individual `xRegistry-labels-NAME` headers, not JSON
- use `/` instead of `.` for filter and inline PATHs

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -15,12 +15,12 @@ automation and tooling.
 - [Registry Formats and APIs](#registry-formats-and-apis)
   - [Attributes and Extensions](#attributes-and-extensions)
   - [Registry APIs](#registry-apis)
-    - [Registry Model](#registry-model)
-    - [Registry Collections](#registry-collections)
-    - [Registry Entity](#registry-entity)
-    - [Groups](#groups)
-    - [Resources](#resources)
-    - [Versions](#versions)
+  - [Registry Model](#registry-model)
+  - [Registry Collections](#registry-collections)
+  - [Registry Entity](#registry-entity)
+  - [Groups](#groups)
+  - [Resources](#resources)
+  - [Versions](#versions)
   - [Inlining](#inlining)
   - [Filtering](#filtering)
 
@@ -48,7 +48,7 @@ For easy reference, the serialization of a Registry adheres to this form:
   "self": "URL",
   "description": "STRING", ?
   "documentation": "URL", ?
-  "tags": { "STRING": "STRING" * }, ?
+  "labels": { "STRING": "STRING" * }, ?
 
   "model": {                            # only if inlined
     "schema": "URI-Reference", ?        # schema doc for the entire Registry
@@ -61,8 +61,8 @@ For easy reference, the serialization of a Registry adheres to this form:
           { "singular": "STRING",       # eg. "definition"
             "plural": "STRING",         # eg. "definitions"
             "versions": UINT ?          # num Versions(>=0). Def=1, 0=unlimited
-            "versionId": BOOL, ?         # Supports client specific Version IDs
-            "latest": BOOL ?             # Supports client "latest" selection
+            "versionId": BOOL, ?        # Supports client specific Version IDs
+            "latest": BOOL ?            # Supports client "latest" selection
           } *
         ] ?
       } *
@@ -80,7 +80,7 @@ For easy reference, the serialization of a Registry adheres to this form:
       "self": "URL",
       "description": "STRING", ?
       "documentation": "URL", ?
-      "tags": { "STRING": "STRING" * }, ?
+      "labels": { "STRING": "STRING" * }, ?
       "format": "STRING", ?
       "createdBy": "STRING", ?
       "createdOn": "TIME", ?
@@ -100,7 +100,7 @@ For easy reference, the serialization of a Registry adheres to this form:
           "latestUrl": "URL",
           "description": "STRING", ?
           "documentation": "URL", ?
-          "tags": { "STRING": "STRING" * }, ?
+          "labels": { "STRING": "STRING" * }, ?
           "format": "STRING", ?
           "createdBy": "STRING", ?
           "createdOn": "TIME", ?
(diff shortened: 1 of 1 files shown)


core/spec.md
```diff
@@ -247,7 +247,7 @@ adhere to the following rules:
   treated the same as an attribute with no value.
 - the string serialization of the attribute name and its value MUST NOT exceed
   4096 bytes. This is to ensure that it can appear in an HTTP header without
-  exceeding implemantion limits (see
+  exceeding implemention limits (see
```
jskeet, line 250: Maybe you'll get lucky third time? ;)

core/spec.md
```diff
@@ -379,10 +379,19 @@ existing entity can be deleted.
     start with an alphanumeric character and be unique within the scope of
     this map
   - Values MAY be empty strings
-  - When serialized as a string (e.g. in an HTTP header), the serialization
-    SHOULD be in JSON without line-breaks
+  - When serialized as an HTTP header, each "name" MUST appear as a separate
+    HTTP header prefixed with `xRegistry-labels-` and the header value
+	MUST be the label's "value".
```
jskeet, line 384: Any reason for this four-space indent?

core/spec.md
```diff
@@ -1713,39 +1709,43 @@ retrieve previous Versions of the Resource. In this respect, Resources have
 a 2-layered definition. The first layer is the Registry Resource entity itself,
 while the second layer is the `versions` collection - the version history of
 the Resource. The Resource entity can be thought of as an alias for the
-"latest" Version of the Resource, and as such, many of the attributes
-shown when retrieving the Resource will come from that Version. However,
-there are a few exceptions:
+"latest" Version of the Resource, and as such, most of the attributes
```
jskeet, line 1712: Having removed "shown", we're missing a verb: "most of the attributes ??? when processing the resource".

core/spec.md
```diff
@@ -1774,6 +1774,40 @@ When serialized as a JSON object, a Resource adheres to this form:
...
+xRegistry-versionsCount: INT
+xRegistry-RESOURCEUrl: URL ?
+Location: URL
+Content-Location: URL ?
+
+...Resource contents... ?
+```
+
+Notes:
+- the `versionsUrl` and `versionsCount` attributes are includes, but not
```
jskeet, line 1806: includes => included?

core/spec.md
```diff
@@ -1774,6 +1774,40 @@ When serialized as a JSON object, a Resource adheres to this form:
...
+Content-Type: ... ?
+xRegistry-id: STRING
+xRegistry-name: STRING ?
+xRegistry-epoch: UINT
+xRegistry-self: URL
+xRegistry-latestId: STRING
+xRegistry-latestUrl: URL
+xRegistry-description: STRING ?
+xRegistry-documentation: STRING ?
+xRegistry-labels-KEY: STRING *
```
jskeet, line 1790: Perhaps show

```text
xRegistry-labels-KEY1: STRING ?
xRegistry-labels-KEY2: STRING ?
xRegistry-labels-...: STRING ?
```

I suspect that would be clearer than using the `*` here.

core/spec.md
```diff
@@ -1962,53 +2063,65 @@ Link: <https://example.com/endpoints/123/definitions&page=2>;rel=next;count=100
...
+
+To create a new Resource, there are 5 different HTTP actions that MAY be used:
+- `POST /GROUPs/gID/RESOURCEs`
+- `PUT  /GROUPs/gID/RESOURCEs/rID`
+- `POST /GROUPs/gID/RESOURCEs/rID`
+- `POST /GROUPs/gID/RESOURCEs/rID/versions`
+- `PUT  /GROUPs/gID/RESOURCEs/rID/versions/vID`
 
-To create a new Resource, an HTTP `POST` MAY be used.
+All five follow the the same overall semantics rules so they are described
```
jskeet, line 2075: semantics rules => semantic rules?

(I tend to use semantics as a noun, but semantic as an adjective... not sure if that's "correct" or just personal habit.)

core/spec.md
```diff
@@ -2017,7 +2130,7 @@ xRegistry-latestId: STRING
 xRegistry-latestUrl: URL
 xRegistry-description: STRING ?
 xRegistry-documentation: STRING ?
-xRegistry-tags: STRING (JSON map) ?
+xRegistry-labels-KEY: STRING *
```
jskeet, line 2133: Same comment as earlier (either ignore the suggestion in both places, or change both)

core/spec.md
jskeet: (This occurs quite a few times.)

core/spec.md
jskeet: Looks like this happens in other places as well...

core/spec.md
```diff
@@ -2446,25 +2511,22 @@ Where:
...
+- complex attributes that have nested values (eg. `labels`) MUST be specified
   in their entirety
 
-Missing Registry HTTP headers MUST be interpreted as a request to delete the
-attribute.
-
-Note: `latestId` is a read-only attribute in this context and therefore
-if present MUST be silently ignored.
+Missing Registry HTTP headers MUST be interpreted as a request to leave the
+corresponding Version attribute unchanged. An attribute with a value of `null`
```
jskeet, line 2518:
```suggestion
corresponding attribute unchanged in the new Version. An attribute with a value of `null`
```

(Otherwise it sounds like it's about an attribute called Version.)

jskeet (commented):
This is only a *very* surface level review; it doesn't attempt to comment on any changes in semantics etc. (I've only had one coffee so far today, and I still haven't taken the time to try to fully understand xRegistry...)

core/spec.md
duglin: I find my typing is getting worse as I get older ;-)  
Thanks for the review

core/spec.md
duglin: Nope - a tab slipped in there. I'll remove them.

core/spec.md
duglin: fixed

core/spec.md
duglin: got 'em!

core/spec.md
duglin: tweaked them all

duglin:
Approved on the 7/27 call

PR #53, opened 2023-07-28
Edits for week of Aug 1

duglin:
Just trying to help the overall flow the doc. Still not complete and still too long, but I think this helps a little.
- Mainly just moving stuff around but I'll add some comments for any new text.
- add some pictures to try to help people grok the model
- added `epoch` to the Registry root entity
- did a major rewrite of the create Resources/Versions sections trying to make it easier to follow and changed it so that the "Creating a Version" section just pointed to the "Creating a Resource" section since it was a dup
- removed the ability to set "latestId" of a Resource via a PUT to the Version. Now you can only do it via a PUT to the Resource with ?latestId=vID.  We kept the ability to specify ?latest=true|false on Version.create() and ?latestId=vID on Version.Delete().

Deleting the dup text should shrink the spec quite a bit, which is good.


duglin:
Approved on the 8/3 call

PR #54, opened 2023-08-09
Edits for week of Aug 7

duglin:
- remove tabs
- minor typos
- extensions can't use spec defined attribute names, even for different entities
- make it clear that it's `labels/name`, not `labels.name` to reference a particular label, for filtering
- s/latestId/latestVersionId/
- s/latestUrl/latestVersionUrl/g
- add PUT /GROUPs/gID for create
- ensure `id` in body on PUT GROUP matches gID in URL


duglin:
Approved on 8/10 call

PR #56, opened 2023-09-08
Week of Sept 6th updates

duglin:
- Allow for bulk create/update of Groups, Resources and Versions
  - Merged "update" and "create" into one section for each
  - Creating/Updating Versions just points to the Resources section
- cleanup/shorten the samples a bit
- remove the ?latest query parameter and made people use the `latest` attribute instead
- lots of typos and clean-up
- make bulk DELETE's response an array instead of a map


duglin:
Approved on the 9/14 call

PR #59, opened 2023-10-18
Week of Oct16 edits

duglin:
- added more TODO discussion items to the primer (else we'll forget)


duglin:
Approved on the 10/19 call

PR #61, opened 2023-11-06
Week of Nov 6 edits

duglin:
- non-json serialization reminder for Primer
- made model.groups and resources a map instead of array
- add 'map' as a TYPE and added a `keyType` to the attribute definition
- s/parentAttributes/siblingAttributes/
- only support filtering on entity root scalar attributes + maps (e.g. labels)

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -41,7 +41,7 @@ automation and tooling.
 A Registry Service is one that manages metadata about Resources. At its core,
 the management of an individual Resource is simply a REST-based interface for
 creating, modifying and deleting the Resource. However, many Resource models
-share a common pattern of grouping Resources (eg. by their "format") and can
+share a common pattern of grouping Resources (e.g. by their "format") and can
 optionally support versioning of the Resources. This specification aims to
 provide a common interaction pattern for these types of services with the goal
 of providing an interoperable framework that will enable common tooling and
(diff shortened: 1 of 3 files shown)


core/spec.md
```diff
@@ -91,51 +91,55 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
...
+            "siblingAttributes": { ... } # See "attributes" above
           } *
         } ?
       } *
     },
 
-    "groups": [
-      { "singular": "STRING",           # eg. "endpoint"
+    "groups": {
+      "STRING": {                       # Plural name, eg. "endpoints"
```
jskeet, line 117: Should this be on the next line instead?

core/spec.md
```diff
@@ -91,51 +91,55 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
...
         "plural": "STRING",             # eg. "endpoints"
-        "attributes": { ... }, ?        # see "attributes" above
+        "singular": "STRING",           # eg. "endpoint"
```
jskeet, line 119: If we're editing this a lot anyway, let's correct "eg." to "e.g." everywhere :)

core/spec.md
```diff
@@ -91,51 +91,55 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
...
+        "attributes": { ... }, ?        # See "attributes" above
 
-        "resources": [
-          { "singular": "STRING",       # eg. "definition"
+        "resources": {
+          "STRING": {                   # Plural name, e.g. "definitions"
```
jskeet, line 123: Again, I think this should be in the next line.

core/spec.md
```diff
@@ -91,51 +91,55 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
...
             "plural": "STRING",         # eg. "definitions"
-            "versions": UINT ?          # num Versions(>=0). Def=1, 0=unlimited
-            "versionId": BOOL, ?        # Supports client specific Version IDs
+            "singular": "STRING",       # eg. "definition"
+            "versions": UINT ?          # Num Versions(>=0). Def=1, 0=unlimited
```
jskeet, line 126: What does "Def" mean here?

core/spec.md
```diff
@@ -91,51 +91,55 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
...
   } ?
 
   # Repeat for each Group type
   "GROUPsUrl": "URL",                              # eg. "endpointsUrl"
   "GROUPsCount": INT                               # eg. "endpointsCount"
-  "GROUPs": {                                      # only if inlined
-    "ID": {                                        # the Group id
-      "id": "STRING",                              # a Group
+  "GROUPs": {                                      # Only if inlined
+    "ID": {                                        # The Group id
```
jskeet, line 141: This comment and the next line look like they're the wrong way round. Or at least odd.

core/spec.md
```diff
@@ -298,15 +302,23 @@ exchange.
...
 
 - they MUST be defined as part of the [Registry Model](#registry-model)
 - the presence of an undefined attribute in a request MUST generate an error
 - they MUST NOT use the name of an attribute defined by this specification,
-  regardless of which entity the attribute is defined for
+  regardless of which entity the attribute is defined for. In other words,
+  a spec defined attribute for GROUPs can not be reused as an extension for
+  a RESOURCE
+- it is RECOMMENDED that extension attributes on different objects not use the
+  same name unless they have the exact same semantics meaning
```
jskeet, line 321:
```suggestion
  same name unless they have the exact same semantic meaning
```

core/spec.md
```diff
@@ -1021,41 +1036,45 @@ Regardless of how the model is retrieved, the overall format is as follows:
...
-          "parentAttributes": { ... }  # Siblings to this "attribute"
+          "siblingAttributes": { ... } # Siblings to this "attribute"
         } *
       }
     } *
   },
-  "groups": [
-    { "singular": "STRING",            # eg. "endpoint"
+  "groups": {
+    "STRING": {                        # Plural name, eg. "endpoints"
```
jskeet, line 1060: Again with the slightly odd location of comments. I'm beginning to think I just don't understand.

core/spec.md
```diff
@@ -1087,6 +1106,11 @@ The following describes the attributes of Registry model:
 
 TODO: check the above example list
 
+- `attributes."STRING".name
+  - The name of the attribute. MUST be the same as the key used in the owning
+    `attributes` attribute.
+  - Type: Sring
```
jskeet, line 1112:
```suggestion
  - Type: String
```

core/spec.md
```diff
@@ -3670,6 +3725,16 @@ references an attribute in that entity (i.e. there is no `PATH`), then if the
 `EXPRESSION` does not match the entity, that entity MUST NOT be returned. In
 other words, a `404 Not Found` would be generated in the HTTP protocol case.
 
+At this time, this specification only supports filtering over scalar attributes
+defined at the root of the xRegistry entities (the Registry itself, Groups,
+Resources and Versions). The one exception to this are maps (such as `labels`)
```
jskeet, line 3730: The singular/plural clash here is bugging me.
```suggestion
Resources and Versions). The one exception to this is support for filtering on maps (such as `labels`)
```

jskeet (commented):
Very shallow review, as you'd no doubt expect from me on anything xRegistry...

core/spec.md
duglin: I wanted to make it clear that they key was the plural name, not the singular one.
I'll see `key=` to the from of the comment to make that clearer.

core/spec.md
duglin: was trying to save space but let me see if anything wraps.... ok we're good.

core/spec.md
duglin: "Default" - I added the "ault" w/o wrapping :-)

core/spec.md
duglin: done

core/spec.md
duglin: done

duglin:
@jskeet as always, thanks for the great review

core/spec.md
jskeet: I think that would be clearer, yes. My confusion is that the name line really *is* the plural name. (Is it an error if the key doesn't match the value of `plural`? And do we note that?)

core/spec.md
jskeet: Massively, massively - it's somewhat shocking just how much clearer that makes it, for me anyway :)

core/spec.md
duglin: Good great - I forgot to change it from Array to Map and I added the MUST for both Groups and Resources:
```
-  - Type: Array
+  - Type: Map where the key MUST be the plural name of the Group
```

PR #62, opened 2023-11-14
Fix nested resource definition in the model

duglin:
Also require map keys be scalar


duglin:
Approved on the 11/30 call

PR #64, opened 2023-12-05
Edits for week of Dec 4

duglin:
- updates to our TODO list for the primer based on these changes
- lots of minor typos/wording tweaks
- lowercase all attributes and map key names
- INT->INTEGER - missed from previous PR
- Add support for `*` (undefined) extensions to the model
- Add support for an `ANY` type
- made a few spots UINTEGER instead of INT (e.g. counts)
- add support for dots in attribute names: `prop1['my.name'].prop2`
- deleting the last Version deletes the Resource
- unknown extension types SHOULD default to `string`. E.g. `xRegistry-foo: 5`   http header
- tweaked what OPTIONAL means... @jskeet should be happy
  - For clarity, OPTIONAL attributes (specification defined and extensions) are OPTIONAL for clients to use, but servers MUST be prepared for them to appear in incoming requests and MUST support them since "support" simply means persisting them in the backing datastore. 
- can no longer modify the constraints on spec defined attributes (e.g. require string to be a timestamp)
- ?schema= returns the full model, not just extension definitions
- removed KeyType from model since all keys are strings now

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -3,17 +3,12 @@
 Possible future topics to cover:
 - Resource.ID and Resource.Version.ID might have the same values but they are
   very different properties with no relation to each other.
-- we require all extensions (and therefore all attributes) to be defined by
-  the "model" so that we know their true serialization (case) even if we see
-  them in a case insensitive serialization (e.g. HTTP headers)
-- we use xxx/yyy/zzz for filters instead of xxx.yyy.zzz so that people have
-  a consistent pattern when traversing the registry try. Meaning, it would
-  be a bit odd for people to write:
-  - `GET .../xxx/yyy/zzz` to reference a nested resource
-  - but then `GET ...?filter=xxx.yyy.zzz.prop=5` to access a property on that
-    same nested resources. So, despite some people might thinking that it's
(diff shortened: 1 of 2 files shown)


core/primer.md
```diff
@@ -3,17 +3,12 @@
...
-  - but then `GET ...?filter=xxx.yyy.zzz.prop=5` to access a property on that
-    same nested resources. So, despite some people might thinking that it's
-	odd to see `/` in a query parameter, the consistency was more important
-	to us. People can now copy-n-paste the same ref string in both uses.
+- it's RECOMMENDED that all extensions be defined in advance as part of the
+  model. However, the spec does support an extension of "*" to allow for
+  unknown/runtime-provided extensions. Since extensions can appear in case-
+  insensitive situations (e.g. http header) we can't know the case of them
+  when storing in the backend. As a result, all attributes and keyName MUST
+  e lowercase.
```
jskeet, line 11:
```suggestion
  be lowercase.
```

duglin: fixed

duglin:
Approved on the 12/7 call

PR #65, opened 2023-12-05
Minor / stylistic suggestions

Fannon:
Nothing that changes the spec, just minor enhancements to formatting, code block usages, etc.

Feel free to pick and choose. 

## Proposed Changes
* Fixing minor typos
* Use of the markdown code block syntax highlighting more to the advantage of readers and writers. When adding syntax highlighting, I've also noticed a few syntax errors that I've fixed as well.  
  * In some cases the use of ```http or ```yaml may be a stretch. Or application/json formatting (via ```http) where the response is actually not a valid JSON due to `*` etc. annotations.

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -85,7 +85,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "specVersion": "STRING",
   "id": "STRING",
   "name": "STRING", ?
-  "epoch": "UINTEGER",
+  "epoch": UINTEGER,
   "self": "URL",
   "description": "STRING", ?
   "documentation": "URL", ?
@@ -126,7 +126,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
           "STRING": {                   # Key=plural name, e.g. "definitions"
             "plural": "STRING",         # e.g. "definitions"
             "singular": "STRING",       # e.g. "definition"
-            "versions": "UINTEGER" ?    # Num Vers(>=0). Default=1, 0=unlimited
+            "versions": UINTEGER ?    # Num Vers(>=0). Default=1, 0=unlimited
             "versionId": BOOLEAN, ?     # Supports client specified Version IDs
             "latest": BOOLEAN, ?        # Supports client "latest" selection
             "hasDocument": BOOLEAN, ?   # Has a separate document. Default=true
@@ -144,7 +144,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
     "ID": {                                        # Key=the Group id
       "id": "STRING",                              # The Group ID
       "name": "STRING", ?
-      "epoch": "UINTEGER",
+      "epoch": UINTEGER,
       "self": "URL",
       "description": "STRING", ?
       "documentation": "URL", ?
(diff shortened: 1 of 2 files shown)


core/spec.md
```diff
@@ -80,12 +80,12 @@ to a "Schema" document - all within the same Registry instance:
...
 For easy reference, the JSON serialization of a Registry adheres to this form:
 
-``` text
+```yaml
 {
   "specVersion": "STRING",
   "id": "STRING",
   "name": "STRING", ?
-  "epoch": UINTEGER,
+  "epoch": "UINTEGER",
```
duglin, line 88: Why the quotes when it's a number?

core/spec.md
```diff
@@ -1850,14 +1850,14 @@ Location: https://example.com/endpoints/456
...
 
-``` text
+```http
 POST /endpoints
 Content-Type: application/json; charset=utf-8
 
 [ { GROUP1 definition }, { GROUP2 definition } ]
+```
 
----
```
duglin, line 1859: same here

core/spec.md
```diff
@@ -3760,6 +3762,8 @@ and per [RFC7230, section 3][rfc7230-section-3], HTTP headers MUST only use
 printable characters from the US-ASCII character set, and are terminated by a
 CRLF sequence with OPTIONAL whitespace around the header value.
 
+> TODO: `precent-encoded` - typo?
+
 When encoding an attribute's value as an HTTP header it MUST be
 precent-encoded as described below. This is compatible with [RFC3986, section
```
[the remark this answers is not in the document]
duglin: :-) yes it's a typo - should be "percent"

core/spec.md
```diff
@@ -3691,7 +3693,7 @@ The abstract processing logic would be:
 
 The format of `EXPRESSION` is:
 
-``` text
+```
```
duglin, line 3696: I think @clemensv needed some tag to appear on these for his tooling, which is why I picked `text` as it was generic.

core/spec.md
```diff
@@ -80,12 +80,12 @@ to a "Schema" document - all within the same Registry instance:
 
 For easy reference, the JSON serialization of a Registry adheres to this form:
 
-``` text
+```yaml
```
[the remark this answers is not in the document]
Fannon: I chose the YAML formatting here because it works surprisingly well with the notation you use. It's not valid JSON either. 

With YAML you'll get somewhat decent visual highlighting. But technically it's not correct. Falling back to ``` or ```text would be an alternative.

core/spec.md
Fannon: Ok I got it. UINTEGERS are not quoted, like boolean. Will revert this particular change.

core/spec.md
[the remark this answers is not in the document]
duglin: ok if `yaml` works let's do that!

duglin:
@Fannon can you sign all of your commits? Or squash them down to one and sign that one?

LGTM

Fannon:
> @Fannon can you sign all of your commits? Or squash them down to one and sign that one?
> 
> LGTM

Yes, will do! (Already did it, but I haven't set this up so my last commits broke it again)

core/spec.md
Fannon: Ok, I added this for consistency reason - so that some markdown code block renderers don't start *guessing*

duglin:
Approved on the 12/7 call

PR #66, opened 2023-12-13
Week of Dec 11 edits - model clarifications

duglin:
- only allow backwards compatible changes to the model
  - e.g. no new required attributes, no deleting of attributes
- cleaned-up the text around extension attributes requirements
- groups/resource names must not be longer than 58 (not 63) due to `count` being appended on the collection attributes


duglin:
Approved on the 12/14 call

PR #67, opened 2024-01-02
Week of Jan 1 changes

duglin:
- Add a psuedo-JSON serialization for Endpoint, Message, Schema specs
- make it clear the `#` starts a comment in samples
- make Endpint/config.endpoints a map so we can add extension attributes per URI
- s/definition/message/g


message/spec.md
```diff
@@ -34,20 +123,19 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
-wants. A receiver that does not support that feature is free to take any
-action it wishes, including no action or generating an error, as long as
-doing so does not violate other requirements defined by this specification.
-However, the RECOMMENDED action is to ignore it. The sender SHOULD be prepared
-for the situation where a receiver ignores that feature. An
-Intermediary SHOULD forward OPTIONAL attributes.
+For clarity, OPTIONAL attributes (specification defined and extensions) are
+OPTIONAL for clients to use, but servers MUST be prepared for them to appear
+in incoming requests and MUST support them since "support" simply means
+persisting them in the backing datastore. However, as with all attributes, if
```
Fannon, line 129: This also includes re-exposing them via the API to consumers?

duglin: yes, they would then need to return them in a GET.

cloudevents/spec.md
```diff
@@ -24,20 +24,19 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+accepting the attribute would result in a bad state (such as exceeding a size
+limit, or results in a security issue), then the server MAY choose to reject
+the request.
 
 In the pseudo JSON format snippets `?` means the preceding attribute is
 OPTIONAL, `*` means the preceding attribute MAY appear zero or more times,
-and `+` means the preceding attribute MUST appear at least once.
+and `+` means the preceding attribute MUST appear at least once. The presence
+of the `#` character means the remaining portion of the line is a comment.
+Whitespace in the JSON snippets are used for readability and are not normative.
```
jskeet, line 39: "Whitespace characters are" or "Whitespace is" perhaps?

core/spec.md
```diff
@@ -1055,6 +1059,25 @@ by the implementation of the Registry Service. This information will usually
 be used by tooling that does not have advance knowledge of the type of data
 stored within the Registry.
 
+When a new implementation is first created, or a new specifcation is based
```
jskeet, line 1062: specifcation => specification

core/spec.md
```diff
@@ -1055,6 +1059,25 @@ by the implementation of the Registry Service. This information will usually
...
+on the xRegistry specification, then it MAY choose to modify the core
+attributes to fit their needs. The changes allowed are restricted to the
```
jskeet, line 1064: their => its? (Given the "it may choose")

core/spec.md
```diff
@@ -1055,6 +1059,25 @@ by the implementation of the Registry Service. This information will usually
...
+- `ifvalue` specifications MAY be defined
+
+To indicate a change to a core attribute, the attribute MUST be defined
+as part of the Registry model. Once a Registry instance is created, further
+changes to these core attribute MUST adhere to the rules defined in the
+[Updating the Registry Model](#updating-the-registry-model) section.
+
+Any core attributes that are not defined in the model of an instance of a
+Registry MUST be implicitly defined. In other words, the Registry's model
+consists of the core attributes (as defined by this specification) overlayed
```
jskeet, line 1077: overlayed => overlaid

core/spec.md
```diff
@@ -2317,14 +2359,20 @@ and the following Resource specific attributes:
...
+  an item in some private store/cache, rather than a networked document
+  location, then it is RECOMMENDED for the value to be a uniform resource
+  name (URN).
 - Constraints:
-  - REQUIRED if the Resource contents are stored outside of the current
+  - REQUIRED if the Resource contents are not stored inside of the current
     Registry
-  - the referenced URI MUST support an HTTP(s) `GET` to retrieve the contents
+  - if the contents are stored in a network accessible endpoint then the
+    referenced URL MUST support an HTTP(s) `GET` to retrieve the contents
```
jskeet, line 2375: I'm confused as to the change from URI to URL here, given that the type of changing from URL to URI (and the name still uses URL...)

endpoint/spec.md
```diff
@@ -29,20 +136,19 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD",
...
+accepting the attribute would result in a bad state (such as exceeding a size
+limit, or results in a security issue), then the server MAY choose to reject
+the request.
 
 In the pseudo JSON format snippets `?` means the preceding attribute is
 OPTIONAL, `*` means the preceding attribute MAY appear zero or more times,
-and `+` means the preceding attribute MUST appear at least once.
+and `+` means the preceding attribute MUST appear at least once. The presence
+of the `#` character means the remaining portion of the line is a comment.
+Whitespace in the JSON snippets are used for readability and are not normative.
```
jskeet, line 151: Another whitespace are. (This may be fine for everyone else; it just sounds odd to me.)

jskeet (commented):
(Not exhaustive, and only nitpicking really.)

cloudevents/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: Might need to tweak the wording, but it's defined as a URI to allow for just about anything - including something that's not network accessible... like a URN or some well known identifier. However, if it is network accessible then it needs to be a URL.  Does that make any sense?

duglin: Let's discuss the URL vs URI in the name during the call... that's a good one :-)

duglin:
Thanks for the review @Fannon and @jskeet !

core/spec.md
duglin: Do you think I need to change the text to make it clearer?

jskeet: Probably not. I wouldn't have spotted it at all if it weren't for the fact that it changed in one direction when the type changed in the opposite direction in the same PR.

duglin:
Approved on the 1/18 call

PR #71, opened 2024-02-06
Week of Feb 5th edits

duglin:
- Add a "readonly" attribute to Resources in the model
- rename "strict" to "deployed" in Endpoint spec
- removed all "-" from attribute names in Endpoint spec
- soften Schema spec to say schema versions SHOULD be backwards compatible,
  not MUST be
- s/required/clientrequired/ & add "serverrequired" for model attributes
- maps must be specified in their entirety when appearing as HTTP headers
- move "origin" to Core spec
- move "format" from Core to Endpoints, Message and Schema specs
- clarify when self/latestversionurl have ?meta (or not)
- clean-up some of the spec's `model.json` files

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -41,11 +41,11 @@ automation and tooling.
 A Registry Service is one that manages metadata about Resources. At its core,
 the management of an individual Resource is simply a REST-based interface for
 creating, modifying and deleting the Resource. However, many Resource models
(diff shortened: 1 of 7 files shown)


duglin:
Jon asked if we should go the other direction and add a word separator. We agreed to revisit this on the next Tues call

PR #74, opened 2024-02-14
Week of Feb 12 changes

duglin:
- More model/spec sync'ing
- clarify that attribute.item.type is only used by map/array attributes (not objects)
- ifvalue.VALUE must be a string serialization of the possible runtime value
- ifvalue.VALUE must not be an empty string
- ifvalue can not include nested ifvalue
- can't implicitly create higher-level entities if they have clientrequired attributes
- add attribute.default to define an attribute's default value
- default attribute.versions to zero (unlimited versions), was 1
- more possible topics for the primer


duglin:
Approved on the 2/15 call

PR #75, opened 2024-02-18
Updates for week of Feb18

duglin:
- define valid map key name chars: `[a-z0-9]`, `-`, `_` or a `.`
- key names: HTTP headers: xRegistry-ATTRIBUTENAME-KEY: VALUE
  - KEY can have `-`, so 3rd+ `-` is part of the key name
- more model clean-up
- add more primer stuff and start to elaborate on the primer topics
- `null` in a scalar array must generate an error
- s/ifValue/ifValues/g
- split `item` into two things: `attributes` and `item`
  - `item` is used for map/array
  - `attributes` are for objects
  - cleans up the model a bit - it was error prone before
- defined a `contenttype` attribute, use Content-Type http header
- allow for `latest` to be specified even when server doesn't support
  client-side setting of "latest version" as long as their desire matches
  what the server was going to do anyway. Else error

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -1,14 +1,74 @@
 # xRegistry Primer
 
-Possible future topics to cover:
-- Resource.ID and Resource.Version.ID might have the same values but they are
-  very different properties with no relation to each other.
+<!-- no verify-specs -->
+
+## Design decisions
+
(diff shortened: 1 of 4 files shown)


duglin:
Approved on the 2/22 call

PR #76, opened 2024-02-19
Message model

deissnerk:
Update message spec:

Fix JSON errors
Change to `messagegroups` / `messages`
Change to `clientrequired` / `serverrequired`

### DIFF
--- a/message/model.json
+++ b/message/model.json
@@ -16,7 +16,7 @@
         "binding": {
           "name": "binding",
           "type": "string",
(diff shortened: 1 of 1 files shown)


duglin:
LGTM

deissnerk:
@duglin 
> We probably don't need the "*required": false
stuff

You are right. It is very noise, and the default is anyways `false`. I wanted to discuss this in today's call. I also wonder, if the nested occurrence of `clientrequired` / `serverrequired` is well-defined for all combinations.

duglin:
Approved on the 2/22 call

PR #79, opened 2024-03-01
Updates for week of March 4

duglin:
- "latest" on a request that would result in "no latest version" is an error
- clarify that attributes with default values don't require their owning
  Objects to be present
- be consistent about case after dashes (e.g. -word vs - Word)
- treat hasDocument resources like Groups. Use of ?meta is meaningless, every
  MUST always be in the HTTP body. No xRegistry HTTP headers
- DELETE will NOT return the deleted resources, use "204"
- "contenttype" is ok to use with RESOURCEurl. No guarantee it matches the
  Content-Type of a the GET to RESOURCEurl though
- URI Template type : per RFC6570 section 3.2.1 (level 1)


duglin:
Approved on the 3/7 call

PR #83, opened 2024-03-28
schema gen update

clemensv:

Update for schema generator and fix in message model (AMQP app props)


duglin:
Approved on the 3/28 call

PR #86, opened 2024-04-12
Changes for week of Apr 15

duglin:
- s/createdon/createdat/g
- s/modifiedon/modifiedat/g
- allow for "this" in "?setdefaultversionid" in cases where server picks ID
- reserve "null" and "this" as vIDs that clients can not use
- s/setdefault/setstickydefaultversion/ in the model   (allow user to set which version is sticky)
- s/setdefault/stickydefaultversion/ as Resource attribute  (is this Resource's default version "sticky")

### DIFF
--- a/core/model.json
+++ b/core/model.json
@@ -51,8 +51,8 @@
       "type": "string",
       "readonly": true
     },
-    "createdon": {
-      "name": "createdon",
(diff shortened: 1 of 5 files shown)


duglin:
Approved on the 4/18 call

PR #89, opened 2024-04-24
Updates for week of Apr 22

duglin:
- clarify when ?setdefaultversionid is required when creating a new Resource
  + one or more Versions, for example:
  + rID doesn't exist
  + `POST /GROUPs/gID/RESOURCEs/rID/versions` + body=collection of Versions
  + collection is a map (no order) so we need guidance on which is to be "default"
- clarify that ?setdefaultversionid is processed before we apply the Resource's properties (IOW, we update the new "default version", not the old one) - for example:
  + defaultVersionID=v3
  + `PUT /GROUPs/gID/RESOURCEs/rID?setdefaultversionid=v5`
  + v5 is updated, not v3

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -3243,17 +3243,9 @@ Where:
 Where:
 - See [Default Version of a Resource](#default-version-of-a-resource) for
   more information when there is no HTTP body and both `meta` and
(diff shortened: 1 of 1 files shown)


duglin:
Approved on the 4/25 call

Issue #103, opened 2024-04-30
Support PATCH and tweak semantics of some attributes


duglin:
Proposal:
![image](https://github.com/xregistry/spec/assets/1944671/857a6493-95f4-4c32-844e-6808a7a3ea27)

Also remove the setdefaultversionid query parameter - use the attribute in the resource instead

Issue #102, opened 2024-04-30
Should Registry entity have "createdat" and "modifiedat" attributes?


duglin:
Proposal: yes

Note that modifiedat would only be changed when the Registries properties are touched, not a child entity.

Issue #93, opened 2024-04-24
Createdat/createdby/modifiedat/modifiedby fields - keep? drop? limit?


duglin:
Questions:
1. Do we want to keep createdat/modifiedat since it's about registry timestamps and not doc timestamps?
2. Do we want to keep createdby/modifiedby even though we don't deal with auth?
3. What should happen during an export/import? Keep the values? Reset on import? Sync op?
4. If we keep them, what do we do if the feature if turned off/on? Erase values? Retain?
5. Who controls whether this info is tracked? Strictly server or user configured?
6. Should people be allowed to erase the values?
7. Should they be "show" type of properties? Meaning, always track them but just control whether they're visible?

Broader CE group: sees value in keeping them, but keeping them under server's control

Proposal:
- make createdat, modifiedat required for server to support and expose, like id, epoch... but mutable
- remove xxxBY fields, create an extension if needed
- add createdat and modifiedat to Registry entity (as with all entities, modifiedat only changes when attributes on the Registry change, not when a child entity changes)

Issue #90, opened 2024-04-24
Challenges w.r.t. uploading multiple Versions


duglin:
**Challenges**:
- When doing a PUT/POST on a Resource, when do we apply the Resource's (default version) properties? Related to order of processing ?setdefaultversionid + POST /versions on a resource that was just added to spec 
- What if the default props differ from the corresponding Version's props?
- If `?setdefaultversionid` can not be used (e.g. uploading multiple Resources), how does someone set it? Same for sticky flag

**Proposal**:
Assume the input is the desired state of the system, at least within the scope of the data provided (no auto-deletes); the processing should result in it looking like they did things in this order:

1. If `?inline` is present, upsert any Versions in the "versions" collection

2. If the Resource is part of a collection of Resources in the request message (e.g. `/`  or  `/GROUPs/gID`  or `/GROUPs/gID/RESOURCEs` ) then:
    - Apply `defaultversionid` Resource property if present on that Resource
    - Apply `stickydefaultversion` Resource property if present on that Resource
    _- Concern: sometimes readonly props aren't readonly_

3. If the Resource is not part of a collection (e.g. `/GROUPs/gID/RESOURCEs/rID` ) then:
    - Apply `?setdefaultversionid` query parameter, if present

4. Apply Resource properties to the current default Version IFF default version is not in the "versions" collection
This means that any diff between Resource and Version, Version wins

If epoch doesn't match on any Resource or any Version, error

Add support for these on PUT & POST operations:
- `?noepoch` - turn off epoch checking
- Only when ?inline is used
  - `?nodefaultversion` - only apply `defaultversionid` properties if no default version is already set (e.g. a new resource)
  - `?nostickydefaultversion` - turn off applying the `stickydefaultversion` property, if present (perhaps this should just be merged into the previous one)

Is it bad that we're inconsistent in that we support setting the "default" properties via json when the resource is part of a collection, but not when it's stand-alone (we require the ?setdefaultversionid query param be used)?

These flags allow people to GET and the POST w/o having to touch the data to remove epoch and the defaultVersion properties, if they don't want to them to be applied

Note: this is related to #88 - once we figure out how to order the uploaded versions, that decision might influence how we deal with the concerns here.

duglin:
As of today, the spec says this w.r.t. `setdefaultversionid` query parameter:
```
- Aside from the special values of `null` and `this`, its value MUST be
  the `id` of a Version for the specified Resource after all Version processing
[...]

PR #104, opened 2024-05-01
Changes for the week of May 1

duglin:
- clarify that `id` is required in responses and doc view, not requests since
      the `id` can be inferred in most other cases
- fix some typos- lots of cleanup, mainly trying to remove duplicate information
- moved `createdby` and `modifiedby` to be "Potential extensions" mentioned
  in the Primer
- added `createdat` and `modifiedat` as Registry attributes, to be consistent
  with Groups, Resources and Versions
- made `createdat` and `modifiedat` required attributes for servers to support
  and expose. They're also now mutable, so clients can set them when needed
  and we preserve their values during an "import"
- all `createdat` and `modifiedat` that are updated during the SAME operation
  that use "now" MUST use the same value. This means that if you upload a
  set of Versions w/o `createdat` timestamps, it will be indeterminate as to
  which Version will be the newest
- added support for PATCH, same as PUT but missing attributes remain unchanged
- made `stickydefaultversion` and `defaultversionid` Resource properties
  mutable - this is now the preferred way to set "default"
  - `stickydefaultvertsion` is processed first, then if 'true' we'll look
    at the `defaultversionid` attribute
- removed `?setdefaultversionid` query parameter when the Resource entity
  is present in the request, however we still need it for cases where
  Versions are modified by the Resource entity isn't present in the request
  - e.g. deleting Versions, or uploading Versions directly to the "versions"
  collection
- removed the ability to set just the default version pointer via a special
  POST to the resource. Instead just PATCH the Resource
- when processing a Resource with nested Versions:
  - process the "versions" collection first
  - apply the "defaultversionid"
  - apply the Resource's "default" attributes IFF the default Version wasn't
    in the incoming "versions" collection - to avoid conflicts

Fixes: #88
Fixes: #90
Fixes: #93
Fixes: #103
Fixes: #102

Also see: 

image

for how each type of attribute is processed during a PUT, POST or PATCH

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -152,3 +152,15 @@ Let's walk through a complex example:
 - The result is: v8 regardless of whether v8 was created with latest=true or
   not
 
+# Potential Extensions
+
+- `createdby`, `modifiedby`<br>
+  These are related to the `createdat` and `modifiedat` attributes in that
+  they would normally be updated at the same time as their corresponding
+  `*at` attribute, and are use to track the identity of the person or
+  component that performed the related operation.
+
+  The xRegistry specfication does not define these since it does not define
+  any authentication mechanisms, or even manage tracking of these identities.
+  However, if an implementation does track this information, these attributes
+  might be of interest.
--- a/core/spec.md
+++ b/core/spec.md
@@ -15,8 +15,14 @@ automation and tooling.
 - [Registry Attributes and APIs](#registry-attributes-and-apis)
   - [Attributes and Extensions](#attributes-and-extensions)
   - [Registry APIs](#registry-apis)
+    - [Registry Collections](#registry-collections)
+    - [Entity Processing Rules](#entity-processing-rules)
   - [Registry Entity](#registry-entity)
+    - [Retrieving the Registry](#retrieving-the-registry)
+    - [Updating the Registry Entity](#updating-the-registry-entity)
   - [Registry Model](#registry-model)
+    - [Retrieving the Registry Model](#retrieving-the-registry-model)
+    - [Updating the Registry Model](#updating-the-registry-model)
   - [Groups](#groups)
     - [Retrieving a Group Collection](#retrieving-a-group-collection)
     - [Creating or Updating Groups](#creating-or-updating-groups)
@@ -61,13 +67,13 @@ Group is similar to a "directory" on a filesystem.
 Resources represent the main data of interest for the Registry. In the
 filesystem analogy, these would be the "files". A Resource exist under a
 single Group and, similar to Groups, have a set of Registry metadata.
-However, unlike a Group which only has Registry metadata, Resources can
-have an associated "document" associated with it. This specification places
-no restriction on the type of content stored in the Resource's document.
+However, unlike a Group which only has Registry metadata, each Resource can
+have a "document" associated with it. For example, a "schema` Resource might
+have a "schema document" as its "document". This specification places no
+restriction on the type of content stored in the Resource's document.
 
 This specification defines a set of common metadata that can appear on both
-Groups and Resources, and allows for Registry (domain-specific) extensions to
-be added.
+Groups and Resources, and allows for domain-specific extensions to be added.
 
 The following 3 diagrams show (from left to right): 1) the core concepts of
 the Registry in its most basic form, 2) a Registry concept model with multiple
@@ -92,6 +98,8 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "description": "STRING", ?
   "documentation": "URL", ?
   "labels": { "STRING": "STRING" * }, ?
+  "createdat": "TIME",
+  "modifiedat": "TIME",
 
   "model": {                            # Only if inlined
     "schemas": [ "STRING" * ], ?        # Available schema formats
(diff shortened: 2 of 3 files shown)


core/primer.md
```diff
@@ -152,3 +152,15 @@ Let's walk through a complex example:
 - The result is: v8 regardless of whether v8 was created with latest=true or
   not
 
+# Potential Extensions
+
+- `createdby`, `modifiedby`<br>
```
jskeet, line 157: I'd expect a trailing double space to end up as a Markdown line break.

jskeet (commented):
One initial tiny comment - more to come...

core/primer.md
duglin: hmm, will need to think about that since we have a check for extra trailing spaces :-)

core/spec.md
```diff
@@ -62,12 +68,12 @@ Resources represent the main data of interest for the Registry. In the
 filesystem analogy, these would be the "files". A Resource exist under a
 single Group and, similar to Groups, have a set of Registry metadata.
 However, unlike a Group which only has Registry metadata, Resources can
-have an associated "document" associated with it. This specification places
-no restriction on the type of content stored in the Resource's document.
+have a "document" associated with it. For example, a "schema` Resource might
```
jskeet, line 71: The "it" here is a little odd, as Resources is plural. (Pre-existing issue.)

jskeet: (Perhaps "each Resource can have a "document" associated with it"?

core/spec.md
```diff
@@ -336,10 +338,8 @@ be one of the following data types:
     alphanumeric characters (`[a-z0-9]`), `-`, `_` or a `.`; be no longer
     than 63 characters; start with an alphanumeric character and be unique
     within the scope of this map
-  - When serialized as an HTTP header, each key MUST appear as a separate
-    HTTP header prefixed with `xRegistry-` and the map's owning attribute
-    name. The header value MUST be the key's "value". See
-    [HTTP Header Values](#http-header-values) for more information
+  - See [Serializaing Resource Documents](#serializing-resource-documents)
+    for more information serializing maps as HTTP headers
```
jskeet, line 342: "information serialization" - add "about" or "on" between the two words?

core/spec.md
```diff
@@ -452,6 +450,12 @@ The definition of each attribute is defined below.
     be based on where the Registry is used. For example, a publicly accessible
     Registry might want to consider using a UUID, while a private Registry
     does not need to be so widely unique.
+  - In cases where an entity's `id` is specified outside of the serialization
+    of the entity (e.g. part of a request URL, or a map key), then its
```
jskeet, line 454:
```suggestion
    of the entity (e.g. part of a request URL, or a map key), its
```

(Or change to "If an entity's `id` [...]")

core/spec.md
```diff
@@ -577,67 +588,45 @@ Then the existing entity can be deleted.
...
-  - If this feature is enabled after an entity has already been created,
-    that entity's `createdat` value MUST remain empty
+  - When present in a write operation request, the value MUST override any
+    existing value, however a value of `null` MUST use the current date/time
+    as the new value
+  - When absent in a write operation request, any existing value MUST remain
+    unchanged, or if not present, set to the current date/time
+  - In cases where multiple entity's `createdat` attribute is set to the
+    current date/time within the same operation, the same value MUST be
+    applied to all of the entities
```
jskeet, line 604:
```suggestion
  - In cases where the `createdat` attribute is set to the
    current date/time on multiple entities within the same operation, the same value MUST be
    applied to all of the entities
```

jskeet: (Current "multiple entity's" doesn't work in terms of plural/singular; "multiple entities'" might be okay but looks clumsy to me.)

core/spec.md
```diff
@@ -577,67 +588,45 @@ Then the existing entity can be deleted.
...
+  - Setting an entity's `modifiedat` value MUST NOT update any parent
+    entity's `modifiedat` value
+  - When present in a write operation request, the following applies:
+    - If the request value is the same as the existing value, then the
+      current date/time MUST be used as its new value
+    - If the request value is different than the existing value, then the
+      request value MUST be used as its new value
+    - If the request value is `null` then the current date/time MUST be used
+      as the new value
+  - When absent in a write operation request, it MUST be set to the current
```
jskeet, line 625: As above.

core/spec.md
```diff
@@ -701,6 +690,42 @@ incorporate authentication and/or authorization mechanisms for the APIs.
 If an API is not supported by the server then a `405 Method Not Allowed`
 HTTP response MUST be generated.
 
+This specification attempts to following a standard REST/HTTP processing model.
```
jskeet, line 693:
```suggestion
This specification attempts to follow a standard REST/HTTP processing model.
```

core/spec.md
```diff
@@ -844,45 +862,50 @@ entity or Registry collection, the overall pattern is defined once in this
...
-     processed.
-   - If the HTTP body is a serialization of xRegistry metadata, and it contains
-     an `id` attribute, then it MUST match the `id` used in the URL. Otherwise
-     an error MUST be generated.
+  - The URL MUST be of the form: `PATH-TO-ENTITY`.
+  - The HTTP body MUST contain the full serialization of the entity to be
+    processed.
+  - The entity processed MUST either be created (if it does not already
+    exist), or updated (if it does exist).
+  - Any missing mutable attribute, or present with a value of `null`, MUST be
```
jskeet, line 882:
```suggestion
  - Any mutable attribute which is either missing or present with a value of `null`, MUST be
```

jskeet (commented):
As usual, I got part way through before running out of time I can realistically spend reviewing. Hope the partial review is still useful.

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

duglin:
Approved on the 4/2 call.
@jskeet thanks again for the review - always very much appreciated!

PR #109, opened 2024-05-05
Changes for week of May 6

### LINKED ISSUE #73: Consider allowing nested `ifvalues`
Consider allowing ifvalue siblingattributes to add additional attributes via a nested ifvalue section

### LINKED ISSUE #91: When do we inline the Resource document vs Base64 encode it?
We need to have a well defined algorithm that servers can follow to decide when/if to use RESOURCE vs RESOURCEbase64 when someone asks for the Resource's document to be inlined (`?inline=RESOURCE`).

Some thoughts:
- w/o a `contenttype` value, we probably can't make any assumption about the data even if it "looks like" it might be in the right format. So, w/o that hint we should probably always base64 encode it
- if the `contenttype` value aligns with the server's serialization format (e.g. app/json, or ...+json) ,AND it is actually valid per that content type (meaning a JSON parser parses it successfully w/o any edits/escaping), then we can be able to write those bytes "as is" as part of the `RESOURCE` property's value. Pretty printing is allowable
- else base64 encode it

It would be nice to have a flag (`?base64`) on GET to force base64 encoding if people want the raw bytes

Do we need a way for someone to indicate that they want their data "escaped" rather than base64 encoded? Meaning, in the JSON case, a json escaped string. For example, perhaps: `?format=default|base64|escaped` where `escaped` means: if all chars are in `0x20-0x7E, \n, \r, \t` then we escape it a return it as an escaped string, otherwise base64 it. This way if people have simple strings like `hello\nworld` they can see it w/o having to base64 decode it. Default format value would be `default`.

Should it be `binary` instead of `base64` if the default way to send binary data for some format isn't base64?

### LINKED ISSUE #94: Allowed nested IfValue defintions?
Proposal: yes

### LINKED ISSUE #97: Update model via `PUT /` - support?
Today people can update the model via `PUT /model` and `PUT /`  + data+model in the body

Seems odd that someone can't define the registry and upload the data in one import operation when they can export it in one.

Proposal: allow it

We could choose to not support it for v1 but we need to decide whether to consider it for v1 or not

### LINKED ISSUE #110: ensure we support ?inline=* type of semantics
To mean: inline everything

Should this include all Resource documents too? Or should we allow for two options:
- ?inline=< all metadata >
- ?inline=< all metadata+resource >

i think having them split would be nice since someone may want the metadata w/o the large download of ALL Resource documents.

duglin:
- add support for updating the "model" along with the rest of the Registry entity. Require ?model
- support nested IfValues
- add a `typemap` attribute to the Resource's model definition to allow
  people to tell us which custom `contenttype` values are actually JSON,
  or String so we can serialize them "inline" when "?meta" is used
- add back in support for inline=*

Fixes #97 
Fixes #91 
Fixes #110 
Fixed #73

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -146,7 +146,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
             "setstickydefaultversion": BOOLEAN, ?    # Supports client "default" selection
             "hasdocument": BOOLEAN, ?   # Has a separate document. Default=true
             "readonly": BOOLEAN, ?      # From client's POV. Default=false
-            "immutable": BOOLEAN, ?     # Once set, can't change. Default=false
+            "typemap": MAP, ?           # contenttype mappings
             "attributes": { ... }, ?    # See "attributes" above
           } *
         } ?
@@ -202,6 +202,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
               "name": "STRING", ?
               "epoch": UINTEGER,
               "self": "URL",
+              "isdefault": BOOLEAN, ?
               "description": "STRING", ?
               "documentation": "URL", ?
               "labels": { "STRING": "STRING" * }, ?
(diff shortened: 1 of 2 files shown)


duglin:
Approved on the 5/9 call

Issue #116, opened 2024-05-13
Do we need to be more precise w.r.t. `contenttype`->`typemap` mapping?


duglin:
E.g.   does `text/html;charset=utf-8` match `text/html` ?

Proposal: just the `type/sub-type` of header value (parameters are excluded, anything after the first `;` is ignoed).

Algorithm:
- extract the media-type from the Resource's `contenttype` attribute by taking everything before the first `;`
- Check user-defined `typemap` keys, if only one matched then return its value
- If more than one matches, and they're all the same 'value', then return that value
- if more than one matches, but they're not all the same `value`, then treat it as "binary"
- if none matches then check the default type map as defined by the spec (text/plain, app/json, *+json) 
- the default ones MAY be overwritten thru an explicit entry in the `typemap`

Note that "*" is a valid char in a media type. Do we need to support escaping "\*" in the pattern?
Case insensitive compares - rfc: `The type and subtype tokens are case-insensitive.`

Another scenario:
- user PUTs to a Resource with `{ "file": "hello" }`, but no `contenttype` - do we treat this as binary or a string"
- what about `{"file": { "foo": "bar"       } }` ? if it's binary, are we allowed to strip whitespaces (or pretty print) before we store the bytes? 
- basically, if we know it's JSON on the input, can the server make any smart decisions despite no contenttype attribute?
- can it set "contenttype" to "app/json" automatically for the user? While I'm not a fan of touching the user's data, at least it would give them a clue as to why we're treating it as JSON and not binary. Otherwise, if we don't set "contenttype" we have no choice but to treat it as binary (per our previous decisions) - is that ok?

What if a user does PUT `{"contenttype": "foo/bar", "file": { "foo": "bar"       } }`  ?

Proposal:
- on input: if resource is inlined, then the server is free to pretty-print/strip whitespaces. Stores result as bytes.
  - no content-type: set it to app/json - PUT: even if it already has a value. PATCH: only if it has no value
  - has content-type: save it, as is
- on output:
  - no content-type: binary
  - app/json: json
  - text/plain: string
  - other: binary

duglin:
Agreed to all except the need to escape "*" - it's not a valid char per the media-type/mime spec

PR #118, opened 2024-05-15
Updates for week of May 15

duglin:
- clarify how to use `typemap`
  - only use the "media-type" (no parameters) of the `contenttype` value
  - set `contenttype` automatically when Resource doc is inlined (sometimes)
  - if more than one key in `typemap` matches, and they're not the same value, the use `binary`
  - define a `binary` type to force using `RESOURCEbase64` if needed

Fixes: #116

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -1906,6 +1906,7 @@ The following describes the attributes of Registry model:
         the xRegistry metadata - in other words JSON, or<br>
     2 - The Resource's document can be considered a "string" and therefore
         can be serialized as a "string" with some escaping.<br>
+
     For some well-known `contenttype` values (e.g. `application/json`) the
     first case can be easily determined by the server. However, for custom
     values the server will need to be explicitly told to when to consider it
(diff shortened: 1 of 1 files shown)


core/spec.md
```diff
@@ -1906,30 +1906,56 @@ The following describes the attributes of Registry model:
...
         can be serialized as a "string" with some escaping.<br>
+
     For some well-known `contenttype` values (e.g. `application/json`) the
     first case can be easily determined by the server. However, for custom
     values the server will need to be explicitly told to when to consider it
     to be of the correct type. The `typemap` attribute allows for this by
     defining a mapping of `contenttype` values to well-known xRegistry format
     types.
 
+    Since the `contenttype` value is a "media-type" per
```
jskeet, line 1917: Remembering adding this to the CloudEvents spec recently, maybe emphasize that this has to be done case-insensitively? I see this is mentioned later on, but it feels useful to have everything to do with lookup specified in one place.

jskeet (commented):
A couple of nits, but otherwise seems fine from my perspective of "not really following xRegistry much" :)

core/spec.md
duglin: done

core/spec.md
```diff
@@ -3465,6 +3488,19 @@ Where:
...
     3 attributes (and any associated data).
+  - When `RESOURCE` is present, the server MAY choose to modify non-semantic
+    significant characters. For example, to remove (or add) whitespace. In
+    other words, there is no requirement for the server to persist the
+    document in the exact byte-for-byte format in which it was provided. If
+    that is desired then `RESOURCEbase64` MUST be used instead.
+  - On a `PUT` or `POST`, when `RESOURCE` is present, if no `contenttype`
+    value is provided then the server MUST set it to same type as the incoming
+    request, e.g. `application/json`, even if the entity previous had a
+    contenttype` value.
```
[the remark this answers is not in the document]
duglin: done

duglin:
Approved on the 5/16 call

PR #124, opened 2024-05-21
Changes for week of May 20

duglin:
- epoch and modifiedat are always updated on any write operation, even empty (`{}` ) PATCH
- s/?meta/$meta/
  - POST to `/GROUPs/gID/RESOURCEs/rID[$meta]` now only creates/updates one Version. 
  - POST to `/GROUPs/gID/RESOURCEs/rID/versions` is used to update or create one or more Versions as metadata
  - Before this PR `/GROUPs/gID/RESOURCEs/rID` was an alias for `.../versions` and we could tell if they wanted to create a single version by doc because of the missing "?meta". But now w/o ?meta, `.../versions$meta` is just too weird so `.../versions` is treated like all other collections- it takes a map, which means it takes a map of Resource metadata
- make all query params optional (via SHOULD) to enable simple static file server implemenations
- `id` is now case insensitive for uniqueness, but case sensitive for look-up
   - e.g. `PUT /dirs/d1/files/f1/versions/v1`  and then `GET /dirs/D1/files/F1/versions/V1` will fail for any of the 3 IDs specified 
- move logic from Makefile into new tools
- add casing explanation to primer

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -11,7 +11,7 @@ Resources, like all xRegistry entities, have unique `id` values. Resource
 meaning of the purpose of the underlying Resource document. Also, since they
 are static values and as the Resource changes over time (meaning, new Versions
 are created), it's important for end-users to have a static URL reference to
-the latest Version of the Resource.
+the default Version of the Resource.
 
 Versions of a Resource, on the other hand, might change quite often and the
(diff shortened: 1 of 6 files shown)


duglin:
Approved on the 5/23 call

Issue #120, opened 2024-05-17
Add support for .well-known file/access


duglin:
See: https://github.com/Fannon/xregistry-brainstorming/blob/main/.well-known/xregistry.json

Split into a few steps:
1 - define our .well-known file
2 - define how the data/files section - to allow for downloading of the xRegistry metadata

@Fannon

Fannon:
One idea: The `.well-known/xregistry.json` could be an alias, pointing to the root of the xRegistry API (if there is one). 
But the root of the xRegistry cannot be assumed to be in `/` of the host. Especially I would recommend to have the major version of the API protocol in there, so I would expect that a typical xRegistry API base path would be like `/xregistry/v1/`.

Fannon:
@duglin , @deissnerk : Here is a first PR: https://github.com/xregistry/spec/pull/125

Fannon:
Related: #106

duglin:
@Fannon should we close this one?

Fannon:
> @Fannon should we close this one?

yes, I think we agreed in Frankfurt that we will not have this and just link to everything relevant from xRegistry base (root entry point)

Issue #106, opened 2024-05-05
Allow for server to advertise its features


duglin:
E.g.
- write operations?
- inlining?
- filtering?
- updating the model?
others?

duglin:
Putting this on hold until we resolve: https://github.com/xregistry/spec/pull/125 since that could negate the need for a model-based discovery of this info.

PR #125, opened 2024-05-23
First proposal how to add .well-known URI location to xRegistry

Fannon:
As discussed in the meeting series, here is a first draft how we could incorporate the .well-known URI concept into the xRegistry.

This proposal only contains the minimum: Pointing to the xRegistry API location, which protocol version it is supporting and its capabilities. 

Related: #120 
Related: #106 

## Proposed Changes

- add  `.well-known/xregistry.json` as the entry point to an xRegistry server
- describe there where to find the xRegistry APIs (in the future, we may have v1 and v2) and their supported capabilities
- add top-level information that are free to give out openly (non-sensitive)
- adds `modelurl` to point to xRegistry metadata (or should this be protocol version dependent, too?)

**Release Note**

TODO: Will write this after we reviewed this.

```release-note

```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,7 +1,0 @@
-[*]
-end_of_line = lf
-indent_style = space
-trim_trailing_whitespace = true
-
-[*.{md}]
-max_line_length: 80
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,7 @@
+[*]
+end_of_line = lf
+indent_style = space
+trim_trailing_whitespace = true
+
+[*.{md}]
+max_line_length: 80
--- a/core/spec.md
+++ b/core/spec.md
@@ -1130,7 +1130,7 @@ following [RFC8615](https://datatracker.ietf.org/doc/html/rfc8615).
 It acts as a single entrypoint from where locations and capabilities of the
 xRegistry implementation can be retrieved. Since this URL is always the same,
 it can be used to find out whether xRegistry is supported and where to find the API.
-It also allows to support multiple versions of the xRegistry protocol at the same time
+It allows for support of multiple versions of the xRegistry protocol at the same time
 and allows implementers to choose their own API base path.
 
 The capabilities described on this level are independent from the xRegistry protocol version or its current meta model.
--- a/core/spec.md
+++ b/core/spec.md
@@ -1137,7 +1137,6 @@ and allows implementers to choose their own API base path.
 {
   # The following top-level attributes are identical to the xRegistry API root response
   # We may consider removing them here if we're bothered by the duplication.
-  "specversion": "STRING", # latest version of spec that is supported (?)
   "id": "STRING",
   "name": "STRING", ?
   "epoch": UINTEGER,
@@ -1150,14 +1149,15 @@ and allows implementers to choose their own API base path.
 
   # Describe where to find the xRegistry API and its capabilities
   "apis": {
-    "v1": { # Version of the xRegistry API protocol. In the future, there could be v2 in parallel.
-      "apiUrl": "URL", ? # Point to root URL of xRegistry API, e.g. /x-registry/v1/
+    "v1": { # Major version of the xRegistry API protocol. In the future, there could be v2 in parallel.
+      "specversion": "STRING", # full specversion, as returned by the API root response
+      "apiurl": "URL", ? # Point to root URL of xRegistry API, e.g. /x-registry/v1/
       "capabilities": ["inline", ""] # TODO: TBD
       # TODO: Also indicate whether API is available without protection or if protected, how to get access?
     }
   }
 
-  "modelUrl": { Registry model }, ? # Point to registry meta model
+  "modelurl": "URL", ? # Point to registry meta model
 }
 ```
 
--- a/core/spec.md
+++ b/core/spec.md
@@ -1156,6 +1156,8 @@ They resemble the capabilities of the xRegistry implementation itself that consu
 
 It is RECOMMENDED to leave this endpoint unprotected as it SHOULD NOT contain any sensitive information.
 
+The well-known location SHOULD be available starting from the base URL, to enable fully automated discovery of the xRegistry capability.
+
 ### Registry Entity
 
 The Registry entity represents the root of a Registry and is the main
--- a/core/spec.md
+++ b/core/spec.md
@@ -1152,7 +1152,7 @@ and allows implementers to choose their own API base path.
     "v1": { # Major version of the xRegistry API protocol. In the future, there could be v2 in parallel.
       "specversion": "STRING", # full specversion, as returned by the API root response
       "apiurl": "URL", ? # Point to root URL of xRegistry API, e.g. /x-registry/v1/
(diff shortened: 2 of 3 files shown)


.editorconfig
deissnerk, line 1: I suppose this needs to go into `.gitignore`?

deissnerk (commented):
Looks good to me

deissnerk:
@Fannon I think you need to sign your commits

.editorconfig
duglin: yea, we don't normally check-in editor specific files, but adding them to the .gitignore is fine

core/spec.md
```diff
@@ -86,6 +86,11 @@ to a "Schema" document - all within the same Registry instance:
  src="./xregfullmodel.png" height="300">&nbsp;&nbsp;&nbsp;<img
  src="./sample.png" height="300">
 
+One purpose of the xRegistry is to enable easy and fully automated,
+machine-readable discovery. The starting point of the discovery is via a
+[well-known location](#registry-well-known-config): `/.well-known/xregistry.json`.
+This is relative to the base URL of the server.
```
duglin, line 92: "of the server". Does this mean of the host or of the xReg server? Meaning: http://example.com or http://example.com/xregistry  if the latter is the base URL of the xReg server?

core/spec.md
```diff
@@ -1117,6 +1122,47 @@ HTTP body.
...
+It acts as a single entrypoint from where locations and capabilities of the
+xRegistry implementation can be retrieved. Since this URL is always the same,
+it can be used to find out whether xRegistry is supported and where to find the API.
+It also allows to support multiple versions of the xRegistry protocol at the same time
+and allows implementers to choose their own API base path.
+
+```yaml
+{
+  # The following top-level attributes are identical to the xRegistry API root response
+  # We may consider removing them here if we're bothered by the duplication.
```
duglin, line 1139: Some of these (like labels) might be container private data, and "self" wouldn't be accurate. It might be best to either remove these or sub-set them down to just simple (safe) things like "id", and make it optional since even that could expose too much info.

.editorconfig
Fannon: Yes, I'll gitignore it. The editorconfig file is not a editor specific file, it's actually a standard that most editors / IDEs understand.

core/spec.md
Fannon: @duglin : In my understanding example.com would be the location to the server (and this would be its base url / hostname). /xregistry would be a path that actually is the entry point of an API that the server offers (but there may be other APIs as well).

core/spec.md
```diff
@@ -1117,6 +1122,47 @@ HTTP body.
...
+  "epoch": UINTEGER,
+  "self": "URL",
+  "description": "STRING", ?
+  "documentation": "URL", ?
+  "labels": { "STRING": "STRING" * }, ?
+  "createdat": "TIME",
+  "modifiedat": "TIME",
+
+  # Describe where to find the xRegistry API and its capabilities
+  "apis": {
```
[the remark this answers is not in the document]
Fannon: Good point. First I wanted to make this an array, but then I remembered that you often preferred to have dictionaries. But since you don't need to have links to this, I would then propose something like:

```ts
apis: Array<{ version: "v1" | "v2", url: "URL" }>
```

core/spec.md
```diff
@@ -1117,6 +1122,40 @@ HTTP body.
...
+The registry comes with a well known location `/.well-known/xregistry.json`,
+following [RFC8615](https://datatracker.ietf.org/doc/html/rfc8615).
+
+It acts as a single entrypoint from where locations and capabilities of the
+xRegistry implementation can be retrieved. Since this URL is always the same,
+it can be used to find out whether xRegistry is supported and where to find the API.
+It also allows to support multiple versions of the xRegistry protocol at the same time
+and allows implementers to choose their own API base path.
+
+The capabilities described on this level are independent from the xRegistry protocol version or its current meta model.
```
Fannon, line 1136: @duglin : I've updated the PR a little bit to reflect our discussions today.

core/spec.md
```diff
@@ -1117,6 +1122,42 @@ HTTP body.
...
 
+### Registry Well-Known Config
+
+The registry comes with a well known location `/.well-known/xregistry.json`,
+following [RFC8615](https://datatracker.ietf.org/doc/html/rfc8615).
+
+It acts as a single entrypoint from where locations and capabilities of the
+xRegistry implementation can be retrieved. Since this URL is always the same,
+it can be used to find out whether xRegistry is supported and where to find the API.
+It also allows to support multiple versions of the xRegistry protocol at the same time
```
duglin, line 1133: ...allows to support multiple versions...  ->  ..allows for support of multiple versions...

core/spec.md
```diff
@@ -1117,6 +1122,42 @@ HTTP body.
...
+  "description": "STRING", ?
+  "documentation": "URL", ?
+  "labels": { "STRING": "STRING" * }, ? #
+
+  # Describe where to find the xRegistry API and its capabilities
+  "apis": [{ # Major version of the xRegistry API protocol. In the future, there could be v2 in parallel.
+    "specversion": "STRING", # full specversion, as returned by the API root response
+    "apiurl": "URL", ? # Point to root URL of xRegistry API, e.g. /x-registry/v1/
+    "capabilities": ["write", "update", "inline", "filter", "..."] # TODO: TBD
+    "modelurl": "URL", ? # Point to registry meta model
```
duglin, line 1151: "modelurl" seems like it would be a duplicate of the `/model` API that's part of the API itself. no?
Or do you see cases where there's no "apiurl" but still a "modelurl" ?
I guess it depends on whether the "files" stuff that we've talked about includes the model definition or whether it's just the metadata.

core/spec.md
duglin: Based on today's discussion I assume line 92 would change to allow for it to live some place else, right?

Fannon: > Based on today's discussion I assume line 92 would change to allow for it to live some place else, right?

I don't think so: The .well-known URI needs to be first discovered. But when you retrieve it, then you can point to the actual xRegistry server implementation(s) and their URLs. They could indeed live on a different server.

core/spec.md
Fannon:
```suggestion
It allows for support of multiple versions of the xRegistry protocol at the same time
```

core/spec.md
Fannon: yes. I actually wasn't really sure if we should include it or not. I put it in, so it can be discussed.

Fannon:
> @Fannon what's the status of this work?

@duglin : I think we haven't fully decided whether to do this or not. Something we can pick up in a meeting and make a decision on? If yes, the proposal here for sure requires some rework and we also need to decide which "capabilitites" we currently have that we want to describe (you would know best, as you implemented them in code).

Fannon:
Closing this, as agreed in meeting on 2024-12-16

It could be that several xRegistry servers are hosted on the same base URL, and we don't have access to the root of the server to create a shared .well-known to point to several registry endpoints
The entry point of the xRegistry can already hold attributes that describe the capabilities of the xRegistry Server or of the model

PR #129, opened 2024-06-12
Changes for week of June 3

duglin:
- s/this/request/ on ?setdefaultversionid=vID when vID isn't known in advance
- s/?inline/?nested/ for write operations
- added an example to explain how to deal with potential conflicts between Resource's "default" attributes and that version's attributes in the `versions` collection on a PUT?nested op

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -157,7 +157,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   # Repeat for each Group type
   "GROUPsurl": "URL",                              # e.g. "endpointsurl"
   "GROUPscount": UINTEGER,                         # e.g. "endpointscount"
-  "GROUPs": {                                      # Only if inlined
+  "GROUPs": {                                      # Only if inlined/nested
     "ID": {                                        # Key=the Group id
       "id": "STRING",                              # The Group ID
       "name": "STRING", ?
@@ -173,7 +173,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
       # Repeat for each Resource type in the Group
       "RESOURCEsurl": "URL",                       # e.g. "definitionsurl"
       "RESOURCEscount": UINTEGER,                  # e.g. "definitionscount"
-      "RESOURCEs": {                               # Only if inlined
+      "RESOURCEs": {                               # Only if inlined/nested
         "ID": {                                    # The Resource id
           "id": "STRING",
           "name": "STRING", ?
@@ -196,7 +196,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
 
           "versionsurl": "URL",
           "versionscount": UINTEGER,
-          "versions": {                            # Only if inlined
+          "versions": {                            # Only if inlined/nested
             "ID": {                                # The Version id
               "id": "STRING",
               "name": "STRING", ?
@@ -444,7 +444,7 @@ The definition of each attribute is defined below:
   - MUST be a non-empty string consisting of [RFC3986 `unreserved`
     characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)
     (ALPHA / DIGIT / "-" / "." / "_" / "~").
-  - Versions MUST NOT use an `id` value of `null` or `this` due to these
+  - Versions MUST NOT use an `id` value of `null` or `request` due to these
     being reserved for use by the `setdefaultversionid` operation.
   - MUST be case insensitive unique within the scope of the entity's parent.
     In the case of the `id` for the Registry itself, the uniqueness scope will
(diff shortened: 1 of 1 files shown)


core/spec.md
```diff
@@ -531,7 +531,7 @@ Then the existing entity can be deleted.
   appended to its `id` if the request asked for the serialization of the
   xRegistry metadata. This would happen when `$meta` was used in the request,
   or when the Resource (or Version) is included in the serialization of a
-  Group, such as when the `inline` feature is used.
+  Group and the the `inline` feature is used.
```
jskeet, line 534: Nit: the the => the

core/spec.md
```diff
@@ -847,45 +849,65 @@ In API view:
...
 the server MUST process each entity in the collection map as a request to
-create or update that entity based on whether it already exists (as determined
-by its key value - its `id`). An entry in the map that isn't a valid entity
-(e.g. is `null`) MUST generate an error.
+create or update that entity according to the semantics of the HTTP method used.
+An entry in the map that isn't a valid entity (e.g. is `null`) MUST generate
+an error.
+
+The `nested` semantics MUST be applied to all nested `COLLECTIONs` attributes
+within the request regardless of the level in the hierarch in which they appear.
```
jskeet, line 871: hierarch => hierarchy

core/spec.md
```diff
@@ -847,45 +849,65 @@ In API view:
...
 Any error while processing a nested collection entity MUST result in the entire
 request being rejected.
 
 An absent `COLLECTIONs` attribute MUST be interpreted as a request to not
-modify the collection at all due to processing of the request, regardless of
-the presence (or absence) of the `inline` query parameter.
+modify the collection at all, regardless of the presence (or absence) of the
+`nested` query parameter.
+
+If a client wishes to replace an entire collection, rather than just add new
```
jskeet, line 899: So this means it can't be done atomically, right?

core/spec.md
```diff
@@ -847,45 +849,65 @@ In API view:
...
+entities, the client MUST use one of the `DELETE` operations on the collection
+first.
+
+In cases where an update operation includes attributes meant to be applied
```
jskeet, line 903: Is it worth an example of this? I'm having a hard time following it.

core/spec.md
```diff
@@ -4293,11 +4317,8 @@ HTTP/1.1 204 No Content
 
 ### Inlining
 
-The `inline` query parameter MAY be used on requests or responses to indicate
-how nested collections or certain (potentially large) attributes are to be
-handled.
-
-#### Inlining GET Responses
+The `inline` query parameter MAY be used on requests to indicate how nested
```
jskeet, line 4320: Is this now *only* valid on read requests, not put/post/patch/etc? If so, I think it would be good to clarify that. (Or maybe clarify the difference between inline and nested anyway...)

core/spec.md
duglin: fixed

core/spec.md
duglin: fixed

core/spec.md
duglin: Correct. Do you see an issue with that? We were trying to keep it simple but if you think it causes pain we could revisit this decision.

jskeet: My lack of knowledge of use cases means it's hard to say for sure. But I can imagine it *potentially* being a problem. Probably something worth just checking with other folks.

core/spec.md
duglin: ok I added one - see what you think

core/spec.md
duglin: ok clarification added

core/spec.md
duglin: On the positive side, if we wanted to support it I think it would be via a PUT to the collection, which as of now results in an error, so we could use that method as a "complete collection replacement". But we'll discuss!

core/spec.md
```diff
@@ -847,45 +848,100 @@ In API view:
...
+If the `versions` collection were not present with the `v1` entity, or if the
+`nested` query parameter was not provided, then the top-level attributes would
+be used to update the default Version (`v1` in this case). However, because
+they are present, the request to update `v1` becomes ambiguous because it is
+not clear if the server is meant to use the top-level attributes or if it
+is to use the attributes under the `v1` entity of the `versions` collection.
+When both sets of attributes are the same then it does not matter. However, in
+this cases the `name` attributes have different values. The paragraph above
+mandates that in these potentially ambiguous cases the entity in the
+`versions` collection is to be used and the top-level attribute are to be
```
jskeet, line 943: attribute => attributes

core/spec.md
jskeet: That looks good. I only follow it "mostly" because I still get confused about xRegistry in general, and I've got other things taking most of my focus right now - but it's definitely better than not having it :)

jskeet (approved):

core/spec.md
jskeet: Definitely better, thanks :)

core/spec.md

duglin:
Approved on the 6/13 call

PR #133, opened 2024-06-25
Changes for week of June 25

### LINKED ISSUE #112: Cross referencing Resources
Do we want to support the idea of one Resource being part of more than one Group? Basically define a Group to include some kind of $ref that's a URL to a Resource in some other group (or even xRegistry) as a way to include it in this other group?
If so, how will filtering work?

@deissnerk do I have this right?

duglin:
- add support for xref (cross referencing Resources
- add support for ximport - sharing of Resource definitions
- drop support for $import (the old "include-like" feature)
- add support for ?export query parameter
- add support for `exportrequired` flag in model

Fixes #112

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -114,6 +114,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
         "immutable": BOOLEAN, ?         # Once set, can't change. Default=false
         "clientrequired": BOOLEAN, ?    # Default: false
         "serverrequired": BOOLEAN, ?    # Default: false
+        "exportrequired": BOOLEAN, ?    # Default: false
         "default": VALUE, ?             # Attribute's default value, scalars
 
         "attributes": { ... }, ?        # If "type" above is object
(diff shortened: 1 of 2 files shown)


core/spec.md
```diff
@@ -223,6 +227,15 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
...
 ```
 
+Note: the cardinality/optionality of a Resource's attributes are specified
+to indicate which are to be present when the Resource's attributes are
+included in its serialization. For example, when `?export` is used, the
+Resource's default attributes will not be present at all, which means listing
+`epoch` as mandatory in the above serialize is technically incorrect in
+general, but when the attributes do appear then `epoch` is mandatory. Since
+listing all attributes as OPTIONAL would reduce the usefulness of the above
+example, this inconsistency is done on purpose.
```
jskeet, line 237: "done on purpose" => "deliberate"? (Just sounds a bit more spec-like...)

jskeet (commented):
Haven't had time to review most of this, but one comment from where I got to.

core/spec.md

duglin:
Approved on the 6/27 call

PR #149, opened 2024-07-30
WIP - Introducing URI scheme `xreg` 

deissnerk:
These changes introduce a URI scheme `xreg` that can be used to identify xRegistry entities independent from their storage location.

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -10,13 +10,16 @@ automation and tooling.
 
 - [Overview](#overview)
 - [Notations and Terminology](#notations-and-terminology)
-  - [Notational Conventions](#notational-conventions)
-  - [Terminology](#terminology)
+    - [Notational Conventions](#notational-conventions)
+    - [Terminology](#terminology)
 - [Registry Attributes and APIs](#registry-attributes-and-apis)
   - [Attributes and Extensions](#attributes-and-extensions)
+    - [Common Attributes](#common-attributes)
+    - [`xid` - Referencing xRegistry Entities](#xid---referencing-xregistry-entities)
   - [Registry APIs](#registry-apis)
     - [Registry Collections](#registry-collections)
     - [Entity Processing Rules](#entity-processing-rules)
+  - [`xid` Resolution API](#xid-resolution-api)
   - [Registry Entity](#registry-entity)
     - [Retrieving the Registry](#retrieving-the-registry)
     - [Updating the Registry Entity](#updating-the-registry-entity)
@@ -437,7 +440,9 @@ attributes. However they MUST adhere to the following rules:
   example, use of a model (or domain) specific prefix could be used to help
   avoid possible future conflicts.
 
-#### `xid` - Referencing xRegistry Entities
+#### `xid`
+
+##### Referencing xRegistry Entities
 
 An `xid` is a URI according to
 [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) that
@@ -448,8 +453,9 @@ is helpful to use a URI, e.g. events or knowledge graphs.
 Each entity has exactly one `xid`. Therefore, comparing two `xid`
 values is possible to determine, if they identify the same or different
 entities. To achieve this, an xRegistry-specific URI scheme `xreg` is defined:
+ABNF https://datatracker.ietf.org/doc/html/rfc5234
 
-`xid` = `"xreg://" authority "/" entity-path "/" entity-id`
+`xid` = `"xreg://" authority "/" entity-path`
 
 `entity-path` = `group-path` / `resource-path` / `version-path`
 
@@ -469,30 +475,30 @@ Where:
 - `rID` is the `id` of a single Resource.
 - `vID` is the `id` of a single Version of a Resource.
 
-##### `authority`
+###### `authority`
 
 The `authority` is a string that uniquely identifies the entity's authority.
-It could be regarded as a namespace and SHOULD be based on a domain name that is
-under the control of the entity's publisher.
+It could be regarded as a namespace and SHOULD be based on a unique name
+like an entry in the domain name system (DNS) that is under the control of
+the entity's publisher.
 
-##### `entity-path`
+###### `entity-path`
 
 As xRegistry entities are organized in a hierarchy, the `entity-path`
 outlines the entity's position in the hierarchy. As the ABNF above shows,
 there can be paths for groups, resources, and versions.
 
-##### `entity-id`
-
 ##### De-referencing a relative URI reference
 
 A relative URI reference can only be de-referenced if the base URI is known
-from the context. 
+from the context.
 
-If the relative URI reference consists of a relative path segment, i.e. it 
-does not begin with a slash, then the base URI is the URI of the surrounding 
+If the relative URI reference consists of a relative path segment, i.e. it
+does not begin with a slash, then the base URI is the URI of the surrounding
 entity.
 
 __Example:__
+
 ```
 xid of surrounding entity: xreg://example.com/messagegroups/group2
 
@@ -501,11 +507,12 @@ Reference: messages/message1
 De-referenced URI: xreg://example.com/messagegroups/group2/messages/message1
 ```
 
-If the relative URI reference consists of an absolute path segment, i.e. it 
-begins with a slash, then the base URI is the authority of the surrounding 
+If the relative URI reference consists of an absolute path segment, i.e. it
+begins with a slash, then the base URI is the authority of the surrounding
 group.
 
-__Example:__ 
+__Example:__
+
 ```
 xid of surrounding group: xreg://example.com/messagegroups/group2
 
@@ -516,27 +523,32 @@ De-referenced URI: xreg://example.com/messagegroups/group1/messages/message1
 
 ```
 
-##### Resolving an `xid`
+##### Binding an `xid`
 
-As an `xid` is an abstract identifier of an entity, it is necessary to 
+As an `xid` is an abstract identifier of an entity, it is necessary to
 transform it into URLs or URL references to access the entity.
 
-For the HTTP API, the information from an `xid` can be used to construct a 
+For the HTTP API, the information from an `xid` can be used to construct a
 URL like this:
 
 `<Registry-base-URL>/GROUPs/{authority}${gID}/RESOURCEs/{rID}/versions/{vID}`
 
-The authority is OPTIONAL if it is identical to the registry's default 
-authority. For a file-based registry, the URL looks similar but makes use of 
-a URL fragment section: 
+The authority is OPTIONAL if it is identical to the registry's default
+authority. For a file-based registry, the URL looks similar but makes use of
+a URL fragment section:
 
 `<File-URL>/#GROUPs/{authority}${gID}/RESOURCEs/{rID}/versions/{vID}`
 
-Bindings to HTTP, AMQP, registry as document  etc.
+Further bindings MAY be defined in the future, e.g. for AMQP or GraphQL.
 
 ##### Comparing `xid` values
 
+A key feature of the `xid` is that it is a URI and can be efficiently compared
+with other `xids`:
 
+__Two `xids` are equal if and only if their authority, their group type name,
+their resource type name (if applicable), and all ID fields they contain are 
+equal.__
 
 #### Common Attributes
 
@@ -588,9 +600,9 @@ The definition of each attribute is defined below:
     be generated.
   - MUST be immutable.
 - Examples:
-  - `a183e0a9-abf8-4763-99bc-e6b7fcc9544b`
-  - `myEntity`
-  - `myEntity.example.com`
+    - `a183e0a9-abf8-4763-99bc-e6b7fcc9544b`
+    - `myEntity`
+    - `myEntity.example.com`
 
 While `SINGULARid` can be something like a UUID, when possible, it is
 RECOMMENDED that it be something human friendly as these values will often
@@ -657,10 +669,10 @@ of the existing entity. Then the existing entity can be deleted.
   might result in an epoch validation error for the second update request
   mention above.
 - Constraints:
-  - MUST be an unsigned integer equal to or greater than zero.
-  - MUST increase in value each time the entity is updated.
+    - MUST be an unsigned integer equal to or greater than zero.
+    - MUST increase in value each time the entity is updated.
 - Examples:
-  - `0`, `1`, `2`, `3`
+    - `0`, `1`, `2`, `3`
 
 ##### `name` Attribute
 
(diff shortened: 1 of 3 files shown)


core/spec.md
```diff
@@ -436,6 +445,114 @@ attributes. However they MUST adhere to the following rules:
...
+entity.
+
+__Example:__
+
+```
+xid of surrounding entity: xreg://example.com/messagegroups/group2
+
+Reference: messages/message1
+
+De-referenced URI: xreg://example.com/messagegroups/group2/messages/message1
```
duglin, line 511: What's funny is that this doesn't follow normal browser processing rules. A browser would look for .../messagegroups/messages/message1

core/spec.md
```diff
@@ -436,6 +445,114 @@ attributes. However they MUST adhere to the following rules:
...
+
+##### Binding an `xid`
+
+As an `xid` is an abstract identifier of an entity, it is necessary to
+transform it into URLs or URL references to access the entity.
+
+For the HTTP API, the information from an `xid` can be used to construct a
+URL like this:
+
+`<Registry-base-URL>/GROUPs/{authority}${gID}/RESOURCEs/{rID}/versions/{vID}`
```
duglin, line 538: Why is "{authority}" after GROUPs?

core/spec.md
```diff
@@ -436,6 +445,114 @@ attributes. However they MUST adhere to the following rules:
...
+`<File-URL>/#GROUPs/{authority}${gID}/RESOURCEs/{rID}/versions/{vID}`
+
+Further bindings MAY be defined in the future, e.g. for AMQP or GraphQL.
+
+##### Comparing `xid` values
+
+A key feature of the `xid` is that it is a URI and can be efficiently compared
+with other `xids`:
+
+__Two `xids` are equal if and only if their authority and all ID fields they
```
duglin, line 553: It's more than just IDs that need to match, they need to match the collection names too.

core/spec.md
```diff
@@ -436,6 +445,114 @@ attributes. However they MUST adhere to the following rules:
...
+[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) that
+uniquely identifies an entity within an xRegistry. Its purpose is to
+reference the entity and express a relation to it, even if it is not stored in
+the local registry. The `xid` can also be used in other places where it
+is helpful to use a URI, e.g. events or knowledge graphs.
+Each entity has exactly one `xid`. Therefore, comparing two `xid`
+values is possible to determine, if they identify the same or different
+entities. To achieve this, an xRegistry-specific URI scheme `xreg` is defined:
+
+`xid` = `"xreg://" authority "/" entity-path "/" entity-id`
```
duglin, line 460: Looking at some examples:
```
xreg://example.com/schemagroups/g1/schemas/s1/versions/v1
xreg://example.com/schemagroups/g1/schemas/s1
xreg://example.com/schemagroups/g1
xreg:///schemagroups/g1
```
Is there a way to reference the registry itself? e.g. `xreg:///regID` or `xreg:////regID` (a lot of /'s) ? Do we think that might be needed in the future? Seems like it.

Does `xreg://example.com/` mean anything? I guess it could be a registry, but w/o an ID you can only have one per authority which doesn't feel right. Would we need to include the registry ID only when there's an authority?

we may want to say that the format used here is BNF (or eBNF).

core/spec.md
```diff
@@ -436,6 +445,114 @@ attributes. However they MUST adhere to the following rules:
...
+- `GROUP`, not shown, is the singular name of a Group type.
+- `gID` is the `id` of a single Group.
+- `RESOURCEs` is a Resource type name (plural). e.g. `definitions`.
+- `RESOURCE`, not shown, is the singular name of a Resource type.
+- `rID` is the `id` of a single Resource.
+- `vID` is the `id` of a single Version of a Resource.
+
+##### `authority`
+
+The `authority` is a string that uniquely identifies the entity's authority.
```
duglin, line 482: Valid chars?

core/spec.md
duglin: I kind of expected something like: `<baseURL>/AUTHORITY$GROUPs/gID/RESOURCEs/rID/versions/vID`

duglin:
Would it make sense to merge "authority" with "origin"? I was thinking it might be nice if xid could be calculated from "self" and "authority/origin"

core/spec.md
deissnerk: This would essentially introduce a namespace concept at the level of the registry, while my proposal choses the group level. This is of course a valid question to discuss. If we replicate metadata, do we expect to separate it on registry level, or would a user prefer to browse all resources (replicated and local) as part of a single registry?

core/spec.md
deissnerk: I agree. I will adjust the text accordingly.

core/spec.md
duglin: An idea from the 10/8 call:
```
<baseURL>/GROUPs/gID@AUTHORITY/RESOURCEs/rID/versions/vID
```

core/spec.md
deissnerk: Good question. If someone decides to use a valid domain name, it is restricted by the domain name system, but that is not necessarily a constraint for all `authorities` in xRegistry. I suppose it would make sense to apply the same rules we have for RESOURCE/GROUP/versionid. WDYT? @duglin

core/spec.md
```diff
@@ -14,9 +14,12 @@ automation and tooling.
   - [Terminology](#terminology)
 - [Registry Attributes and APIs](#registry-attributes-and-apis)
   - [Attributes and Extensions](#attributes-and-extensions)
+    - [Common Attributes](#common-attributes)
+    - [`xid`](#xid)
   - [Registry APIs](#registry-apis)
     - [Registry Collections](#registry-collections)
     - [Entity Processing Rules](#entity-processing-rules)
+  - [`xid` Resolution API](#xid-resolution-api)
```
duglin, line 22: any reason this isn't indented one more since it's still a "Registry API", right?

core/spec.md
```diff
@@ -436,6 +440,117 @@ attributes. However they MUST adhere to the following rules:
   example, use of a model (or domain) specific prefix could be used to help
   avoid possible future conflicts.
 
+#### `xid`
+
+##### Referencing xRegistry Entities
+
+An `xid` is a URI according to
+[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) that
+uniquely identifies an entity within an xRegistry. Its purpose is to
```
duglin, line 449: s/an Registry/a Registry/

Also, is it just "within a Registry" or is it mean to be more globally identifying? I thought it was for cross-registry identification too

core/spec.md
```diff
@@ -436,6 +440,117 @@ attributes. However they MUST adhere to the following rules:
...
+reference the entity and express a relation to it, even if it is not stored in
```
duglin, line 450: "express a relation to it"... from where?

core/spec.md
```diff
@@ -436,6 +440,117 @@ attributes. However they MUST adhere to the following rules:
...
+the local registry. The `xid` can also be used in other places where it
+is helpful to use a URI, e.g. events or knowledge graphs.
+Each entity has exactly one `xid`. Therefore, comparing two `xid`
+values is possible to determine, if they identify the same or different
+entities. To achieve this, an xRegistry-specific URI scheme `xreg` is defined.
+Below, [ABNF](https://datatracker.ietf.org/doc/html/rfc5234)  notation is used
+to describe how `xids` are structured:
+
+`xid` = `"xreg://" authority "/" entity-path`
```
duglin, line 459: based on the new "@" stuff, is this still correct?

duglin: And, doesn't "authority" need to be defined either in the BNF or in the WHERE stuff below?

core/spec.md
```diff
@@ -436,6 +440,117 @@ attributes. However they MUST adhere to the following rules:
...
+xid of surrounding entity: xreg://example.com/messagegroups/group2
+
+Reference: messages/message1
+
+De-referenced URI: xreg://example.com/messagegroups/group2/messages/message1
+```
+
+If the relative URI reference consists of an absolute path segment, i.e. it
+begins with a slash, then the base URI is the authority of the surrounding
+group.
```
duglin, line 513: I think mentioning "group" here could be misleading, perhaps just "is the authority of the base URI" ?

core/spec.md
```diff
@@ -436,6 +440,117 @@ attributes. However they MUST adhere to the following rules:
...
+
+__Example:__
```
duglin, line 515: We use:
```
**Examples:**
```
for the header of example in other spots

core/spec.md
```diff
@@ -436,6 +440,117 @@ attributes. However they MUST adhere to the following rules:
...
+For the HTTP API, the information from an `xid` can be used to construct a
+URL like this:
+
+`<Registry-base-URL>/GROUPs/{gID}@{authority}/RESOURCEs/{rID}/versions/{vID}`
+
+The authority is OPTIONAL if it is identical to the registry's default
+authority. For a file-based registry, the URL looks similar but makes use of
+a URL fragment section:
+
+`<File-URL>/#GROUPs/{gID}@{authority}/RESOURCEs/{rID}/versions/{vID}`
```
duglin, line 541: is the "#" needed?

core/spec.md
```diff
@@ -436,6 +440,117 @@ attributes. However they MUST adhere to the following rules:
...
+- `gID` is the `id` of a single Group.
+- `RESOURCEs` is a Resource type name (plural). e.g. `definitions`.
+- `RESOURCE`, not shown, is the singular name of a Resource type.
+- `rID` is the `id` of a single Resource.
+- `vID` is the `id` of a single Version of a Resource.
+
+###### `authority`
+
+The `authority` is a string that uniquely identifies the entity's authority.
+It could be regarded as a namespace and SHOULD be based on a unique name
```
Fannon, line 482:
```suggestion
It could be regarded as a namespace and MUST be unique within the registry and SHOULD be globally unique, based on a unique name
```

core/spec.md
```diff
@@ -1263,6 +1379,72 @@ HTTP body.
...
+If the applied request method is `GET`, the body of the response MUST be a
+JSON object as follows:
+
+```yaml
+{
+  "xid": "URI",     # The xid to resolve
+  "locations": [
+    {
+      "url": "URL",   # The URL to access the entity
+      "authoritative": boolean
```
duglin, line 1424: s/boolean/BOOLEAN/

What does "authoritative" mean in this context?

core/spec.md
```diff
@@ -1263,6 +1379,72 @@ HTTP body.
...
+
+The placeholder `{xid}` depicts the `xid` to resolve. As `xids` MAY be full
+URIs or URI references, the `xid` MUST be percent-encoded accordingly.
+
+For `GET` and `HEAD` requests, the server MUST respond as follows:
+
+- If the `xid` is not found, the server MUST respond with an HTTP
+  `404 Not Found`.
+- If the `xid` is found and there exists exactly one URL to access the entity,
+  the server MUST respond with an HTTP `308` where the `Location` header
```
duglin, line 1402: I'm not sure 308 is correct because what if the resolver service wants to actually redirect to a new location for the resolver service? I wonder if just returning a 200 is good enough.

core/spec.md
```diff
@@ -1263,6 +1379,72 @@ HTTP body.
...
+  MUST contain the URL to access the entity.
+- If the `xid` is found and there are multiple URLs to access the entity, the
+  server MUST respond with an HTTP `300 Multiple Choices` where the choices
```
duglin, line 1405: Like above, I'm not sure a 300 (or 3xx in general) is correct since I don't think we're doing a "redirect" in the HTTP sense. I think we're calling a function and the result is one or more URLs - that's not a redirect, that's a response. We can then choose how the response looks (headers vs body) in the single vs multiple cases.

core/spec.md
```diff
@@ -1263,6 +1379,72 @@ HTTP body.
...
+    } *
+  ] ?
+}
+```
+
+#### `locations`
+
+The `locations` array contains the URLs to access the entity. In addition to
+the `url`, the `authoritative` attribute indicates whether the URL points to
+the registry that is authoritative for the entity.
```
duglin, line 1440: I think you need to define "authoritative" w/o using the word "authoritative". Do you really mean that the response can be trusted (https://www.merriam-webster.com/dictionary/authoritative)? Or do you mean that this is the "original source of truth" for this entity (meaning, all others are just aliases/pointers) ?

core/spec.md
```diff
@@ -2748,7 +2944,20 @@ Groups include the following common attributes:
...
+##### `authority` Attribute
+
+- Type: String
+- Description: The authority under which this group is defined. If not present,
+  the according `defaultauthority` will be applied.
+- Constraints:
+  - OPTIONAL.
+  - If present, MUST be non-empty.
+  - MUST be a valid authority as defined under [authority](#authority).
+  - If not present, the `defaultauthority` will be applied, when the group
```
duglin, line 2956: I'd avoid repeating this sentence - I'd drop the one on line 2951.

core/spec.md
deissnerk: I adjusted this PR accordingly, but we had further discussions about authorities for models **and** data. I captured some notes about that discussion here: #181

core/spec.md
deissnerk: Yes, for formal correctness, we will need more ABNF rules. When we have concluded on #181 I will add them.

core/spec.md
deissnerk: The "#" corresponds to what we currently have for files. I'm not sure I understand the question.

core/spec.md
deissnerk: Well, we can discuss this. I thought it might be neat if even a standard web browser could be used for this. I could imagine returning 200 and the JSON in case of a GET. We could then do the redirect only for the HEAD call. Behaving like a redirect by returning the according headers without actually being one looks strange to me. If we use 30x, a web browser would automatically retrieve the entity. Someone who is aware of xRegistry, could still decide to ignore the 30x. 
The redirect approach only works for queries that specify an `xid`. For calls that might return data sets for multiple xids, a redirect doesn't make sense.

core/spec.md
deissnerk: Always plural, even if it is just one?

core/spec.md
deissnerk: I'm not sure it's a registry API. It may also be offered independently. When we work on the overall document structure we can revisit this.

core/spec.md
duglin: I was more focused on the `*` vs `_` :-)

deissnerk: Ok, that part was clear. Just wanted to be sure not to miss anything. 😄

duglin:
@deissnerk should we close this?

duglin:
ping @deissnerk

duglin:
closing per our previous discussions - we can re-open as needed

PR #155, opened 2024-08-19
Week of Aug19 changes

### LINKED ISSUE #148: model's Resource.readonly only works for types, not instances
Setting Resource.readonly to 'true' is fine if all instances of that Resource are read-only, but what if only certain instances are? Or what if just certain Group instances have read-only resources?

Proposal:
- Remove the "readonly" aspect from the model
- Define a "readonly" attribute (bool) for Resources
  - Defaults to "false"
  - system managed, "true" = user can't touch

duglin:
- fix "attributes" in "messages" model for CE format
- s/schemauri/schemaurl/ for now, to align with the spec
- remove 'readonly' attribute from model Resources
- add 'readonly' attribute to Resource instances - server controlled
- s/time/timestamp/ where appropriate
- update codegen:
  - s/createdon/createdat/
  - s/modifiedon/modifiedat/
  - add "origin" as a common attribute
  - remove "format" as a common attribute

Fixes #148

### DIFF
--- a/cloudevents/model.json
+++ b/cloudevents/model.json
@@ -1,0 +1,12 @@
+{
+  "schemas": [
+    "xRegistry-json"
+  ],
+  "groups": {
+    "$includes": [
+      "../endpoint/model.json#groups",
+      "../message/model.json#groups",
+      "../schema/model.json#groups"
+    ]
+  }
+}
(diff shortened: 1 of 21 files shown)


duglin:
Approved on the 8/22 call

PR #161, opened 2024-08-22
Add primer

### LINKED ISSUE #140: Create primer
Before creating a PR with the full text, I wanted to collect some bullet points and get your ideas. This is how I would approach this Primer:

* CloudEvents are harmonizing different event structures and increase interoperability, but they do not provide an answer to the following questions:
  * Where can I produce / consume events? What endpoints exist?
  * How can I validate events?
  * How can I version events - and see which versions exist?
  * If I specified extensions – how can I know which events are using which extensions?
* There are many schema registries out there, like:
  * Confluent Schema Registry, Azure Event Hubs Registry, Apicurio Registry, Solace..
* If an event travels through multiple brokers, it's schema has to be introduced in the registry of each broker. Thanks to CloudEvents the schema stays identical and is quite independent from the broker & protocol, but clients still have to mind the implementation differences between the different registries per broker.
  * e.g. if they are trying to validate an event

* xRegistry therefore comes in with a similar motivation as CloudEvents: Harmonize another piece of technology in the messaging / eventing ecosystem to increase interoperability and detach event handling from broker products and protocols.
* xRegistry stands for extensible registry and in its core provides metadata about resources. That does not only sound very abstract, it actually is hence its extensibility.
* As an example, you could have endpoints, message definitions and schemas.
  * *I would then go on and showcase how this sample registry would solve the questions I raised in the beginning before making a transition to the spec.*

I vaguely remember someone saying that the motivation behind xRegistry was not only to describe CloudEvents but that you had a broader use case in mind, I do not remember which one that was though. So if there are key points missing to this narrative, please mention them in the comments.

manuelottlik:
Fixes #140 

## Proposed Changes

- Create primer
-
-

**Release Note**

```release-note
Create a first version of the primer
```


core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+- [Possible Use Cases](#possible-use-cases)
+- [Design decisions or topics of
+  interest](#design-decisions-or-topics-of-interest)
+
+## History
+
+The CNCF Serverless Working group was originally created by the CNCF's Technical
+Oversight Committee to investigate Serverless Technology and to recommend some
+possible next steps for some CNCF related activities in this space. After
+creating the common event format [CloudEvents](https://github.com/cloudevents)
```
duglin, line 28: I wouldn't say "comment event format", while we do have the structured mode, the idea of a CEF (meaning, one event shape to rule the all) has failed many times and that's not really what CE is. It's just about "common metadata, in a well defined location". I'd just leave it as "After creating [CloudEvents]..."

core/primer.md
```diff
@@ -2,6 +2,185 @@
 
 <!-- no verify-specs -->
 
+## Abstract
+
+This non-normative document provides an overview of the xRegistry specification.
```
duglin, line 7: It might be nice to turn one of the "xRegistry" references in here (perhaps the first one) into a link to the spec itself.

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+Oversight Committee to investigate Serverless Technology and to recommend some
+possible next steps for some CNCF related activities in this space. After
+creating the common event format [CloudEvents](https://github.com/cloudevents)
+as a foundation for an interoperable ecosystem for event-driven applications the
+next step was to create a metadata model for declaring CloudEvents, their
+payloads and associating those declarations with application endpoints. As a
+result, the xRegistry (extensible registry) specification was created.
+
+xRegistry was initially part of CloudEvents, called "CloudEvents Discovery" but
+later moved into it's own repository and CNCF project.
```
duglin, line 35: I'd drop "CNCF" for now. While yes it's a project under the CNCF, phrasing it this way makes it sound like an incubator project, or something more official. As of now it's more like a side-project and at some point we'll need to bring it up for formal status within the CNCF.

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+
+## Motivation
+
+While CloudEvents are harmonizing different event structures across event
+providers and increase interoperability between different brokers and their
+protocols, they do not address:
+
+- **Discovery:** Where to produce/consume events? What endpoints exist?
+- **Validation:** How to validate event structures?
+- **Versioning:** How to version events – and see which versions exist?
```
duglin, line 45: I wonder if we need to reword this slightly since we don't really "version events" as much as we "version the metadata describing the events" (e.g. their schema). It's almost the same, but I wouldn't want people to think that we're versioning the instances of events.

We should probably add something about how to figure out events are serialized - their envelope (if any and formats (xml, json & schema).

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+registry and clients later have to deal with the implementation differences
+between different registries, for example when trying to validate an event
+structure. This hinders the interoperability of event brokers CloudEvents had in
+mind. xRegistry therefore comes in with a similar motivation to CloudEvents:
+Harmonize another piece of technology in the messaging / eventing ecosystem to
+increase interoperability and decouple event handling from broker products and
+protocols.
+
+Even though xRegistry was built with eventing in mind, the concept of this
+registry specification is far more powerful. Take a look at the [Possible Use
```
duglin, line 62: I'd give a hint here about how it goes beyond eventing... perhaps:
..is far more powerful and can be used for the exchange of any type of message.

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+
+Whenever registries get large, you will realistically want to use the API server
+or CLI to create the registry and then use its export function to generate files
+for the file or static file server representation.
+
+### File
+
+The registry is represented as a single JSON file.
+
+The primary use case for this representation is a registry which purpose is to
```
duglin, line 104: s/which/whose/

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+describe a single project. It basically becomes a declaration manifest of the
```
duglin, line 105: I'm not sure what the right word is but I don't think it's "project" since an xReg could describe many different projects (e.g. services). Perhaps it would be better to focus on why people would use "file" rather than what's in the registry. E.g. they want to work off-line, or check it into git (like you mention next), or for optimization when running the info thru tooling (e.g. one GET instead of lots) ....   since I think people would choose each of the 3 choices based on reasons other than "what's in the registry", and more because of how they want to access (or host) the data.

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+application and can therefore be used in public git repositories or your own
+projects. If you are new to xRegistry and are just trying to get the idea of the
+project, this is a great point to start.
+
+Writing up a registry by hand can be a great way of getting the idea, but it can
+get tedious quickly. This is when you want to use the API server.
+
+### Static File Server
+
+The registry is represented by multiple JSON files served via a static file
```
duglin, line 115: perhaps: ... by multiple JSON files, where each one represents a single entity within the registry, served via a...

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+server (e.g. S3 or similar) that follows the folder and file structure of the
+API server.
+
+The primary use case for this representation is the transition between File and
+API Server representation. When a single file gets too large or you want to
+share your resources beyond the scope of a single application, splitting up your
```
duglin, line 121: ...to share individual entities of the registry rather than the entire registry as a single JSON document, splitting up your...

core/primer.md
```diff
@@ -2,6 +2,185 @@
...
+### CloudEvents
+
+Since xRegistry emerged from the CloudEvents project and initially was called
+"CloudEvents Discovery" the obvious use case is to manage all the CloudEvents of
+your event-driven architecture inside xRegistry. The spec supports "CloudEvents"
+as a format and allows you to define further restrictions on how your
+CloudEvents look, what extensions they use or which fields are required for your
+specific use-case even though the original spec marks them as optional. In
+addition, the `dataschema` attribute of a CloudEvent can then point to a
+xRegistry endpoint making this two projects that work hand in hand.
```
duglin, line 167: s/endpoint/schema document/

duglin:
Overall LGTM! and thanks!
Just a few minor suggestions.

core/primer.md
manuelottlik: I rephrased Versioning and added Serialization, does that match your expectations?

core/primer.md
manuelottlik: Is this the change you intended?

core/primer.md
manuelottlik: I changed the paragraph, is this what you were looking for?

core/primer.md
duglin: yup that works - thanks

core/primer.md
duglin: yes - thanks!

core/primer.md
duglin: yup thanks

duglin:
still LGTM!

duglin:
Approved on the 8/29 call

PR #162, opened 2024-08-25
Changes for the week of Aug 26

duglin:
- make `epoch` global to the Resource
- check `versionid` on update when updating the Resource's default Version props
- add `location` to Resource model attributes (resource, version, both)
- removed `exportrequired`
- added back in `$import` (as `$include`)
- added `?resource` to exclude the default Version attributes on a Resource
- added `/export` - same as `/?export` - for static file server cases

### DIFF
--- a/core/sample-model.json
+++ b/core/sample-model.json
@@ -218,7 +218,7 @@
             }
 
             "defaultversionsticky": {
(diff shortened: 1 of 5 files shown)


duglin:
Approved on the 8/29 call

PR #163, opened 2024-09-03
Changes for week of Sept 3

duglin:
- Updated message, endpoint and schema specs to latest common attributes
- lots of clean-up/catch-up


tools/dict
```diff
@@ -114,24 +186,39 @@ unicode
...
+www
 xml
+xmlschema
+xpath
 xreg
 xregbasicmodel
 xregfullmodel
 xregistries
 xregistry
+xregsitry
```
jskeet, line 220: Really?

jskeet: (These are typos on lines 218 and 293 of the new version of schema/spec.md)

jskeet (commented):
Mostly looks fine, other than the "add a typo to dict instead of fixing it" :)

tools/dict
duglin: yea, this is why I need to you remove my stuff :-)
thanks!

jskeet (approved):
Glad to be able to be mildly useful :)

duglin:
> Glad to be able to be mildly useful :)

More than mildly... I missed your review last week

duglin:
Approved on the 9/5 call

Issue #164, opened 2024-09-05
Schema: "format" doesn't match schema doc -what to do?


duglin:
What if the schema doc in a Schema resource doesn't match the "format" attribute's value?
What if the doc at the schemaurl location doesn't?

Obviously the 1st variant is easier to detect, but what should the spec say about these situations?

Proposal:
Basically nothing. I don't think the spec should mandate any verification of the contents of the schema doc. However, implementations are free to add that feature if they want to as an extension. So we should add text saying it's out of scope for the spec to mandate it.

duglin:
Proposal:
- add a "validation" capability, so registries can advertise if they'll do validation or not, but it's not required that they do
- define what happens (which error) if validation fails
- put it on Resource level with default "true"
- only required to validate for formats that the server understands AND if format is present

Issue #159, opened 2024-08-21
Is there a way in Message to say CE binary vs structured?


duglin:
I think we need a way, no?
@clemensv

Issue #154, opened 2024-08-09
COLLECTIONscount should be STRONGLY RECOMMENDED...


duglin:
...and not REQUIRED because it might be too expensive to calculate it if all we're doing is returning the count+url and not the actual entities themselves.

Issue #153, opened 2024-08-07
Should .../rID$meta work even for Resources that have "hasdocument=false"?


duglin:
Should .../rID$meta work even for Resources with hasdoc=false? For consistency.

Today we say an error must be generated.

Also, should Registry and Groups support `$meta` too ? 

Note that GET w/o $meta on a Resource with hasdoc=false WILL NOT return xReg metadata as http headers.

Proposal: Yes for $meta on Resources w/o docs, but no for Registry/Groups.

duglin:
Approved on 9/13
- $meta when hasdoc=false returns the xReg metadata as the doc, same as w/o $meta, but...
- URLs (e.g. self) in the Resource MUST include $meta, same as if hasDoc=true

PR #167, opened 2024-09-13
Changes for week of Sep 16

duglin:
- s/format/envelope/g    - e.g. CE
- add envelopeoptions
- add envelopemetadata
- s/binding/protocol/g   - e.g. HTTP
- s/message/protocoloptions/g
- allow $meta on Resources with hasdoc=false
- make "format" optional on Resources in schema spec
- add an optional "validation" attribute to Schema Resources
- clarify how filters + inlining work (only children of leaves can be added)
- require ?nested when child/nested collections are present on write ops
    
Fixes: #153
Fixes: #154
Fixes: #151
Fixes: #164
Fixes: #132
Fixes: #168
Fixes: #159

### DIFF
--- a/cloudevents/schemas/document-schema.avsc
+++ b/cloudevents/schemas/document-schema.avsc
@@ -88,16 +88,6 @@
             },
             {
               "type": "string",
-              "name": "format",
-              "doc": "Endpoint metadata format identifier. If set, all definitions MUST use this format value"
-            },
-            {
-              "type": "string",
-              "name": "binding",
-              "doc": "Endpoint message binding identifier. If set, all definitions MUST use this binding value"
-            },
-            {
-              "type": "string",
               "name": "channel",
               "doc": "tbd"
             },
@@ -107,9 +97,93 @@
               "doc": "tbd"
             },
             {
-              "type": "GenericRecord",
-              "name": "config",
-              "doc": "Configuration information for this endpoint"
+              "name": "envelope",
+              "type": [
+                {
+                  "type": "record",
+                  "name": "EnvelopeCloudevents10Type",
+                  "fields": [
+                    {
+                      "type": "GenericRecord",
+                      "name": "envelopeoptions",
+                      "doc": "Envelope metadata constraints"
+                    }
+                  ]
+                }
+              ],
+              "doc": "Envelope metadata format identifier. If set, all definitions MUST use this format value"
+            },
+            {
+              "name": "protocol",
+              "type": [
+                {
+                  "type": "record",
+                  "name": "ProtocolAmqp10Type",
+                  "fields": [
+                    {
+                      "type": "GenericRecord",
+                      "name": "protocoloptions",
+                      "doc": "Configuration information for this endpoint"
+                    }
+                  ]
+                },
+                {
+                  "type": "record",
+                  "name": "ProtocolMqtt50Type",
+                  "fields": [
+                    {
+                      "type": "GenericRecord",
+                      "name": "protocoloptions",
+                      "doc": "Configuration information for this endpoint"
+                    }
+                  ]
+                },
+                {
+                  "type": "record",
+                  "name": "ProtocolMqtt311Type",
+                  "fields": [
+                    {
+                      "type": "GenericRecord",
+                      "name": "protocoloptions",
+                      "doc": "Configuration information for this endpoint"
+                    }
+                  ]
+                },
+                {
+                  "type": "record",
+                  "name": "ProtocolHTTPType",
+                  "fields": [
+                    {
+                      "type": "GenericRecord",
+                      "name": "protocoloptions",
+                      "doc": "Configuration information for this endpoint"
+                    }
+                  ]
+                },
+                {
+                  "type": "record",
+                  "name": "ProtocolKAFKAType",
+                  "fields": [
+                    {
+                      "type": "GenericRecord",
+                      "name": "protocoloptions",
+                      "doc": "Configuration information for this endpoint"
+                    }
+                  ]
+                },
+                {
+                  "type": "record",
+                  "name": "ProtocolNATSType",
+                  "fields": [
+                    {
+                      "type": "GenericRecord",
+                      "name": "protocoloptions",
+                      "doc": "Configuration information for this endpoint"
+                    }
+                  ]
+                }
+              ],
+              "doc": "Endpoint protocol identifier. If set, all definitions MUST use this protocol value"
             },
             {
               "type": {
@@ -215,16 +289,16 @@
                       "doc": "Reference to a base definition for this definition, either via a (relative) URL or a fragment identifier. The base definition is overridden by this definition. If not present, this definition does not override any base definition"
                     },
                     {
-                      "name": "format",
+                      "name": "envelope",
                       "type": [
                         {
                           "type": "record",
-                          "name": "FormatNoneType",
+                          "name": "EnvelopeNoneType",
                           "fields": []
                         },
                         {
                           "type": "record",
-                          "name": "FormatCloudevents10Type",
+                          "name": "EnvelopeCloudevents10Type",
                           "fields": [
                             {
                               "type": {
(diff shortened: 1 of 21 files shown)


cloudevents/schemas/document-schema.avsc
```diff
@@ -264,99 +338,132 @@
...
+                            },
+                            {
+                              "type": "GenericRecord",
+                              "name": "envelopeoptions",
+                              "doc": "Envelope metadata constraints"
                             }
                           ]
                         }
                       ],
                       "doc": "Message format identifier. This attribute MUST be the same as the 'format' attribute of the definition group, but is not automatically inherited because definitions may be cross-referenced across groups"
```
jskeet, line 352: Is this deliberately still "format"? (There are many other occurrences as well.) I'm assuming that only *some* `format` attributes have been renamed to `envelope`, but I haven't quite got my head round which yet.

schema/spec.md
```diff
@@ -270,9 +276,12 @@ core xRegistry Resource
   specification defines a set of common [schema format names](#schema-formats)
   that MUST be used for the given formats, but applications MAY define
   extensions for other formats on their own.
+
+  For many schema registry use cases this attribute might be very important for
```
jskeet, line 280:
```suggestion
  For many schema registry use cases this attribute is important for
```

There's already hedging in "For many schema registry use cases", so I don't think we need the "might be" (and "very" doesn't feel helpful to me - although that's a personal preference of course).

schema/spec.md
```diff
@@ -348,6 +357,25 @@ Versions for a schema named `com.example.telemetrydata`:
...
 ```
 
+#### `validation`
+
+- Type: Boolean
+- Description: Indicates whether or not the server will validate the Resource's
+  document(s) against the specified `format` attribute. A value of `true`
+  indicates that the server MUST reject any request that would cause any
+  Version of this Resource to be invalid per the rules as defined by the
+  `format` specification. Note, this includes a request to set this attribute
```
jskeet, line 367: I don't understand the note - is it saying "If there's a request to set this attribute to true, the server must validate the rest of the intended state of the resource document immediately"? (So you don't get to set it to true with an invalid document.)

schema/spec.md
```diff
@@ -348,6 +357,25 @@ Versions for a schema named `com.example.telemetrydata`:
...
+  to `true`.
+
+  A value of `false` indicates that the server MUST NOT do any validation.
```
jskeet, line 370: This is interesting. It feels to me like it would be useful to make "validate but don't prevent" an option - e.g. "The server mustn't reject requests causing an invalid version, but can report warnings". I don't know where it would report warnings, admittedly.

schema/spec.md
```diff
@@ -348,6 +357,25 @@ Versions for a schema named `com.example.telemetrydata`:
...
+
+  If `format` is not specified, or if the value is not known by the server
+  (but it an allowable value), then the server MUST NOT perform any validation.
```
jskeet, line 373: "any" validation feels like it's potentially asking for trouble. Would it still be allowed to reject incoming JSON which specified the same property multiple times, for example? (That's technically-valid JSON, but feels reasonable to reject.) Presumably it's *always* okay to reject requests which don't have valid JSON, or are broken in other ways.

Basically I'm saying that we should limit the scope of "must not perform any validation" otherwise it sounds like "the server must not reject any request at all".

tools/schema-generator.py
```diff
@@ -843,6 +843,7 @@ def main():
     model_definition = { "groups": {} }
     for input_file in args.input_files:
         with open(input_file) as file:
+            # print(f"File: {input_file}")
```
jskeet, line 846: My guess is this was for debugging purposes - I'd suggest either removing, uncommenting, or making it conditional on some debug flag.

schema/spec.md
duglin: done

schema/spec.md
duglin: Yes, and in particular it means all Versions of the Resource need to be checked too since it seems like a lie to me to have "validation"=true AND have some Versions of the Resource be non-compliant per the "format" attribute. Would it help if I elaborated more  in the spec?

schema/spec.md
duglin: As you said... how do you return warnings?

I guess one way to do it would be to have some kind of "status" attribute that's server controlled that would tell you whether or not the doc is valid. Then not just the current user would know but anyone checking the registry later too. Let's discuss because it does make things a bit more complex for users.

schema/spec.md
duglin: The intent was to let any array of bytes in. Imagine cases where people are purposely uploading invalid docs, perhaps for sharing testing (invalid) docs or they're using a format that's only valid because of the tooling they're using (like some kind of new templating language for the format).  I see two ways for users to say "let anything in"... 1) turn off validation,  2) use a format that's the server doesn't know about and therefore can't check (which includes "no format").

tools/schema-generator.py
duglin: this isn't the first time I spent too long trying to figure out where to add this debug line so I decided to make my life easier. But... as payback for all of your time doing these reviews I'll remove it :-)

schema/spec.md
jskeet: Right, thanks for the context. Presumably it's still only for the resource body though? I can't remember enough about other aspects of xRegistry requests to give concrete aspects, but I'd expect a sort of firewall between "the body of the resource, which is somewhat-opaque like CloudEvent data" and "resource metadata, authentication etc" which I'd still hope to be validated.

I'm hoping that you're able to clarify this in a more technically correct way, of course :)

cloudevents/schemas/document-schema.avsc
duglin: I just went thru and tried to use a phrase like "envelope format" to qualify it a bit more to avoid confusion when 'format' is actually a valid attribute near by. See if that help.

schema/spec.md
jskeet: Definitely worth elaborating more. I think it's also worth *considering* the use case of "I had broken resources before, but anything from now on should be valid" (in terms of any *new* versions). Or is the intention that if you had broken ones and you want validation, you should delete all the broken ones first?

message/schemas/document-schema.avsc
```diff
@@ -233,99 +233,132 @@
...
                           ]
                         }
                       ],
-                      "doc": "Binding identifier. This attribute MUST be the same as the 'binding' attribute of the definition group, but is not automatically inherited because definitions may be cross-referenced across groups"
+                      "doc": "Protocol identifier. This attribute MUST be the same as the 'protocol' attribute of the definition group, but is not automatically inherited because definitions may be cross-referenced across groups"
                     },
                     {
                       "type": "string",
                       "name": "schemaformat",
                       "doc": "The schema format applicable to the message payload, equivalent to the 'format' attribute attribute of the schema registry"
```
jskeet, line 318: Was this deliberately not updated to "envelope format"? Also, I suspect "attribute attribute" is unintentional :)

message/schemas/document-schema.json
```diff
@@ -58,18 +58,13 @@
           "description": "The schema format applicable to the message payload, equivalent to the 'format' attribute attribute of the schema registry"
```
jskeet, line 58: Same as previous comment.

message/schemas/openapi.json
```diff
@@ -1862,32 +1876,27 @@
             "description": "The schema format applicable to the message payload, equivalent to the 'format' attribute attribute of the schema registry"
```
jskeet, line 1876: And another :)

schema/model.json
```diff
@@ -23,9 +23,13 @@
...
               "type": "string",
-              "description": "Schema format identifier for this schema version",
-              "clientrequired": true,
-              "serverrequired": true
+              "description": "Schema format identifier for this schema version"
+            },
+            "validation": {
+              "name": "validation",
+              "type": "boolean",
+              "description": "Verify compliance with specified 'format'",
```
jskeet, line 31: Should this be:

- 'format'
- envelope format (no quotes)
- 'envelope'

? There are other occurrences of the same thing - I haven't commented on them.

schema/spec.md
duglin: yes, I was assuming that people would be forced to update all existing Versions before they could turn on validation. Annoying, but I think that's better than allowing for things to be in an inconsistent state.  That's why I have a bit of a reservation about allowing "old ones" to be non-compliant... how do we explain to new users of that Resource why some are invalid or even what the date is for the switch? "Required compliance" seems like the easier path. But let's discuss on the call.

Added a bit more text to make this clear.

schema/spec.md
duglin: `Indicates whether or not the server will validate the Resource's document(s) against the specified `format` attribute`. It's the phrase "Resource's document" that's they key to me as that means it's that secondary doc associated with each Version that's checked and not the xReg metadata.  Think I need to elaborate more?

jskeet: That's fair enough - I suspect the nature of line-by-line PR review made me less sensitive to the context :)

schema/model.json
duglin: I changed it to "schema `format`".

duglin:
Approved on the 9/19 call

PR #173, opened 2024-09-24
Changes for the week of Sept 23

### LINKED ISSUE #147: In the CloudEvents cases make message.ID=ce-type
In Endpoints and Message specs - maybe?
Can we make this a MUST?

Proposal:
Just in message spec: RECOMMEND that msg.ID=ce-type when CE is being used, for easy ID maintenance

### LINKED ISSUE #171: Schema isn't using RESOURCE* attribute
Today schema has:
```
          "schemaobject": ANY, ?         # TODO why is this not using RESOURCE*?
          "schema": "STRING", ?
          "schemauri": "URI" ?
```
instead of:
```
          "RESOURCEurl": "URL", ?
          "RESOURCE": ... Resource document ..., ?
          "RESOURCEbase64": "STRING", ?
```
or:
```
          "schemaurl": "URL", ?
          "schema": ... Resource document ..., ?
          "schemabase64": "STRING", ?
```
Typo?

duglin:
- make schema use the RESOURCE* attributes
- recommend that msg.id = ce-type attribute
- clarify that filter=name=* only matches when "name" is set (even "")

Fixes #171
Fixes #147

### DIFF
--- a/endpoint/spec.md
+++ b/endpoint/spec.md
@@ -600,12 +600,36 @@ TODO merge this into the previous 'protocol' section
   - If `protocoloptions.protocol` is a well-known protocol, the options MUST be
     compliant with the [protocol's options](#protocol-options).
 
+#### `messagegroups`
+
+The `messagegroups` attribute is an array of URI-references to message
+definition groups. The `messagegroups` attribute is used to reference
+message definition groups that are not inlined in the endpoint definition.
+
+Example:
+
+```yaml
+{
+  "protocol": "HTTP/1.1",
+  "options": {
+    "method": "POST"
+  },
+  "messagegroups": [
+    "https://example.com/registry/messagegroups/mygroup"
+  ]
+}
+```
+
 #### `messages`
 
 Endpoints are supersets of
 [message definition groups](../message/spec.md#message-definition-groups) and
 MAY contain inlined messages. See
 [Message Definitions](../message/spec.md#message-definitions).
+
+It is STRONGLY RECOMMENDED that the `messageid` values of the inlined messages
+be unique across all messages within the `messages` collection and the messages
+referenced by the `messagegroups` references.
 
 Example:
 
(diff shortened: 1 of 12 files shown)


endpoint/spec.md
```diff
@@ -600,13 +600,41 @@ TODO merge this into the previous 'protocol' section
...
+definition groups. The `messagegroups` attribute is used to reference
+message definition groups that are not inlined in the endpoint definition.
+
+It is STRONGLY RECOMMENDED that the `messageid` values of the referenced
+message definitions be unique across all message groups and all messages
+defined within the `messages` collection of this message group.
+
+Example:
+
+```yaml
```
jskeet, line 615: Should this be json instead of yaml?

jskeet: (This seems to happen elsewhere too...)

endpoint/spec.md
```diff
@@ -600,13 +600,41 @@ TODO merge this into the previous 'protocol' section
...
 
+#### `messagegroups`
+
+The `messagegroups` attribute is an array of URI-references to message
+definition groups. The `messagegroups` attribute is used to reference
+message definition groups that are not inlined in the endpoint definition.
+
+It is STRONGLY RECOMMENDED that the `messageid` values of the referenced
+message definitions be unique across all message groups and all messages
+defined within the `messages` collection of this message group.
```
duglin, line 611: Possibly: add: it's ok to reject update request if server chooses to validate that msgIDs are unique

on 9/26 call - backout this recommendation

duglin:
Approved on the 9/26 call w/o the one recommendation. See meeting notes for more inof

endpoint/spec.md
duglin: I checked and we consistently use "```yaml" even when it's json - so I'll leave this as is.

PR #174, opened 2024-09-24
Add meaning of xRegistry to README.md

manuelottlik:
We already explain the name in the Primer, but this change makes it even more obvious to new readers.


duglin:
LGTM 
Can you sign your commit?

duglin:
Approved on the 9/26 call

PR #176, opened 2024-10-02
Changes for week of Sept 30

duglin:
- fix code block
- multi-DELETE must ignore all attributes except ID and epoch

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -1187,6 +1187,7 @@ Content-Type: application/json; charset=utf-8
   "COLLECTIONscount": UINTEGER",                   # e.g. "endpointscount"
   "COLLECTIONs": { map of COLLECTION entities } ?  # If inlined
 }
+```
 
 ##### Deleting Entities in a Registry Collection
 
@@ -1218,7 +1219,7 @@ DELETE PATH-TO-COLLECTION
 {
   "KEY": {                                            # SINGULARid of entity
     "epoch": UINTEGER ?
-    # Remainder of entity serialization is OPTIONAL
+    # Remainder of entity serialization is ignored
   } *
 } ?
 ```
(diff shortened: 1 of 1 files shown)


core/spec.md
```diff
@@ -1233,11 +1234,9 @@ Where:
   to ensure that the value matches the entity's current `epoch` value and if it
   differs then an error MUST be generated.
 - If the entity's unique identifier is present in the object, then it MUST
-  match its corresponding key value. And for Versions, if `RESOURCEid` is
-  present then it MUST match the Resource's `SINGULARid` (`rID`) specified
-  in the URL.
+  match its corresponding `KEY` value.
```
jskeet, line 1237: I don't quite follow this change - hoping you'll explain it in the meeting.

jskeet (approved):
Fine except me not understanding one of the changes.

core/spec.md
duglin: it's just saying that:
```
  "v1": {
    "versionid": "v2"
  }
```
is invalid, "versionid" should have been "v1". Is it the phrase "entity's unique identifier" that's confusing?

jskeet: It's what was *removed* that's confusing me - in terms of why we don't want that any more.

duglin: Ah, I didn't want implementations to have to jump through hoops to check other fields when technically they don't matter. The only one that's critical is "epoch" (if present). I kept "id" as something to check just because if it doesn't match the KEY then your json itself is wrong and you might be deleting the wrong thing, which is different than forcing the impl to check with the backend/DB for verification. If you think checking the rID is still needed (if present) I can revert that part.

jskeet: I don't know enough about the rID part to have a meaningful opinion - I'd suggest just talking it through in the meeting later, and hopefully it'll all be fine. (I don't have a concrete objection.)

duglin: So I don't forget... one other reason I excluded rID is that in order to even know what the rID attribute's name is (e.g. schemaid), the impl would need to go to the model/DB and ask for it. When everything was just "id" that kind of look-up wasn't necessary. I can't imagine that's a big deal, but it feel into the category of "how much work do we want the impl to do just to 'double check' the user on a 'delete'?"  Yes, let's discuss on the call - not much of an agenda anyway :-)

duglin:
Approved on the 10/3 call

Issue #145, opened 2024-07-26
Remove requirement that schema versions are backwards compatible


duglin:
Provide a "compatibility" type of flag on the group (or resource) to indicate what type of compatibility checking (if any) the system should do. See: https://docs.confluent.io/platform/current/schema-registry/fundamentals/schema-evolution.html#compatibility-types for some options

duglin:
Proposal:
- define a "compatibility" extension in the schema registry model for SchemaGroups, non-strict enum
- default it to "backward"
- implementations can change the enum to make it strict, add values, or limit it to just a subset

values - taken from: https://docs.confluent.io/platform/current/schema-registry/fundamentals/schema-evolution.html#compatibility-types
<img width="583" alt="image" src="https://github.com/user-attachments/assets/24da677c-3344-4834-b33d-8002da44576c">

duglin:
Some options:
1 - define in core spec with Confluent's definitions
2 - define in core spec with just Confluent's names - subsequent specs define the names
3 - define in core spec w/o any names - subsequent specs (or impls) define the enum values
4 - do not define in core spec - define in subsequent specs
5 - do not define in core spec - define a "well known" extension (just names or names+meanings?)

duglin:
Proposal: 
- define the "compatibility" attribute in core as an enum with Confluent's values
- it's optional - in core spec
- no default value - in core spec
- subsequent specs can define default values, make it required and define the meaning of the enum values, add/remove enum values
- if a default value is defined in the model, then server MUST fill in the value if client omitted the attribute

PR #177, opened 2024-10-04
Updates for week of Oct 7

duglin:
- minor wording tweak w.r.t. epoch value
- 'location' is only allowed to appear in the model for Resource attributes
- define a 'compatibility' attribute for Resources. Define just some abstract
  values that people MAY use. Domain specific specs will narrow/expand the list
  and define what each value means for their domain/docs.
- define what "compatibility" means for schemas and make "compatibility"
  default to "backward". Remove the requirement that all versions are backwards
  compatible.
- new code gen files
- show line numbers of offending lines in some of the spec checking tools

Fixes: https://github.com/xregistry/spec/issues/145

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -181,6 +181,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
           "xref": "URL", ?                         # Ptr to linked Resource
           "epoch": UINTEGER,
           "readonly": BOOLEAN, ?                   # Default is "false"
+          "compatibility": "STRING", ?             # Default is "none"
 
           # These are inherited from the "default" Version
           "name": "STRING", ?
@@ -552,8 +553,7 @@ of the existing entity. Then the existing entity can be deleted.
   owning Resource's value. Updating a Resource or any of its Versions will
   update the shared Resource `epoch` value for all of those entities. This
   means that concurrent updates to different Versions of the same Resource
-  might result in an epoch validation error for the second update request
-  mention above.
+  might result in an epoch validation error for the second update request.
 - Constraints:
   - MUST be an unsigned integer equal to or greater than zero.
   - MUST increase in value each time the entity is updated.
@@ -1899,6 +1899,8 @@ The following describes the attributes of Registry model:
     used at the Resource level or Version level, but not both at the same time.
   - Type: String
   - OPTIONAL
+  - MUST only be used when defining top-level Resource attributes. Usage at
+    any other level in the model MUST generate an error.
   - If present, MUST be: `resource`, `version` or `both` - case sensitive.
   - When not present, the default value is `version`.
 
@@ -3179,7 +3181,7 @@ and the following Resource level attributes:
   by clients. This specification makes no statement as to when Resource are to
   be read-only.
 - Constraints:
-  - When not present, the default value is `false`,
+  - When not present, the default value is `false`.
   - REQUIRED when `true`, otherwise OPTIONAL.
   - When present, it MUST be a case sensitive `true` or `false`.
   - A request to update this attribute, on a non-read-only Resource, MUST be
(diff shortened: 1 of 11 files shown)


core/spec.md
```diff
@@ -3189,6 +3191,50 @@ and the following Resource level attributes:
...
+- Description: indicates whether Versions of this Resource need to adhere to
+  a certain compatibility rule. For example, a "backward" compatibility
+  value would indicate that all Versions of a Resource are mandated to be
+  backwards compatible with the next oldest Version, as determined by their
+  `createdat` timestamp attributes.
+
+  The exact meaning of what each `compatibility` value means might vary based
+  on the data model of the Resource, therefore this specification only defines
+  a very high-level abstract meaning for each to ensure some degree of
+  consistency. However, domain specification specifications are expected to
```
jskeet, line 3205: Nit: "domain specification specifications"

core/spec.md
```diff
@@ -3189,6 +3191,50 @@ and the following Resource level attributes:
     be silently ignored. See [Registry APIs](#registry-apis) for more
     information.
 
+##### `compatibility` Attribute
+- Type: String (with a defined enumeration of possible values)
```
jskeet, line 3195: Perhaps we should say "resource-specified enumeration of possible values" (or similar - resource may be the wrong word here)? Given that a specification can extend as well as contract the list below, *any* value could be included.

schema/spec.md
```diff
@@ -242,26 +243,28 @@ The type of the resource is `schema`. Any single `schema` is a container for
 one or more `Versions`, which hold the concrete schema documents or schema
 document references.
 
-Any new schema Version that is added to a schema definition SHOULD be backwards
-compatible with all previous Versions of the schema, meaning that a consumer
-using the new schema would be able to understand data encoded using a prior
-Version of the schema. If a new Version introduces a breaking change, it SHOULD
-be registered as a new schema with a new name.
+All Versions of a schema MUST adhere to the semantic rules of the schema's
+`compatibility` attribute. This specification defines "compatibility" to mean
```
jskeet, line 247: I think this definition of compatibility should either be dropped, or expanded to be much more precise. (I suspect dropping it would be better, given that we've already said that precise meanings are going to be resource-specific.)

core/spec.md
duglin: went with "resource-specified enumeration of possible values"

core/spec.md
duglin: it's always amuses me what my fingers type out of habit. I still can't type "contain" w/o first typing "container" and having to hit backspace twice. :-)

schema/spec.md
duglin: Given this is a domain specific spec (schema docs), I think making it more precise might be the better option. Do you have any suggestion for how to tighten it up? Or do you think that each schema language might introduce variants? @clemensv @deissnerk any thoughts?

jskeet: This is where my domain knowledge is insufficient. I'll just say that "backward compatible" is a far fuzzier concept than is normally appreciated, and *absolutely* depends on the precise context.

It's probably worth at least taking a stab at some precise meanings, while acknowledging that we may not get somewhere that we're happy with.

duglin: Current text:
```
This specification defines "compatibility" to mean that a consumer using one Version of 
the schema would be able to understand data encoded using another Version of the schema.
```
 Does this sound any better?
```
This specification defines "compatibility" for schemas as follows: version A of a schema is
said to be compatible with version B of a schema if any document that adheres to
the rules specified by schema A also adheres to rules specified by schema B.
```

jskeet: Hmm... that's one approach, but it may not be sufficient to really mean that a consumer would be able to understand the data. There can be additional constraints on "if field X exists, a consumer *must* do Y" which make that compatibility hard to define generically.

jskeet: I think the first bullet should be the other way round, if we expect version B to be later than version A. But otherwise, I think that's *at least* better than the current PR :)

duglin:
Approved on the 10/17 call with the text changes proposed.

PR #180, opened 2024-10-18
Schema generator update for new identifier model

clemensv:

Fixes #

Changes the identifier model to RESOURCEid and GROUPid

**Release Note**

```release-note
This change implements the revision of the "id" model where all ids of groups and resources are
prefixed with the singular group or resource name, e.g. "schemaid" for the schema resource. The
ids are also no longer required in the document representation since they can be inferred from
the containing dictionary entry keys.
```

### DIFF
--- a/tools/schema-generator.py
+++ b/tools/schema-generator.py
@@ -445,7 +445,7 @@ def generate_json_schema(model_definition, for_openapi=False) -> dict:
             props.update(copy.deepcopy(json_common_attributes))
             
             if resource.get("hasdocument", True):
-                 resource_schema = {
+                resource_schema = {
                     "type": "object",
                     "properties": props,
                     "oneOf": [
--- a/tools/schema-generator.py
+++ b/tools/schema-generator.py
@@ -80,29 +80,27 @@ json_type_mapping = {
 }
 
 json_common_attributes = {
-    "id": {"type": "string"},
-    "name": {"type": "string"},
-    "epoch": {"type": "integer"},
-    "self": {"type": "string", "format": "uri"},
-    "description": {"type": "string"},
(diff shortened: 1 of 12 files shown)


duglin:
Does it change "id" to "registryid" too?

tools/schema-generator.py
```diff
@@ -442,11 +440,14 @@ def handle_attributes(resource_schema, attributes):
...
             resource = resolve_resource(group, resource)            
             resource_name = resource["singular"]
-
+            props = {}
+            props[resource_name+"id"] = {"type": "string", "description": f"ID of the {resource_name} object"}
+            props.update(copy.deepcopy(json_common_attributes))
+            
             if resource.get("hasdocument", True):
-                resource_schema = {
+                 resource_schema = {
```
jskeet, line 448: I think this was accidental. We now have this at 5 spaces of indent, then 3 spaces in the next line.

jskeet (approved):
I think this may be the most trivial possible change request... otherwise fine.

tools/schema-generator.py
duglin: I'm confused. I see 4 space indents being used on the lines around this one.

jskeet: They were, yes - but after the change, there's 5 here (and 3 below). `if re` is 5 letters, whereas the indent should just be lined up after `if r`.

duglin: oh sorry - I thought this was my PR (not Clemens') so I was looking at an untouched version of this file.

duglin: Fixed

duglin:
Approved on the 10/24 call

PR #183, opened 2024-10-22
Changes for week of Oct 21

duglin:
- rename $meta to $json
- Move Resource level attributes into a nested Object ("meta")
- add a "metaattributes" to model for Resources
- remove "location" model attribute since we now have "metaattributes"
- update domain specific specs, models and code gen files
- removed the ?resource and ?attribs flags as they're  not needed any more

### DIFF
--- a/cloudevents/schemas/document-schema.avsc
+++ b/cloudevents/schemas/document-schema.avsc
@@ -748,15 +748,6 @@
                     },
                     {
                       "type": "string",
-                      "name": "compatibility"
-                    },
-                    {
-                      "type": "boolean",
-                      "name": "validation",
-                      "doc": "Verify compliance with specified schema 'format'"
-                    },
-                    {
-                      "type": "string",
                       "name": "format",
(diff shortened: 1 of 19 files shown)


jskeet (commented):
No comments on this one from me, as it's all stuff that requires actual knowledge of xRegistry, I think.

duglin:
Approved on the 10/24 call

jskeet:
I've been thinking about the `$json` bit - I personally prefer `$structure` in terms of being clear about the content returned (rather than the format). I know it's a bit longer, but is that actually a problem? In a choice between brevity or clarity, I generally favor the latter :)

duglin:
> I've been thinking about the `$json` bit - I personally prefer `$structure` in terms of being clear about the content returned (rather than the format). I know it's a bit longer, but is that actually a problem? In a choice between brevity or clarity, I generally favor the latter :)

Thsanks @jskeet I'll bring this up with the group.

PR #185, opened 2024-10-30
Changes for week of Oct 28

### LINKED ISSUE #8: xReg: issues with xReg (was discovery)
These might still apply to xReg - need to check:

WIP

https://github.com/cloudevents/spec/blob/master/discovery.md#get-services
- [ ] `Any Service previously returned to a client that does not appear in this result can be assumed to be no longer available.` I think this needs to be limited to just unfiltered queries since a missing entry from a filtered set could mean that it was deleted or that the filtered field changed.
- [ ] `Discovery endpoints MUST support filtering with the following attributes and MUST reject any unsupported filters. ` we should probably say what the HTTP response code is.
- [x] is Epoch allowed to be a negative int?
- [x] "name" must be non-empty string
- [x] typo: dataschama

Moved from: https://github.com/cloudevents/spec/issues/863

### LINKED ISSUE #179: Should ?model really be ?inline=model for consistency?
Proposal: yes

This would then mean that ?model on PUT/PATCH would leverage ?nested instead and follow the same semantics.

### LINKED ISSUE #182: s/GROUPs/GROUPS/g and others too?

### LINKED ISSUE #184: Support PATCH on Collections (groups, resources, versions)
Main reason this popped up is because there's no good way to update the "meta" sub-objects for multiple resources w/o also including the default versions in the request. It would be nice to be able to do:
```
PATCH .../RESOURCEs

{
  "rID2": {
    "meta": {
      "labels": { "stage": "dev" }
    }
  },
  ...
}
```

duglin:
- s/$json/$structure/g
- Support PATCH on collections
- s/GROUPs/GROUPS/g
- s/RESOURCEs/RESOURCES/g
- move ?model to ?inline=model, require ?nested on writes
- make no statement about 2nd GET missing entities meaning "deleted or not"

Fixes #8
Fixes #184
Fixes #182
Fixes #179

### DIFF
--- a/cloudevents/schemas/document-schema.json
+++ b/cloudevents/schemas/document-schema.json
@@ -1209,7 +1209,7 @@
                     "qos": {
                       "type": "integer",
                       "minimum": 0,
-                      "description": "The MQTT QoS level\u00c3\u00b6. May be 0, 1, or 2"
+                      "description": "The MQTT QoS level. May be 0, 1, or 2"
                     },
                     "retain": {
                       "type": "boolean",
--- a/core/spec.md
+++ b/core/spec.md
@@ -103,7 +103,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "createdat": "TIMESTAMP",
   "modifiedat": "TIMESTAMP",
 
-  "model": {                            # Only if requested
+  "model": {                            # Only if inlined/nested
     "schemas": [ "STRING" * ], ?        # Available schema formats
     "attributes": {                     # Registry level attributes/extensions
       "STRING": {                       # Attribute name (case sensitive)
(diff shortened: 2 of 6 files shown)


cloudevents/spec.md
```diff
@@ -165,7 +165,7 @@ schema for its payload.
           "schemaformat": "Protobuf/3.0",
           "schema": "syntax = \"proto3\"; message Metrics { float metric = 1; } }"
 
-          "metaurl": "https://example.com/endpoints/com.example.telemetry/messages/com.example.telemetry/meta",
+          "metaurl": "https://example.com/endpoints/com.example.telemetry/messages/com.example.telemetry/metadata",
```
jskeet, line 168: I'm intrigued as to this change, given that we're not changing (here, anyway) "meta" to "metadata" other than in this URL.

It's as if the spec can't decide whether "metadata" is clearer than "meta" or not...

jskeet (approved):
One aspect I'd love to hear the background on, but generally fine. Thanks for `$json` => `$structure` :)

cloudevents/spec.md
duglin: No big mystery... I messed up :-)
For some reason while doing $json->$structure I also made the meta->metadata change even though we agreed not to. By the time I realized my mistake I had already made lots of other edits and rather than starting over I did a metadata->meta round of edits... and missed some. Good catch!

duglin:
Approved on the 10/31 call

PR #188, opened 2024-11-06
Changes for week of Nov 4

### LINKED ISSUE #14: xReg: Feels like channel should be an array
@n3wscott wrote:

Feels like channel should be an array.

Originally posted by @n3wscott in https://github.com/cloudevents/spec/pull/1041#discussion_r942655170

Moved from: https://github.com/cloudevents/spec/issues/1062

### LINKED ISSUE #19: xReg: How do people know what content-type the messages will be?
We have a `format` property that will tell people that the message is a CloudEvent, but nothing in the Endpoint metadata to indicate whether it'll be xml, json, avro, ... or what the possible formats might be so people could specify that choice on the Subscribe() operation (if it supported it).

We should consider adding something to indicate the content-type(s) of the messages.

Moved from: https://github.com/cloudevents/spec/issues/1162

### LINKED ISSUE #146: Add statement in Endpoint spec that message.ID needs to be unique
under "messages" it already has to be unique because it's in a map.
but the "messagegroups" attribute could cause dups. We may not be able to make it a MUST, so SHOULD might be good enough and just warn people.

Proposal:
STRONGLY RECOMMENDED that they do it

### LINKED ISSUE #172: URL vs URI
We're inconsistent about when we use `schemauri` (message) vs `schemaurl` (schema), and even `RESOURCEurl`.

Proposal: make all of these URIs instead of URLs since it seems like any of them could want to point to a well-known (but not GETtable) entity.

### LINKED ISSUE #186: ?inline=* will show the model, but should it be special?
Now that ?model is gone and we treat it like all other inlinable sub-objects, `?inline=*` will inline/show all inlinable entities (model, groups, resources, resource docs, meta). However, "model" is different because it's not really "user data" as much as it's "the metadata describing the user data". As such, a normal user might not really want to see pages and pages of JSON for the model when they do "GET /?inline=*". It'll just feel like noise. Granted, this is only really an issue when they print the response - when in code they can just ignore that attribute.

However, should we do something special with "model"? Some options:
1 - no, treat it like all others and if they want to exclude it then they need to `?inline` just the ones they want - e.g. `?inline=endpoints.messages.versions,endpoints.messages.meta` for the endpoint registry
2 - allow `?inline=*` to pull in everything except "model". To include model they would need to explicitly mention it: `?inline=model,*`
3 - allow for exclusions - e.g.: `?inline=*&inline=-model`
4 - allow for the `*` to be more like a wildcard and allow it to mean "everything from this level down". E.g. `?inline=endpoints.*` means everything under `endpoints`, so `messages, messages.version, messages.meta`. An easier option 1. Note that `?inline=*` still has the exact same semantics.

Proposal: do 2 and 4
This will allow `?inline=*` to do what I think most people want (everything except "model") by default, but allow for an easy way to also include "model" with just a few more characters. I want to include 4 (even though it's technically a different issue/solution) because it just feels like a really nice UX feature to add while we're looking at this feature. But I'm ok with leaving option 4 out of the spec for now and adding it later if people are uncomfortable with it at this time.

duglin:
- message model fixes/missing attributes
- make ?inline=* exclude 'model' by default. Need to use 'model' explicitly
- fix some uri vs url inconsistencies
  - "endpoint.protocoloptions.authorization.resourceurl" ->
    "endpoint.protocoloptions.authorization.resourceuri"
  - "endpoint.protocoloptions.endpoint.uri" ->
    "endpoint.protocoloptions.endpoint.url"
- Add text about unique messageid values in Endpoints
- Add some clarifying text around `channel`
- Add 'datacontenttype' to message def
- Add 'data" prefix to message's: schemaformat, schema and schemauri
- require `datacontenttype` to match any other "content-type" type of field that might convey the same info - e.g. under CE envelope metadata
- regen schemas

Fixes #19
Fixed #186 
Fixes #172
Fixes #146
Fixed #14

### DIFF
--- a/endpoint/spec.md
+++ b/endpoint/spec.md
@@ -79,12 +79,12 @@ this form:
         # Common protocol options
         "endpoints": [
           {
-            "uri": "URI"                        # plus endpoint extensions
+            "url": "URL"                        # plus endpoint extensions
           } *
         ], ?
         "authorization": {
           "type": "STRING", ?
-          "resourceurl": "URL", ?
+          "resourceuri": "URI", ?
           "authorityuri": "URI", ?
           "grant_types": [ "STRING" * ] ?
         }, ?
@@ -283,6 +283,16 @@ The following attributes are defined for Endpoints:
   `producer` endpoint and a `consumer` endpoint, all with the same `channel`
   attribute value of "pipeline1", or some further qualification like
   "pipeline1-stage1", etc.
+
+  This specification does not make any statement about whether two endpoints
+  that do not share the same non-empty value have any relationship or not.
+  They might, but how this is determined is out of scope of this specification.
+  Additionally, while it is expected that this attribute's value will be a
+  single value, given this specification does not place any constraints on
+  its syntax or semantic meaning, implementations might choose to "encode"
+  multiple values within this single string. That would then imply that the
+  comparison algorithm of two `channel` values might need to be more
+  complicated than a "string compare" in those cases.
 
   When this property has no value it MUST either be serialized as an empty
   string or excluded from the serialization entirely.
@@ -497,22 +507,22 @@ TODO merge this into the previous 'protocol' section
   - Each object key MUST contain a `uri` attribute with a valid, absolute
     URI (URL)
 - Examples:
-  - `[ {"uri": "https://example.com" } ]`
+  - `[ {"url": "https://example.com" } ]`
   - ```
     [
-      { "uri": "tcp://example.com" },
-      { "uri": "wss://example.com" }
+      { "url": "tcp://example.com" },
+      { "url": "wss://example.com" }
     ]
     ```
   - ```
     [
       {
-        "uri": "tcp://example.com",
+        "url": "tcp://example.com",
         "priority": 1,
         "status": "down"
       },
       {
-        "uri": "wss://example.com",
+        "url": "wss://example.com",
         "priority": 2,
         "status": "up"
       }
@@ -627,10 +637,6 @@ Endpoints are supersets of
 MAY contain inlined messages. See
 [Message Definitions](../message/spec.md#message-definitions).
 
-It is STRONGLY RECOMMENDED that the `messageid` values of the inlined messages
-be unique across all messages within the `messages` collection and the messages
-referenced by the `messagegroups` references.
-
 Example:
 
 ```yaml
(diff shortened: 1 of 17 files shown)


clemensv:
Looks good

clemensv:
Replace lines 300-322 of the schema generator:

``` python
def handle_attributes(resource_schema, attributes):
        """
        This function takes in a resource schema and a dictionary of attributes and their properties.
        It iterates through each attribute and creates a JSON schema for it based on its properties.
        The function also handles nested attributes and conditional attributes using the "ifvalues" property.
        The resulting schema is added to the resource schema.
        """
        for attr_name, attr_props in attributes.items():
            if attr_props["type"] == "object":
                attr_schema = {"type": "object", "description": "", "properties": {}}
                if "attributes" in attr_props:
                    handle_attributes(attr_schema, attr_props["attributes"])
            else:
                attr_schema = copy.deepcopy(json_type_mapping[attr_props["type"]])
                        
            if "description" in attr_props:
                attr_schema["description"] = attr_props["description"]
            if "description" in attr_schema and attr_schema["description"] == "":
                del attr_schema["description"]                
                            
            if attr_props["type"] == "object" or attr_props["type"] == "map" or attr_props["type"] == "array":
                if "item" in attr_props:
    ```

**Comment by duglin** [comment]
@clemensv should I see any issue with the code gen? I don't

File: message/model.json [hunk 1]
```diff
@@ -592,10 +780,24 @@
                           "description": "MQTT qos",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 787: remove

message/model.json
```diff
@@ -604,10 +806,24 @@
                           "description": "MQTT retain",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 813: remove

message/model.json
```diff
@@ -616,10 +832,24 @@
                           "description": "MQTT topic-name",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 839: remove

message/model.json
```diff
@@ -639,10 +869,24 @@
                           "description": "MQTT qos",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 876: remove

message/model.json
```diff
@@ -651,10 +895,24 @@
                           "description": "MQTT retain",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 902: remove

message/model.json
```diff
@@ -675,10 +947,24 @@
                           "description": "MQTT message-expiry-interval",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 954: remove

message/model.json
```diff
@@ -687,10 +973,24 @@
                           "description": "MQTT response-topic",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 980: remove

message/model.json
```diff
@@ -699,10 +999,24 @@
                           "description": "MQTT correlation-data",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 1006: remove

message/model.json
```diff
@@ -711,10 +1025,24 @@
                           "description": "MQTT content-type",
                           "type": "object",
                           "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
```
duglin, line 1032: remove

message/model.json
```diff
@@ -812,18 +1158,30 @@
...
+                                "type": "boolean"
                               }
                             }
                           }
                         },
                         "query": {
                           "name": "query",
                           "description": "The HTTP query parameters",
-                          "type": "object"
+                          "type": "map",
```
duglin, line 1181: Make this an array of objects, like headers.

endpoint/spec.md
```diff
@@ -657,6 +663,33 @@ Example:
...
+be done via examination of some other metadata - such as the message's
+`envelopemetadata.type` value along with its `envelopeoptions.format` value.
+These details are of scope for this specification to define and are left as
+an implementation detail.
+
+Implementations MAY choose to generate an error if it detects duplicate
+`messageid` values across the `messages` collection message definitions and
+the `messagegroups` referenced message definitions, if that is the desired
+constraints for their users.
+
```
duglin, line 692: @jskeet this was the "may generate an error" text you asked for a few weeks ago.

duglin:
Approved on the 11/14 call

PR #195, opened 2024-11-26
Edits for week of Dec 2

duglin:
- Reorder "defaultversion*" attrs: id, url, sticky - based on importance
- Make it clear that ?nested isn't needed when `model` or `meta` are present
  on requests
- epoch/modifiedat are only updated when nested entities are added or
  removed from the collection, not when just edited
- added `?` before each query parameter (eg `filter`->`?filter`) for consistency

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -207,9 +207,9 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
             "compatibility": "STRING", ?           # Default is "none"
             # TODO - add timestamp?
 
-            "defaultversionsticky": BOOLEAN, ?     # Default is "false"
             "defaultversionid": "STRING",
-            "defaultversionurl": "URL"
+            "defaultversionurl": "URL",
+            "defaultversionsticky": BOOLEAN ?      # Default is "false"
           }, ?
           "versionsurl": "URL",
           "versionscount": UINTEGER,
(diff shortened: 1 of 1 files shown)


core/spec.md
```diff
@@ -5127,12 +5137,12 @@ to have any inlinable attributes (except for `RESOURCE*`) that are present in
...
 
-Note: If the Registry can not return all expected data in one response then it
-MUST generate an HTTP `406 Not Acceptable` error and SHOULD include a error
-message in the HTTP body indicating that the response is too large to be
-sent in one message. In those cases, the client will need to query the
-individual inlinable attributes in isolation so the Registry can leverage
-[pagination](../pagination/spec.md) of the response.
+Note: If the Registry can not return all expected data in one response because
+it is too large then it MUST generate an HTTP `406 Not Acceptable` error and
+SHOULD include a error message in the HTTP body indicating which attributes
```
deissnerk, line 5142:
```suggestion
SHOULD include an error message in the HTTP body indicating which attributes
```

deissnerk (commented):
LGTM
As you consistently introduced the `?queryParameter` notation, it might be useful to add a sentence to the _Notational Conventions_ section explaining that query parameters are always highlighted with a leading `?`. WDYT?

core/spec.md
duglin: Fixed the "a" and added some text about ?query

duglin:
Approved on the 12/5 call

PR #208, opened 2024-12-17
2024-12-17 meeting notes

Fannon:
This is not meant for merging, but contains the meeting notes from 2024-12-17 xregistry meeting with @duglin , @clemensv , @deissnerk  and @Fannon .


message/model.json
```diff
@@ -236,9 +239,14 @@
                               "name": "required",
                               "description": "The uri required",
                               "type": "boolean"
-                            }
+                            },
                           }
                         },
+                        "dataschemaid": {
+                          "type": "relation",
+                          // If we point to a specific version, we append /versions
```
Fannon, line 247:
```suggestion
                          // If we point to a specific version, we append /versions
                          // If we point to resource or specific version, append /versions?
```

core/spec.md
```diff
@@ -1276,7 +1276,24 @@ The serialization of the Registry entity adheres to this form:
   "registryid": "STRING",
```
Fannon, line 1276: If we have `xid`, do we still need the current ids?

endpoint/model.json
```diff
@@ -765,7 +785,20 @@
           "type": "array",
           "description": "The message groups that are supported by this endpoint",
           "item" : {
-            "type": "uri"
+            // New type: "relation", pointer to data inside of the xregistry
+            "type": "relation",
+            // Target "type", if omitted it can point to any type inside the registry
```
duglin, line 790: oops, the spec has this as required not optional. Let's discuss along with https://github.com/xregistry/spec/issues/226

core/spec.md
duglin: Agreed on 1/15 call to keep "SINGULARid"

endpoint/model.json
duglin: agreed on 1/15

duglin:
@Fannon I think we have all of these edit in the spec now.
I'll let you check and then close it

Fannon:
Closing as this was not meant to be merged, just meeting notes that have now been incorporated by other PRs.

Issue #84, opened 2024-04-04
An outsiders view on xRegistry


manuelottlik:
Disclaimer: I am probably the youngest and least experienced person in this group and also just joined recently. At the same time, this is my first „real“ open source contribution, so I would love some feedback.

I took the time over the easter holidays to read the primer and spec from the beginning to the end. These are my observations & questions from an outsider’s & end user perspective:

I think the following clarifications would make it easier for people to understand the purpose of xRegistry:
* Why does this project exist?
  * What problem is it trying to solve which already existing projects / products do not solve yet?
* What is the targeted use case of this product?
  * If there would be one main thing that should be achieved using this product, what would it be?
  * Are there example use cases?
* What people / personas are supposed to work with the product?
  * Vendors like Microsoft? End users like HDI?
  * Platform engineers? Any developer whose team owns a schema?
* What is in scope of that project, what is out of scope?
  * Is authorization to entities in scope? (Who can edit which schemas?)
* Is there a page dedicated to explaining the xregistry schema language / format?

I think these things would make it easier for people to adopt or contribute to xRegistry:
* How are we documenting architectural decisions like the ones we constantly make in our calls?
* Should we update the roadmap? It makes the project look dead when it clearly is not.
* Should we include the zoom link to the tuesday & friday calls in the README? It would make it easier for people to contribute.

I think using established description languages could be a great addition to the spec:
* Can we use a schema language (e.g. JSON schema) to describe the model of the registry?
* Can we use OpenAPI to describe the REST interface of the registry?

Again, even though this is far from the type of stuff we are discussing on our weekly calls, I hope it benefits the graduation of the specification – feedback is appreciated!

duglin:
@manuelottlik do we still need this issue?

duglin:
@manuelottlik do you want to create a PR(s) to address the open issues?

manuelottlik:
I added 1 and 3 to the meeting minutes for Friday and created a separate issue for the roadmap.

duglin:
With the other issues/PR, can we close this now?

PR #219, opened 2024-12-23
Add value proposition section to the primer

lailabougria:
Connected to #84

Based on our conversation in the meeting of Dec 5th and a chat I had with @clemensv, I wrote a rough first draft for a value proposition of the specification. I approached this from my own perspective: that of a user who wants to understand what xRegistry offers to improve how I build and maintain event-driven systems, based on the limited knowledge I have of the spec at the moment. 

We can use this as a way to validate initial assumptions and, hopefully, this can be improved upon, extended, and finally included in the primer. Please point out any false assumptions I made, or any other angles I’ve missed and I’m happy to make changes accordingly.

## Proposed Changes

- Adds a value proposition to the primer

**Release Note**

```release-note

```

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -14,6 +14,7 @@ normative technical details.
 ## Table of Contents
 
 - [History](#history)
+- [Value proposition](#value-proposition)
 - [Motivation](#motivation)
 - [Design Goals](#design-goals)
 - [Representations](#representations)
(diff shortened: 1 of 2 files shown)


core/primer.md
```diff
@@ -35,6 +35,112 @@ xRegistry (extensible registry) specification was created.
 xRegistry was initially part of CloudEvents, called "CloudEvents Discovery" but
 later moved into it's own repository.
 
+## Value proposition
+
+xRegistry provides a specification to define metadata, schemas and extensions
```
duglin, line 40: @lailabougria thanks! Everything you wrote here is correct and well said. However, it's presented within the context of event-driven systems. While that's obviously a valid usecase, and a major driver of our work, we need to be careful not to give the impression that xRegistry is just about eventing related topics. The core spec itself has nothing to do with events, protocols, endpoints, schemas.... it's only when we get to the secondary specs (endpoints, messages, schemas) that an eventing-specific view is introduced. We want to make sure that people know that xRegistry itself can be use for just about any kind of metadata as long as it can fit into our overall model of Groups and (optional) Resources/versions.

What do you think about having a more generic paragraph as an intro to this and then use the rest of the text as more of a "let's look how this generic xReg spec can be used in a more specific/real-world technology space" ?

lailabougria: Yes, absolutely, that's something that I can work on and extend. I started from the EDA context as that was the easiest path for me to validate any assumptions I made, and I'm happy to work on this further. I'll attend the meeting tomorrow, and we can discuss it shortly if there's time.

lailabougria: @duglin Take a look at the latest change to see if that alleviates the concern?

core/primer.md
```diff
@@ -33,7 +34,130 @@ associating those declarations with application endpoints. As a result, the
...
+The registry contains granular definitions for events defining their schemas,
+which serve as a contract between publishers and consumers. This doesn’t only
+apply to the message payload, but also to the message metadata, allowing
+granular control of required fields on the metadata level. The registry does not
+only serve as a definition of the required schemas, but can also be used to
+enforce it. A registry service could use the registry to perform validation
+on-publish, rejecting any events that don’t adhere to the schema for that
+context, avoiding the problem of poison messages.
+
+### Payload reduction
```
duglin, line 115: This is just saying that payloads can have a URL to a schema doc instead of embedding it, right?

core/primer.md
```diff
@@ -33,7 +34,130 @@ associating those declarations with application endpoints. As a result, the
...
+contextual meaning, which is why applications often use serialization mechanisms
+which include schema information (such as JSON) in payloads. Having a
+centralized registry, enables applications to rely on lighter weight
+serialization formats, significantly reducing bandwidth, while still allowing
+accurate real-time analysis of incoming data using the registry.
+
+### Schema-based data contract generation
+
+xRegistry provides a specification to manage metadata in files (for example,
+stored next to relevant source code), but the metadata can also be managed in a
```
duglin, line 127: While the rest of this section is true, I think storing the files next to the code is too quickly passed-over. Being able to link the proper version of the model and corresponding code via co-location in git huge. Perhaps we can expand a bit on this (even with just one more sentence) by talking about how much easier it makes it to recreate the state of the system/environment when everything can be stored together w/o complicated tooling or "linking" of data sources.

core/primer.md
```diff
@@ -33,7 +34,130 @@ associating those declarations with application endpoints. As a result, the
...
+registry. This is not only limited to event consumers, but may also include
+generation of event store inserts. This can further speed up and streamline the
+development of both producers and consumers, significantly reducing development
+and testing time and minimizing the risk of introducing bugs at this level.
+Clemens Vasters built a proof-of-concept that shows this approach in
+the [Avrotize VS Code Extension](https://github.com/clemensv/avrotize).
+
+### Basis for further developments
+
+CloudEvents serve as a well-defined event-format, which is applicable
```
duglin, line 149: I get nervous when people call CE an "event-format" because it's not really meant to be that, even though it does offer a JSON event format that people MAY use if they don't already have a format defined. But, and this is critical, CE is really more about its metadata and where the metadata can be found independent of the event format. CE is NOT meant to be "one event format to rule them all". That has been tried and failed, many times.

So, perhaps instead say something like: CloudEvents serve as a mechanism to aid in the delivery of events from a producer to a consumer, which is applicable independently ....

duglin:
Overall it looks good - just a few minor comments.
Also, let's add:
```
apicur
avrotize
aws
clemens
clemensv
microsoft
vasters
```
to `tools/dict` so the spell checker is happy.

core/primer.md
lailabougria: Yes that's the intent, was it unclear?

duglin: no, it's fine - I just wanted to make sure I wasn't missing some other point. I'm slow sometimes :-)

core/primer.md
lailabougria: Addressed in [`0b2ba34` (#219)](https://github.com/xregistry/spec/pull/219/commits/0b2ba34b42d3a45b2812af948bf045c6bb7bb9d9)

core/primer.md
lailabougria: Addressed in [`08032ac` (#219)](https://github.com/xregistry/spec/pull/219/commits/08032accc895eb10afd75e1bd3e85793f93d18b5)

duglin:
@lailabougria it might best if you squash this down to just one commit and sign that one, otherwise you'll need to sign all of the commits.

duglin:
Looks like some bad links in the primer:
```
core/primer.md: /endpoint does not exist
core/primer.md: /schema does not exist
core/primer.md: /message does not exist
```

duglin:
try relative links instead of absoute.... e.g.  `../message/spec.md` instead of `/message/spec.md`.
Also you should be able to run the tests locally if you want:  `make -f tools/Makefile docker`
(if you have docker installed). Or just `make -f tool/Makefile` if you have python installed locally

lailabougria:
I didn't know that, maybe it's worth adding to the contributing guidelines? All tests passed this time locally...

duglin:
good idea- I added it to our CONTRIBUTING doc

core/primer.md
```diff
@@ -33,7 +34,140 @@ associating those declarations with application endpoints. As a result, the
...
+aims to address this with an agnostic specification, providing a common base
+across these offerings.
+
+### Versioning
+
+As systems evolve, the endpoints that emit events may change, as may the events
+themselves. Changes can sometimes break consumers making compatibility
+strategies essential to indicate to consumers how they may expect events or
+endpoints to evolve. xRegistry allows indicating a registry’s compatibility
+strategy which is enforced when making changes to the registry, avoiding
```
duglin, line 99: This line might need to change based on our recent discussions.

Issue #214, opened 2024-12-17
Add text to Primer for "group suffix" strings


duglin:
- Suggest using a domain or "authority" suffix for group type names to avoid name collisions when exporting groups

Issue #213, opened 2024-12-17
Add `xid` attribute to all entities


duglin:
- format: `/[GROUPS/gid[/RESOURCES/rid[/versions/vid]]]`
- it's value is basically the tail part of the `self` URL (everything after the base URL of the registry) w/o `$structure` or query parameters.
- Note it MUST start with `/` and the `xid` of the registry is `/`
- e.g `schemagroups/blobstore/schemas/blobCreated` for the blobCreated schema resource
- `xid` values can be used in `relation` type of attributes and MUST match the `target` model type

See #209

Issue #212, opened 2024-12-17
Add `selfshort` feature/URL


duglin:
- As a sibling to `self` this attribute contains a "short" URL version of the "self" URL.
- When the `shortself` capability is enabled the `selfshort` attribute will appear on all entities
- note that resources with "hasdoc=true" MUST have 2 `selfshort` values, one that include `$structure` and one that doesn't. Which one is shown in `shortself` will match which normal URL is used in the `self` attribute
- When the capability is enabled, there MUST be a `resolver` API/feature enabled that will accept GET/HEAD requests to the `shortself` URL and returns a 308 with a `Location` value that is the same as the `self` URL (plus `$structure` if necessary)
- when the entity behind the shorturl is deleted then the shorturl MUST be deleted from the resolver and a 404 would be returned by an HTTP request to that short URL. If a new entity is created at the same `self` URL then the same selfshort URLs MUST be recreated in the resolver
- the resolver service URL is `RegistryBasePath`+`/r` and the `selfshort` URL MUST be `.../r?u=STRING` where "STRING" must be as short as possible
- Note that URLs to collections will not have shortURLs - is this a problem? There's no place to show them unless we create something like `versionsurlshort` as a sibling to `versionsurl` (and all other collection URLs), which is possible but very noisy. Same for `/model` and `/capabilities` URLs.
- Future: support the ability to define shortURLs for non-registry URLs

duglin:
If we require the same short-url to be used even after an entity was deleted and recreated, then we're stuck with some kind of hash/sha algorithm, and those typically return fixed length values. And those can end up being longer than the original URL. Some examples:
```
     "self": "http://xregistry.example.com/",
"shortself": "http://xregistry.example.com/r?u=1b2m2y8aSGtPGaMy7pHcFG",

     "self": "http://xregistry.example.com/endpoints/e1",
"shortself": "http://xregistry.example.com/r?u=tokbUg9t7xN0ox8hNrY5YW",

     "self": "http://xregistry.example.com/endpoints/e1/messages/blobCreated",
"shortself": "http://xregistry.example.com/r?u=iyXYSy3BpIdCPLCRZpZHxa",

     "self": "http://xregistry.example.com/schemagroups/g1/schemas/popped$structure",
"shortself": "http://xregistry.example.com/r?u=jyXFhvRYIzIRj178TxEPRq",
```

Some options:
1 - deal with it, always use the hash string
2 - deal with it, but allow the shortself to be "self" if it's longer than "self"
[...]

Issue #211, opened 2024-12-17
Add a "compatiblewith" well-known model label


duglin:
See #210 

Add a "well-known labels" section to the spec and include this one in there.

ping @Fannon for descriptive text for the label.

Fannon:
@duglin  - do you have added it already, as its marked Done? 
If yes, should I create a PR to propose more description - e.g. in a section that describes well-known labels?

duglin:
@Fannon yes it's part of https://github.com/xregistry/spec/pull/222

I added this stuff: https://github.com/xregistry/spec/pull/222/files#diff-1748a963736c86b1ae7e1139dbbffbb26f059c25811b9c4899e2e2d4a0e605a2R1991

You can either do another PR to expand on it (and run the risk of having to rebase after my PR is merged), or can just add a comment to that PR with any changes you want to make.  Up to you.

Fannon:
Thanks Doug! I'll try to find some time these days and would then start with a comment suggestion first. 
If more things pile up, I'll consider a dedicated PR.

PR #222, opened 2024-12-23
Week of Jan 6 changes

### LINKED ISSUE #199: DELETE on .../meta? And other unspecified APIs
Ban or leave unspecified?

### LINKED ISSUE #216: remove "origin"

### LINKED ISSUE #217: add a "capabilities" section
- sibling to "model", available via `?inline=capabilities`
- not shown by default
- also available via `/capabities`
- add all optional features to this
- move the "schemas" section of the model to be a capability

duglin:
- clarify errors for bad paths and methods
- Remove "origin"
- add `/capabilities` API and "capabilities" Registry attr
  - requires `?inline=capabilities` reads, just presence on writes(like model)
  - remove "schemas" from model, moved into "capabilities"
- add 'xid'. Change xref to start with /
- add "labels" to model schema : Registry, Groups and Resources
  - add a "compatiblewith" well-known label
- add support for `relation` type of URLs
- add text about how to deal with sending incomplete results
  - error unless you can be tricky via something like pagination spec
- allow ?export, ?inline, ?filter on all APIs to control responses
- explain that "generated" errors != must send back to client - e.g. security
- use highest alphabetical versionid (case insensitive) when all versions have
  the same "createdat" value. Remove requirement for ?setdefaultversionid
- `DELETE RESOURCES` must put `epoch` under `meta` if present
- Primer: explain why GROUPS and RESOURCES have attribute name restrictions
- Primer: encourage domain/company prefix/suffixed on Groups (and Resources)
- Add `capability.enforcecompatibility` so servers can indicate whether they'll
  enforce the `compatibility` attribute on the owning Resource
- Clarify rules for updating Resources when dealing with 'xref'.
  - presence of any attribute on a 'write' to an xref'd resource is an error
  - xref->normal:
    - epoch = max(oldEpoch, targetResourceEpoch)+1
	- createdat = oldCreatedAt
	- modifiedat = now
- Updating a Version's `createdat` CAN make it the latest/default
- add support for "shortself"
- add "@" as a valid char for IDs

Fixes https://github.com/xregistry/spec/issues/199
Fixes https://github.com/xregistry/spec/issues/216
Fixes https://github.com/xregistry/spec/issues/217
Fixes https://github.com/xregistry/spec/issues/213
Fixes https://github.com/xregistry/spec/issues/210
Fixes https://github.com/xregistry/spec/issues/211
Fixes https://github.com/xregistry/spec/issues/209
Fixes https://github.com/xregistry/spec/issues/187
Fixes https://github.com/xregistry/spec/issues/200
Fixes https://github.com/xregistry/spec/issues/202
Fixes https://github.com/xregistry/spec/issues/205
Fixes https://github.com/xregistry/spec/issues/207
Fixes https://github.com/xregistry/spec/issues/214
Fixes https://github.com/xregistry/spec/issues/203
Fixes https://github.com/xregistry/spec/issues/204
Fixes https://github.com/xregistry/spec/issues/206
Fixes https://github.com/xregistry/spec/issues/212

### DIFF
--- a/core/model.json
+++ b/core/model.json
@@ -1,7 +1,4 @@
 {
-  "schemas": [
-    "xregistry/0.5"
-  ],
   "attributes": {
     "specversion": {
       "name": "specversion",
@@ -19,6 +16,12 @@
     "self": {
       "name": "self",
       "type": "url",
+      "readonly": true,
+      "serverrequired": true
+    },
+    "xid": {
+      "name": "xid",
+      "type": "uri",
       "readonly": true,
       "serverrequired": true
     },
--- a/core/primer.md
+++ b/core/primer.md
@@ -477,3 +477,36 @@ Group and Resource names limited to only 58? Because when they appear as
 Collection attributes and we append `url` or `count` to them they still have
 to fit within the 63 character limit, and so we need to take into account
 the length of the word `count`.
+
+# Why must Group type and Resource type names be valid attribute names?
+
+Since Group type and Resource type names will appear as attribute names (e.g.
+`GROUPScount` and `RESOURCEurl`), the character set (and length) of their
+names need to be restricted in the same way as a attribute names.
+
+# Choosing unique Group and Resource names
+
+Aside from choosing names that give some descriptive meaning to the purpose
+of these entities, care should be taken when choosing names for entities that
+might be imported into other Registries. For example, Group names are often
+suffixed (or prefixed) with domain (or company) specific words to avoid
+name collisions with other similarly named entities. This might be useful
+for Resource names too (if they might co-exist under shared Groups), but
+normally ensuring uniqueness at the Group level is sufficient.
+
+# Are `self` and `shortself` attributes static?
+
+In general any URL reference to an entity in the Registry should remain
+static for the lifetime of the entity, otherwise anything persisting that
+reference will break when it attempts to use it. So, in this sense, yes
+these attributes should be static in general. However, the specification
+can not mandate this because there may be times when the risk of breaking
+existing references becomes necessary.
+
+For example, perhaps a Registry needs to move to a new host, and the old
+one is no longer available, even to offer a redirect. Or, if the server
+needs to switch to a new "URL shortener" service and so all of the "shortself"
+values need to change.
+
+Obviously, situations like these should be rare but the specification needs
+to allow for them.
--- a/core/spec.md
+++ b/core/spec.md
@@ -20,6 +20,7 @@ automation and tooling.
   - [Registry Entity](#registry-entity)
     - [Retrieving the Registry](#retrieving-the-registry)
     - [Updating the Registry Entity](#updating-the-registry-entity)
+  - [Registry Capabilities](#registry-capabilities)
   - [Registry Model](#registry-model)
     - [Retrieving the Registry Model](#retrieving-the-registry-model)
     - [Updating the Registry Model](#updating-the-registry-model)
@@ -39,8 +40,10 @@ automation and tooling.
     - [Creating or Updating Versions](#creating-or-updating-versions)
     - [Retrieving a Version](#retrieving-a-version)
     - [Deleting Versions](#deleting-versions)
-  - [Inlining](#inlining)
-  - [Filtering](#filtering)
+  - [Configuring Responses[#configuring-responses]
+    - [Inlining](#inlining)
+    - [Filtering](#filtering)
+    - [Exporting](#exporting)
   - [HTTP Header Values](#http-header-values)
 
 ## Overview
@@ -95,6 +98,8 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "specversion": "STRING",
   "registryid": "STRING",
   "self": "URL",
+  "shortself": "URL", ?
+  "xid": "URI",
   "epoch": UINTEGER,
   "name": "STRING", ?
   "description": "STRING", ?
@@ -103,12 +108,32 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "createdat": "TIMESTAMP",
   "modifiedat": "TIMESTAMP",
 
+  "capabilities": {                     # Supported capabilities/options
+    "enforcecompatibility": BOOLEAN, ?
+    "flags": [                          # Query parameters
+      "epoch",? "export",? "filter",? "inline",? "nested",?
+      "nodefaultversionid",? "nodefaultversionsticky",? "noepoch",?
+      "noreadonly",?  "schema",? "setdefaultversionid",? "specversion",?
+      "STRING" *
+    ],
+    "mutable": [                        # What is mutable in the Registry
+      "capabilities",? "entities",? "model",? "STRING"*
+    ], ?
+    "pagination": BOOLEAN, ?
+    "schemas": [ "xRegistry-json/0.5", "STRING" * ], ?
+    "shortself": BOOLEAN, ?
+    "specversions": [ "0.5", "STRING"* ], ?
+
+    "STRING": ... *                     # Extension capabilities
+  }, ?
+
   "model": {                            # Only if inlined/nested
-    "schemas": [ "STRING" * ], ?        # Available schema formats
+    "labels": { "STRING": "STRING" * }, ?
     "attributes": {                     # Registry level attributes/extensions
       "STRING": {                       # Attribute name (case sensitive)
         "name": "STRING",               # Same as attribute's key
         "type": "TYPE",                 # string, decimal, array, object, ...
+        "target": "STRING", ?           # If "type" is "relation"
         "description": "STRING", ?
         "enum": [ VALUE * ], ?          # Array of scalar values of type "TYPE"
         "strict": BOOLEAN, ?            # Just "enum" values or not.Default=true
@@ -137,6 +162,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
       "STRING": {                       # Key=plural name, e.g. "endpoints"
         "plural": "STRING",             # e.g. "endpoints"
         "singular": "STRING",           # e.g. "endpoint"
+        "labels": { "STRING": "STRING" * }, ?
         "attributes": { ... }, ?        # Group level attributes/extensions
 
         "resources": {
@@ -148,13 +174,14 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
             "setdefaultversionsticky": BOOLEAN, ?    # Supports client "default" selection
             "hasdocument": BOOLEAN, ?     # Has separate document. Default=true
             "typemap": MAP, ?             # contenttype mappings
+            "labels": { "STRING": "STRING" * }, ?
             "attributes": { ... }, ?      # Version attributes/extensions
             "metaattributes": { ... } ?   # Resource attributes/extensions
           } *
         } ?
       } *
     } ?
-  } ?
+  }, ?
 
   # Repeat for each Group type
   "GROUPSurl": "URL",                              # e.g. "endpointsurl"
@@ -163,12 +190,13 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
     "KEY": {                                       # Key=the Group id
       "GROUPid": "STRING",                         # The Group ID
       "self": "URL",
+      "shortself": "URL", ?
+      "xid": "URI",
       "epoch": UINTEGER,
       "name": "STRING", ?
       "description": "STRING", ?
       "documentation": "URL", ?
       "labels": { "STRING": "STRING" * }, ?
-      "origin": "URI", ?
       "createdat": "TIMESTAMP",
       "modifiedat": "TIMESTAMP",
 
@@ -181,13 +209,14 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
           "RESOURCEid": "STRING",
           "versionid": "STRING",
           "self": "URL",                           # Resource URL, not Version
+          "shortself": "URL", ?
+          "xid": "URI",
           "epoch": UINTEGER,                       # Version's epoch
           "name": "STRING", ?
           "isdefault": true,
           "description": "STRING", ?
           "documentation": "URL", ?
           "labels": { "STRING": "STRING" * }, ?
-          "origin": "URI", ?
           "createdat": "TIMESTAMP",
           "modifiedat": "TIMESTAMP",
           "contenttype": "STRING, ?
@@ -201,6 +230,8 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
           "meta": {                                # Only if inlined
             "RESOURCEid": "STRING",
             "self": "URL",                         # URL to "meta" object
+            "shortself": "URL", ?
+            "xid": "URI",
             "xref": "URL", ?                       # Ptr to linked Resource
             "epoch": UINTEGER,                     # Resource's epoch
             "createdat": "TIMESTAMP",              # Resource's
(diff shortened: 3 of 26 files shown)


core/spec.md
```diff
@@ -1754,16 +1988,18 @@ Regardless of how the model is retrieved, the overall format is as follows:
...
+    Registry to be stored without changing the schema of the model.
+  - If present, MUST be a map of zero or more name/value string pairs.
+  - Keys MUST be non-empty strings.
+  - Values MAY be empty strings.
+  - The following labels are defined for convenience and MAY also be used
+    as labels for Groups or Resource model types:
+    - Name: `compatiblewith`
+      Value: A URI reference to Registry that this Registry's model is
+      compatible with. This specification does not mandate any runtime
+      validation of this claim.
```
Fannon, line 2002: Here we need to clarify how to point to a particular model, it's not sufficient to point to an xregistry itself as it could contain all kinds of models. Compatibility will always be between particular models, not between the registries as a whole.

```suggestion
    - Name: `compatiblewith`
      Value: A URI reference to an xRegistry model that this model is
      compatible with. All models that share the same reference are interoperable. 
      This specification does not mandate any runtime validation of this claim.
```

But we need to clarify probably a few more points:
* How to construct this URL? Ideally, we would point to the model, e.g. on group side we'd point from the group model to e.g. https://xregistry.cncf.org/model#groups/schemagroups (do we have a link where I can retrieve just a partial model? Otherwise we need anchor tags here. 
* What does compatible mean? 
  * It's primarily about ecosystem compatibility around a model (like we have with OpenAPI, JSON Schema of a "major" version). 
  * There's import and export compatibility, which depends on how the own model was extended:
    * Adding optional attributes: Full import and export compatibility
    * Removing optional attributes: Full import and export compatibility
    * Adding further constraints to existing attributes: Full import and export compatibility
    * Adding new mandatory attribute or making optional attribute mandatory: Full export compatibility, not import.

This is probably too much for the place here? Where would you put content that explains a concept in more detail, to not overload the spec itself?

One open question: Should we also point out a version of the model here? Most specs at least have a <major>.<minor> version, although the compatibility should be sufficiently expressed with just <major>. Maybe a second label? New minor releases can indicate additional attributes / features in the model.

duglin: I suspect we may want to treat these "well-known labels" the same way we treat extensions in CloudEvents. Meaning, anyone can propose new ones at any time and they might even change their definitions over time as people start to use them (or even remove them). So, it might be best to create a 2nd doc that has a different lifecycle than the spec itself and has a much lower bar for edits (and no official versioning). WDYT?

Fannon: Yes, I like the idea! 

We also need to put the "compatiblewith" into the 3 CNCF provided models, so when you import them you already have them by default. That would make this label also more visible. 

I thought of another simplification: How about you can just add any URI, and each model defines which it is? You could also point to the model specification, which may be more user friendly e.g. https://github.com/xregistry/spec/blob/main/endpoint/spec.md

Fannon: @duglin I can see three kinds of well-known labels: 1) labels on the meta-model (like `compatiblewith` and `modelversion` (?)), 2) labels on the instance level (not sure we have any) and 3) labels defined as part of a model, e.g. the endpoint model.

Here is a PR to add it to the extension specs: https://github.com/xregistry/spec/pull/225

core/model.json
```diff
@@ -22,6 +19,12 @@
       "readonly": true,
       "serverrequired": true
     },
+    "xid": {
+      "name": "xid",
+      "type": "url",
```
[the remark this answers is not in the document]
duglin: fixed in json files and spec

core/primer.md
```diff
@@ -477,3 +477,19 @@ Group and Resource names limited to only 58? Because when they appear as
 Collection attributes and we append `url` or `count` to them they still have
 to fit within the 63 character limit, and so we need to take into account
 the length of the word `count`.
+
+# Why must Group type and Resource type names be valid attribute names?
+
+Since Group type and Resource type names will appear as attribute names (e.g.
+`GROUPScount` and `RESOURCEurl`), the character set (and length) of their
+names would need to be restricted in the same way as a attribute names.
```
jskeet, line 485:
```suggestion
names need to be restricted in the same way as attribute names.
```

core/spec.md
```diff
@@ -366,6 +406,11 @@ be one of the following data types:
   - See [Serializing Resource Documents](#serializing-resource-documents)
     for more information about serializing maps as HTTP headers.
 - `object` - a nested entity made up of a set of attributes of these data types.
+- `relation` - a relative URL reference to a type of entity in the Registry.
```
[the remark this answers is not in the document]
jskeet: Drive-by comment, without either knowing the motivation behind the change or having expertise in the difference between URI and URL - if we make this URI after all, we should potentially revert some of the other URI=>URL changes too.

core/spec.md
```diff
@@ -333,6 +366,13 @@ Therefore, the hierarchical structure of the Registry model is defined in such
 a way that it can be represented in a single file, including but not limited
 to JSON, or via the entity graph of a REST API.
 
+If an error occurs during the processing of a request, even if the error was
+during the creation of the response (e.g. an invalid `?inline` value was
+provided), then an error MUST be generated and the entire request (even any
+update actions) MUST be undone. In general, when an error is generated it
```
jskeet, line 372: Is there a transactionality element here, in terms of requiring that no changes applied due to the processing of the request must be observable until the processing has completed? (i.e. you can't just start making changes which could be observed, then revert them later).

core/spec.md
```diff
@@ -333,6 +366,13 @@ Therefore, the hierarchical structure of the Registry model is defined in such
...
+SHOULD be sent back to the client. However, this MAY not happen if the server
+determines there is a good reason to not do so - such due to security concerns.
```
jskeet, line 374:
```suggestion
determines there is a good reason to not do so - such as due to security concerns.
```

core/spec.md
```diff
@@ -333,6 +366,13 @@ Therefore, the hierarchical structure of the Registry model is defined in such
 a way that it can be represented in a single file, including but not limited
 to JSON, or via the entity graph of a REST API.
 
+If an error occurs during the processing of a request, even if the error was
+during the creation of the response (e.g. an invalid `?inline` value was
+provided), then an error MUST be generated and the entire request (even any
+update actions) MUST be undone. In general, when an error is generated it
+SHOULD be sent back to the client. However, this MAY not happen if the server
```
jskeet, line 373: Presumably if an error occurs, the server still has to send *an* error back to the client, but it doesn't need to have full information?

core/spec.md
```diff
@@ -526,11 +572,70 @@ of the existing entity. Then the existing entity can be deleted.
...
+- Type: URL
+- Description: A unique absolute URL for an entity. This attribute MUST be
+  an alternative URL for the owning entity's `self` attribute. The intention
+  is that `shortself` SHOULD be shorter in length than `self` such that it
+  MAY be used when the length of the URL referencing the owning entity is
+  important. For example, in cases where the size of a message needs to be
+  as small as possible.
+
+  This specification makes no statement as to how this URL is constructed,
+  to which host/path it references, or whether an HTTP request to this URL
```
jskeet, line 592: Possibly remove HTTP from here, to allow for other schemes which may also have redirection?

core/spec.md
```diff
@@ -526,11 +572,70 @@ of the existing entity. Then the existing entity can be deleted.
...
+  will directly perform the desired operation or whether it returns an HTTP
+  redirect to the full `self` URL requiring the client to resend the request.
+
+  If an entity is deleted and then a new entity is created that results in
+  the same `self` URL, this specification does not mandate that the same
+  `shorturl` be generated, but it MAY do so.
+
+  This attribute MUST only appear in the serialization if the `shortself`
```
jskeet, line 600: This feels like a potential problem to me - or I may just have misunderstood. Consider the following sequence of events:

- Initial state: a registry with the `shortself` capability enabled
- t0: resource X is created with a long full resource URL, and a shortened URL created using URL shortener 1
- t1: registry has the `shortself` capability disabled (so the first short URL is no longer visible/updatable)
- t2: the provider of URL shortener 1 warns of turndown
- t3: the creator of the entity migrates to URL shortener 2, updating entities in all registies where the `shortself` capability is enabled - but can't update entity X
- t4: URL shortener 1 is turned down
- t5: the registry containing entity X has `shortself` re-enabled

At that point entity X is effectively broken, and the creator couldn't have pre-emptively fixed it.

Maybe this is too obscure an issue to be worth considering.

jskeet: (When writing this comment, I hadn't got to the bit saying it has to be read-only...)

core/spec.md
```diff
@@ -526,11 +572,70 @@ of the existing entity. Then the existing entity can be deleted.
...
+  values. In this sense, implementations might create a `shortself` that is
+  static for the lifetime of the entity and the capability controls whether
+  the attribute is serialized or not.
+
+- Constraints:
+  - MUST be unique across all entities in the Registry.
+  - MUST be a non-empty absolute URL referencing the same entity as the `self`
+    URL, either directly or indirectly via an HTTP redirect.
+  - MUST be a read-only attribute in API view.
+  - MUST NOT change during the lifetime of the entity.
```
jskeet, line 612: I suspect this could cause problems - migrating from one shortening service to another feels like a reasonable use case... although I can see how anything using this effectively as an ID could be confused by changes.

Worth discussing in the subgroup with folks more knowledgeable than myself?

core/spec.md
```diff
@@ -752,8 +845,24 @@ Support for any particular API defined by this specification is OPTIONAL,
...
-If an API is not supported by the server then a `405 Method Not Allowed`
-HTTP response MUST be generated.
+
+If an OPTIONAL HTTP path is not supported by an implementation, then any
+use of that API MUST generate a `404 Not Found` error.
+
+If a HTTP method is not supported for a defined HTTP path then a
+`405 Method Not Allowed` error MUST be generated.
+
+Implementation MAY defined extension APIs however the following rules MUST
```
jskeet, line 855:
```suggestion
Implementations MAY support defined extension APIs however the following rules MUST
```

(Or something similar...)

core/spec.md
```diff
@@ -752,8 +845,24 @@ Support for any particular API defined by this specification is OPTIONAL,
...
+apply:
+- New HTTP paths that extend non-root paths MUST NOT be defined.
+- New root HTTP paths MAY be defined as long as they do not use Registry level
+  HTTP paths or attribute names. This include extension and GROUPS collection
+  attribute names.
+- Additional HTTP methods for specification defined HTTP paths MUST NOT be
+  defined.
+
+For example, a new API with an HTTP path of `/my-api` is allowed, but APIs with
```
jskeet, line 864: Doesn't this mean that the specification can never add any new root paths without that being a breaking change?
I wonder whether it's worth reserving a URI space - maybe even `/extensions/` and saying everything has to be under that?

jskeet (commented):
Only a partial review due to time constraints, but hopefully it's useful.

core/spec.md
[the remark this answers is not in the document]
duglin: the general pattern, I believe, we follow is that URLs are for HTTP GET-able addresses, while URIs are for addresses that could be concrete (like URLs) but could also be abstract and not HTTP GET-able. I believe @deissnerk wants the option for "relations" to be abstract in the future. Right Klaus?

@jskeet do you think some of our URLs need to be URIs?

core/primer.md
duglin: fixed

core/spec.md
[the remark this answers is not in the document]
jskeet: I have no opinion on what should be which - I just noted that there's a change earlier on.

I would also note that URL doesn't imply HTTP - if this is the expectation for xregistry, it may be worth documenting. (That might already be documented of course.)

core/spec.md
duglin: I kind of think there has to be, at least on a per API call basis, otherwise we run the risk of people seeing a Registry that's in a bad/invalid state.

Perhaps we could add a section to the spec (or primer) to talk about how:
- our expectation is that there's a transaction system in place where each individual API call is processed within the scope of one Tx such that if something goes wrong during ANY of its processing then entire thing can be rolled back
- this also then implies that any changes being made within that one Tx are not visible to other API calls (Txs) until it is committed 
- if someone wants to expose parts of their "desired set of changes" in an incremental fashion, they can do so by breaking up the work into individual API calls. 

Thoughts?

core/spec.md
duglin: fixed

core/spec.md
duglin: I think we need to leave it as an impl choice since there are some systems that might think it's best to just close the connection rather than even give any error message at all - again for security reasons. I think this could be one of those "self correcting issues" where if people make poor choices then people will stop using that system. This is why I prefer to use the phrase "generate an error" rather than "send (or respond with) an error"... keeps the spec flexible.

core/spec.md
duglin: done

core/spec.md
duglin: Yup, this is a server controlled value so any updates to it would need to be done by the server, not the owner/creator of the entity. So the scenario as you described it is a bit wonky... however, you do raise a good point about switching to a new URL shortener service. This text was added because changing the shortself, in general, is dangerous and so I wanted it to be as static as "self", but I just realized that we don't even say that "self" is static...and it can't be for a variety of reasons. I was just trying to avoid people changing "shortself" randomly.

So, how about this:
- I'll remove the "- MUST NOT change during the lifetime of the entity." and change "static" to "known" in this paragraph... just to loosen things
- I'll add text to the Primer about how both "self" and "shorturl" are meant to be static for the lifetime of the owning entity, however there may be situations where they need to change... moving hosts, changing short-url service... so while the spec doesn't ban changing them, care should be taken when doing so since any reference using the old values might break people

I've gone ahead and made these changes since I think it's at least better than what was there before your review.

thoughts?

core/spec.md
duglin: See previous comment.... good catch!

core/spec.md
duglin: done

core/spec.md
duglin: Why would adding a new root path to the spec be a breaking change, as long as it's optional ?

Or instead of "breaking" did you mean "potentially conflicting with an extension" ? Yes that can happen, but that's true of all extensions (even attribute extensions). It's just the nature of how extensions work... unless we create "extension buckets/spaces" as you suggest. But then we're back to this (very long winded) "X-" prefix on http header extensions rathole ;-)

core/spec.md
jskeet: Sounds good - I'd misinterpreted who provided the shorturl; if it's server-provided, that's much simpler.

duglin:
As always... thanks for the review @jskeet 

And yes I did think about/try to create separate commits but I quickly realized that I kept needing to make updates to previously completed changes and I don't think I can update an old commit w/o major pain.

luckily, most changes do not really overlap so hopefully it wasn't too painful of a review (aside from the size of it).

core/spec.md
jskeet: > Or instead of "breaking" did you mean "potentially conflicting with an extension" ?

Yes, as I view that as a breaking change - it means that doing something that *was* valid is no longer valid. Surely that's reasonable to count as "breaking".

And yes, this is like the CloudEvents discussion we had - basically I personally thing CE went the wrong route here, as I think we've now come to the conclusion we can't add *any* core attributes without taking a 2.0. I'd like xregistry to avoid getting into that situation. (I seem to remember writing a lot about this before - https://docs.google.com/document/d/1n91IEyB8gD8JSBiAYyYo8YbDne4KD3YvCa5BrdemZt4/edit?tab=t.0#heading=h.yflgqnc5ygl0 may be what I'm thinking of, or there could be something else.)

core/spec.md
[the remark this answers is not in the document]
duglin: I think we need to make this a bit clearer... we don't want to mandate http, but it is the only protocol we support today and we've been a bit lazy about using generic wording vs falling into the "we're using http" world.

I added this issue: https://github.com/xregistry/spec/issues/230 to think more about this.

duglin:
Notes from 1/9:
- describe 'xid' as the ID within the registry, not focus on self-path

duglin:
Approved on the 1/9 call

PR #225, opened 2024-12-31
Add compatiblewith and modelversion label to domain specific specs

Fannon:
Relates #222 

Unfortunately, I did not find out how I could add a well-defined model into the `model.json` files. Is this possible? A problem could be that it's just an extension bag of properties and now we want to define at least one label inside of it.

Btw. I have troubles running the makefile locally on a Windows machine. Would need to install Linux Subsystem to make it work, I guess. But I've added an https://editorconfig.org/ file that tells editors / IDEs how to handle spacing etc. by default.

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -20,6 +20,7 @@ automation and tooling.
   - [Registry Entity](#registry-entity)
     - [Retrieving the Registry](#retrieving-the-registry)
     - [Updating the Registry Entity](#updating-the-registry-entity)
+  - [Registry Capabilities](#registry-capabilities)
   - [Registry Model](#registry-model)
     - [Retrieving the Registry Model](#retrieving-the-registry-model)
     - [Updating the Registry Model](#updating-the-registry-model)
@@ -39,8 +40,10 @@ automation and tooling.
     - [Creating or Updating Versions](#creating-or-updating-versions)
     - [Retrieving a Version](#retrieving-a-version)
     - [Deleting Versions](#deleting-versions)
-  - [Inlining](#inlining)
-  - [Filtering](#filtering)
+  - [Configuring Responses](#configuring-responses)
+    - [Inline](#inline)
+    - [Filter](#filter)
+    - [Compact](#compact)
   - [HTTP Header Values](#http-header-values)
 
 ## Overview
@@ -95,6 +98,8 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "specversion": "STRING",
   "registryid": "STRING",
   "self": "URL",
+  "shortself": "URL", ?
+  "xid": "URI",
   "epoch": UINTEGER,
   "name": "STRING", ?
   "description": "STRING", ?
@@ -103,12 +108,32 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "createdat": "TIMESTAMP",
   "modifiedat": "TIMESTAMP",
 
-  "model": {                            # Only if inlined/nested
-    "schemas": [ "STRING" * ], ?        # Available schema formats
+  "capabilities": {                     # Supported capabilities/options
+    "enforcecompatibility": BOOLEAN, ?
+    "flags": [                          # Query parameters
+      "epoch",? "compact",? "filter",? "inline",?
+      "nodefaultversionid",? "nodefaultversionsticky",? "noepoch",?
+      "noreadonly",?  "schema",? "setdefaultversionid",? "specversion",?
+      "STRING" *
+    ],
+    "mutable": [                        # What is mutable in the Registry
+      "capabilities",? "entities",? "model",? "STRING"*
+    ], ?
+    "pagination": BOOLEAN, ?
+    "schemas": [ "xRegistry-json/0.5", "STRING" * ], ?
+    "shortself": BOOLEAN, ?
+    "specversions": [ "0.5", "STRING"* ], ?
+
+    "STRING": ... *                     # Extension capabilities
+  }, ?
+
+  "model": {                            # Only if inlined
+    "labels": { "STRING": "STRING" * }, ?
     "attributes": {                     # Registry level attributes/extensions
       "STRING": {                       # Attribute name (case sensitive)
         "name": "STRING",               # Same as attribute's key
         "type": "TYPE",                 # string, decimal, array, object, ...
+        "target": "STRING", ?           # If "type" is "relation"
         "description": "STRING", ?
         "enum": [ VALUE * ], ?          # Array of scalar values of type "TYPE"
         "strict": BOOLEAN, ?            # Just "enum" values or not.Default=true
@@ -137,6 +162,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
       "STRING": {                       # Key=plural name, e.g. "endpoints"
         "plural": "STRING",             # e.g. "endpoints"
         "singular": "STRING",           # e.g. "endpoint"
+        "labels": { "STRING": "STRING" * }, ?
         "attributes": { ... }, ?        # Group level attributes/extensions
 
         "resources": {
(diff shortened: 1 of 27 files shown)


duglin:
> Unfortunately, I did not find out how I could add a well-defined model into the `model.json` files. Is this possible? A problem could be that it's just an extension bag of properties and now we want to define at least one label inside of it.

I'm not following. What do you mean by "add a well-defined model into model.json" ? Do you mean "how to include a list of possible values w/o actually adding them as labels?"

> Btw. I have troubles running the makefile locally on a Windows machine. Would need to install Linux Subsystem to make it work, I guess. But I've added an https://editorconfig.org/ file that tells editors / IDEs how to handle spacing etc. by default.

Is it because of the linux shell commands that are run?

endpoint/spec.md
```diff
@@ -33,7 +33,11 @@ this form:
   "name": "STRING", ?
   "description": "STRING", ?
   "documentation": "URL", ?
-  "labels": { "STRING": "STRING" * }, ?
+  "labels": {
+    "modelversion": "1.0",
+    "compatiblewith": "https://github.com/xregistry/spec/blob/main/endpoint/spec.md",
```
duglin, line 38: This feels a bit odd... sort of like saying "I'm compatible with myself".
Is this trying to define the "compatiblewith" value that extensions models are meant to use?

Fannon: Think of it like `"openApi": "3.1"` in OpenAPI or `"asyncAPI": "2.1"` in AsyncAPI. 

Then vendors like MSFT or SAP can create their own extensions of such standards, but there they will also point out that it's still a compatible e.g., OpenAPI 3.1 document.

Somehow you need to state that this a document follows a model and that the model is actually an interoperable standard. And yes, the base model is compatible with itself, but "compatiblewith" sounds a bit unusual.

duglin: Right, so I would think that in endpoint's model.json the label name might be something like "self" instead since we're not really trying to claim any kind of compatibility statement. But it then provides the value that extension models can use in their "compatibilewith" label... if they don't know where it lives. Or am I not following?

Fannon: Yes, it's indeed a "self" statement, but "compatible" is also true. If you deploy that model as is, it's basically compatible with the standard model, it just doesn't have any extensions and modifications. 

But I wonder if when you import such a model, you want to keep this "compatiblewith" statement by default. If it's not there, you have no guarantuee that you understand that model in someone elses xRegistry.

duglin:
@Fannon we talked about this on today's call and we're ok with it. Is it ready to merge or do you want to make more changes?

Fannon:
> @Fannon we talked about this on today's call and we're ok with it. Is it ready to merge or do you want to make more changes?

I'm not sure if we can define the labels like this. Labels is first an open collection of key, values right? How could we define in the meta-model that some keys have a pre-defined meaning?

duglin:
Can we not just define them as part of the core spec where we define "compatiblewith" ? Or we can create a dedicated section (or even new doc) for them to live

schema/spec.md
```diff
@@ -143,13 +147,20 @@ being created.
 In "semantic Versioning" terms, you can think of a **schema** as a "major
 version" and the **schema Versions** as "minor versions".
 
+## Schema Registry Model
+
+The formal xRegistry extension model of the Schema Registry resides in the
+[model.json](model.json) file.
+
+By importing and keeping the `compatiblewith` label, interoperability on the
+CNCF defined endpoint model is stated. See Interoperability **TODO: link**.
```
duglin, line 156: where did you see this "Interoperability" link pointing to?

Fannon: I wasn't sure if we need a dedicated section to explain this further.  
If not, we can remove this todo for now

```suggestion
CNCF defined endpoint model is stated.
```

core/spec.md
```diff
@@ -2101,8 +2101,11 @@ The following describes the attributes of Registry model:
   - Values MAY be empty strings.
   - The following labels are defined for convenience and MAY also be used
     as labels for Groups or Resource model types:
+    - Name: `modelversion`
+      Value: The version of the model. For many specs its a combination of
+      major and minor, e.g. `1.2`.
     - Name: `compatiblewith`
-      Value: A URI reference to Registry that this Registry's model is
+      Value: An absolute URL to the model specification that this model is
```
Fannon, line 2108: @duglin suggestion for different phrasing, using the new URL terms. I think this has to be absolute URL.

core/spec.md
```diff
@@ -2101,8 +2101,11 @@ The following describes the attributes of Registry model:
   - Values MAY be empty strings.
   - The following labels are defined for convenience and MAY also be used
     as labels for Groups or Resource model types:
+    - Name: `modelversion`
```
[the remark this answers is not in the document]
duglin: it's easier is you just squash the commits down to one and sign that one

core/spec.md
duglin: ok I'll pick this up for the other PR- then we don't need to worry about rebasing and you can drop this file from your PR.

Fannon: I've squashed it now as you suggested

core/spec.md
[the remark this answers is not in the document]
Fannon: Ok, it was easier. I was getting merge conflicts that I didn't understand otherwise.
Now pipeline is happy

duglin:
@Fannon rebase is needed - it would be easier if you just deleted the core spec file

duglin:
unless you want to make more changes

schema/spec.md
```diff
@@ -150,13 +154,20 @@ being created.
 In "semantic Versioning" terms, you can think of a **schema** as a "major
 version" and the **schema Versions** as "minor versions".
 
+## Schema Registry Model
+
+The formal xRegistry extension model of the Schema Registry resides in the
+[model.json](model.json) file.
```
duglin, line 160: Just wondering, why is the above line only in the schema model and not the others?

duglin: And should these labels be in those?

Also, I just realized, we should probably add `?` to the end of each label (in all docs) since w/o that people will think they're required labels.

Fannon: yes, they should have ` ?` suffix. 

> Just wondering, why is the above line only in the schema model and not the others?

I thought It was already in the other files and missing here, so I added it for consistency.

schema/spec.md
```diff
@@ -34,7 +34,11 @@ this form:
   "name": "STRING", ?
   "description": "STRING", ?
   "documentation": "URL", ?
-  "labels": { "STRING": "STRING" * }, ?
+  "labels": {
+    "modelversion": "1.0",
+    "compatiblewith": "https://github.com/xregistry/spec/blob/main/schema/spec.md",
```
Fannon, line 39:
```suggestion
    "modelversion": "1.0", ?
    "compatiblewith": "https://github.com/xregistry/spec/blob/main/schema/spec.md", ?
```

@duglin  like this?

endpoint/spec.md
```diff
@@ -34,7 +34,11 @@ this form:
...
```
Fannon, line 39:
```suggestion
    "modelversion": "1.0", ?
    "compatiblewith": "https://github.com/xregistry/spec/blob/main/endpoint/spec.md", ?
```

message/spec.md
```diff
@@ -39,7 +39,11 @@ this form:
   "name": "STRING", ?
   "description": "STRING", ?
   "documentation": "URL", ?
-  "labels": { "STRING": "STRING" * }, ?
+  "labels": {
+    "modelversion": "1.0",
+    "compatiblewith": "https://github.com/xregistry/spec/blob/main/message/spec.md",
```
Fannon, line 44:
```suggestion
    "modelversion": "1.0", ?
    "compatiblewith": "https://github.com/xregistry/spec/blob/main/message/spec.md", ?
```

duglin:
Did you do a rebase on the latest branch? The last two commits look like they’re from “main”. There PR should only show your edits

schema/spec.md
duglin: Yet thanks

schema/spec.md
duglin: Oh interesting, I never really noticed it before. Thanks

duglin:
@Fannon can you squash and rebase this PR? I'm nervous about it including commits from the 'main' branch. It's probably ok but I'd prefer a clean PR that's purely additive and doesn't include changes not related to you stuff.

Fannon:
> @Fannon can you squash and rebase this PR? I'm nervous about it including commits from the 'main' branch. It's probably ok but I'd prefer a clean PR that's purely additive and doesn't include changes not related to you stuff.

@duglin Now I have to rebase again, because DCO troubles coming back. But it should be no problem to have commits here from main, because that always happen when main goes forward and the PR gets updated to latest state. I think you can configure github to only allow merges if 1) they are on latest sync with main and 2) squash merged. Then you should get a nice clean single commit into main. I found this workflow to be easier than rebasing and squashing in the PR branch. 

I'll have a look tomorrow, when back on my work laptop.

duglin:
LOL I'm not sure what happened but now your *1* commit has 27 files changed and I'm pretty sure not all of them are part of this work.

Fannon:
> LOL I'm not sure what happened but now your _1_ commit has 27 files changed and I'm pretty sure not all of them are part of this work.

Sorry, I'm really not used to work across forks. It should be OK now.

duglin:
thanks! Merging

PR #245, opened 2025-01-24
Week of Jan 27

### LINKED ISSUE #241: Is our defaultversion logic correct?
Today the spec says this:
- - -
Since this attribute and defaultversionid are closely related, the processing of them in a request message MUST adhere to the following:

- The defaultversionsticky attribute is applied first. As a reminder, in the PATCH case, if this attribute is missing in the request then this attribute remains unchanged in the Resource.
- If the resulting value of this attribute is false then the sticky aspect MUST be turned off, and any defaultversionid in the request MUST be ignored. The newest Version MUST be the default Version.
- If the resulting value of this attribute is true then the sticky aspect MUST be turned on, and any defaultversionid attribute from the request is applied - where a value of null means "newest". If the request was a PUT or POST and did not have a defaultversionid then the implicit value of null MUST be used, resulting in "newest". If the request was a PATCH and did not have a defaultversionid then the current default Version MUST be used. A reference to a Version that does not exist MUST generate an error.
- - -
Is this ok?

Another thing to consider: `POST .../rID?setdefaultversionid=v5` will turn on the sticky bit.

Should the logic for modifying the meta attributes be:
- presence of "defaultversionid" in a request w/o "defaultversionsticky" has an implied "defaultversionsticky:true" (any existing "defaultversionsticky" is ignored)
- presence of "defaultversionid" + "defaultversionsticky:false" -> ignore "defaultversionid" / sticky=false

I'm wondering if it's a bit odd (or unfair) that people can set defaultversionid via a query parameter w/o having to explicitly turn on sticky, but via JSON they need to send both parameters?

### LINKED ISSUE #242: replace "relation" with "xid"?
since they're both the same.

### LINKED ISSUE #243: Should ?compact remove $details from URLs?
e.g.
```
{
  "fileid": "f1",
  "self": "/dirs/d1/files/f1$details",
  "xid": "/dirs/d1/files/f1",

  "metaurl": "http://xregistry.soaphub.org/reg-TestRegistry/dirs/d1/files/f1/meta",
  "versionsurl": "http://xregistry.soaphub.org/reg-TestRegistry/dirs/d1/files/f1/versions",
  "versionscount": 2
}
```
Notice "self" has "$details", which make sense when it's an absolute URL. But as a relative URL, trying to point to something in this doc, I'm not sure it makes a lot of sense. Or does including it provide consistency?

@clemensv what do you think tooling would prefer?

### LINKED ISSUE #246: Rename ?compact to ?doc and have a clean split between API and Doc views
Today the definition of a spec-defined attribute looks like this:
- - -
##### `self` Attribute

- Type: URL
- Description: A server generated unique URL for the current entity. In the case of pointing to an entity in a [Registry Collection](#registry-collections), the URL MUST be a combination of the base URL for the collection appended with the `SINGULARid` of the entity.

  When this URL references a Resource or Version, and the Resource has the `hasdocument` model aspect set to `true`, and `?compact` mode is not enabled, then it MUST include `$details` appended to its `SINGULARid` if the request asked for the serialization of the xRegistry metadata. This would happen when `$details` was used in the request, or when the Resource (or Version) is included in the serialization of a parent entity. If `?compact` mode is enabled, then `$details` MUST appear when this URL is an absolute URL, otherwise for relative URLs it MUST NOT appear.

  If the Resource or Version has the `hasdocument` model aspect set to `false`, then this URL MAY be appended with `$details` only if the request was directed at the Resource or Version and included `$details`. Otherwise in all other situations `$details` MUST NOT be included.

- Constraints:
  - MUST be unique across all entities in the Registry.
  - MUST be a non-empty absolute URL, except when `?compact` is enabled
    then it MUST be a relative URL.
  - MUST be a read-only attribute in API view.
  - REQUIRED in responses and document view, OPTIONAL in requests.
- Examples:
  - `https://example.com/registry/endpoints/ep1`
- - -

I'd like to propose that we have a more clear split between "api view" and "document view". The above text (sort of) jumps between the two and it might be hard to following. So what if we had 3 sections to the definition:
1 - a generic section that talks about what the attribute is for - and this applies to both api and doc view
2 - an API view list of constraints
3 - a Doc view list of constraints

For example:
- - -
##### `self` Attribute

- Type: URL
- Description: A server generated unique URL for the current entity. Its full value MUST be an absolute URL based on the URL of th Registry root appended with the hierarchy of the Registry entities/collections leading to the entity.

- General Constraints:
  - Its full URL MUST be unique across all entities in the Registry.
  - MUST be a read-only attribute.

- API View Constraints:
  - MUST be a non-empty absolute URL based on the URL of the Registry.
  - When serializing Resources, if the `hasdocument` aspect set to `true`, then this URL MUST include the `$details` suffix to its `SINGULARid` if it is serialized in the HTTP body response. If the aspect is set to `false`, then this URL's `SINGULARid` MAY be appended with `$details` only if the request was directed at a particular Resource/Version and included `$details` in the request URL.
  - REQUIRED in responses.
  - OPTIONAL, and ignored, in requests.
 
- Document View Constraints:
  - MUST be a relative URL, relative to the URL used to generate the document.
  - MUST NOT include the `$details` suffix on its `SINGULARid`.
  - REQUIRED.

- Examples:
  - `https://example.com/registry/endpoints/ep1`
- - -

And, since we tend to mix the topics of "doc view" and "compact", I think it might reduce confusion/complexity if we renamed `?compact` to `?doc`. I'm not sure renaming `/export` to `/doc` makes as much sense since `/export` is an alias for more than just `?compact`.

duglin:
- make it clear that relative URLs need to be based off of the of the request.
- $details should NOT be included in ?compact/document view
- s/?compact/?doc/g and add more text explaining "document view"
- tweaks our defaultversion* processing logic to make it more user friendly
- replace RELATION type of XID type - since they're the same definition

Fixes #243
Fixes #241
Fixed #246
Fixed #242

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -43,7 +43,7 @@ automation and tooling.
   - [Configuring Responses](#configuring-responses)
     - [Inline](#inline)
     - [Filter](#filter)
-    - [Compact](#compact)
+    - [Doc](#doc)
   - [HTTP Header Values](#http-header-values)
 
 ## Overview
@@ -111,7 +111,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "capabilities": {                     # Supported capabilities/options
     "enforcecompatibility": BOOLEAN, ?
     "flags": [                          # Query parameters
-      "epoch",? "compact",? "filter",? "inline",?
+      "doc",? "epoch",? "filter",? "inline",?
       "nodefaultversionid",? "nodefaultversionsticky",? "noepoch",?
       "noreadonly",?  "schema",? "setdefaultversionid",? "specversion",?
       "STRING" *
@@ -133,7 +133,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
       "STRING": {                       # Attribute name (case sensitive)
         "name": "STRING",               # Same as attribute's key
         "type": "TYPE",                 # string, decimal, array, object, ...
-        "target": "STRING", ?           # If "type" is "relation"
+        "target": "STRING", ?           # If "type" is "xid"
         "description": "STRING", ?
         "enum": [ VALUE * ], ?          # Array of scalar values of type "TYPE"
         "strict": BOOLEAN, ?            # Just "enum" values or not.Default=true
@@ -317,7 +317,6 @@ information about each data type):
 - `INTEGER`
 - `MAP`
 - `OBJECT`
-- `RELATION`
 - `STRING`
 - `TIMESTAMP`
 - `UINTEGER`
@@ -325,6 +324,7 @@ information about each data type):
 - `URIREFERENCE`
 - `URITEMPLATE`
 - `URL`
+- `XID`
 - `TYPE` - one of the allowable data type names (MUST be in lower case) listed
   in [Attributes and Extensions](#attributes-and-extensions)
 
@@ -376,9 +376,28 @@ concerns.
 
 In the remainder of this specification, in particular when defining the
 attributes of the Registry entities, the terms "document view" or "API view"
-will be used to indicate whether the serialization of the entity in question
+will be used to indicate whether the serialization of the entity in a response
 is meant for use as a stand-alone document or as part of a REST API message
-exchange - if there is a difference between the two usages.
+exchange. The most notable differences are that in document view:
+
+- References (e.g. URLs) between entities within responses will use relative
+  references rather than absolute ones. This is an indication to tooling that
+  the entity in question can be found within the current document and does not
+  need an additional HTTP `GET` to retrieve it.
+- Duplicate data will be removed. In particular, Resources will not include
+  attributes from the "default" Version as part of their serialization. This is
+  done with the assumption the nested `versions` collection will most likely
+  include the "default" Version, so duplicating that information is redundant.
+
+Most of these differences are to make it easier for tooling to use the
+"stand-alone" document view of the Registry. For a complete list of the
+differences in "document view" see the [Doc](#doc) flag and the
+[Exporting](#exporting) section.
+
+Note that "document view" only refers to response messages. There is no
+"document view" concept for requests. However, "document view" responses
+are designed such that they can be used in request messages as they still
+convey the same information as an "API view" response.
 
 ### Attributes and Extensions
 
@@ -407,7 +426,7 @@ be one of the following data types:
   - See [Serializing Resource Documents](#serializing-resource-documents)
     for more information about serializing maps as HTTP headers.
 - `object` - a nested entity made up of a set of attributes of these data types.
-- `relation` - MUST be a `URL` (xid) reference to another entity defined within
+- `xid` - MUST be a `URL` (xid) reference to another entity defined within
   the Registry. The actual entity attribute value MAY reference a non-existing
   entity (i.e. dangling pointer), but the syntax MUST reference a
   defined/valid type in the Registry. This type of attribute is used in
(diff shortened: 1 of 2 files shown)


core/spec.md
```diff
@@ -376,9 +376,28 @@ concerns.
...
 is meant for use as a stand-alone document or as part of a REST API message
-exchange - if there is a difference between the two usages.
+exchange. The most notable differences are that in document view:
+
+- References (e.g. URLs) between entities within responses will use relative
+  references rather than absolute ones. This is an indication to tooling that
+  the entity in question can be found within the current document and does not
+  need an additional HTTP `GET` to retrieve it.
+- Duplicate data will be removed. In particular, Resources will not include
+  attributes from the "default" Version as part of its serialization. This is
```
jskeet, line 388:
```suggestion
  attributes from the "default" Version as part of their serialization. This is
```

(Or change "Resources" to "a Resource".

core/spec.md
```diff
@@ -427,9 +446,27 @@ be one of the following data types:
   [RFC 1738](https://datatracker.ietf.org/doc/html/rfc1738).
 
 All relative URIs and URLs that reference entities within the Registry MUST
-begin with `/` and MUST be relative to the root path of the Registry service.
+begin with `/`. With the exception of `xid` values, relative URIs/URLs that
+reference entities that are part of a document view `GET` response (`?doc`
+is enabled), MUST be relative to the path specified on the request. Relative
```
jskeet, line 451: I don't understand this... if it's meant to be "relative to the path" why would it start with a "/"?

My understanding is:

- If the URL begins with a "/", it's an absolute *path*, but relative to the originating server
- If a URL *doesn't* begin with a "/" then it's a relative path

So a "relative URL beginning with a `/` but relative to the path specified on the request" sounds like a contradiction in terms to me.

core/spec.md
```diff
@@ -427,9 +446,27 @@ be one of the following data types:
...
+`xid` values MUST be relative to the root of Registry regardless of the path
+specified on the request.
+
+For clarity, if a Registry has a Schema Resource at
+`/schemagroups/g1/schemas/s1`, then the entity's `self` URL (when serialized
+in document view) would change based on the path specified on the `GET`
+request:
+
+| GET Path | `self` URL |
```
jskeet, line 460: This table makes it clear that yes, there really is a disconnect here between my understanding of relative URLs and what's being stated.

I'm thinking mostly in HTML terms - where if you fetch (say) `http://example.com/myreg/schemagroups/g1/` then link to `/schemas/s1`, that would refer to `http://example.com/schemas/s1`, *not* `http://example.com/myreg/schemagroups/g1/schemas/s1`.

I think it's worth seeing whether other people would have the same expectation as me, or whether I'm just plain wrong and no-one else is going to be confused.

core/spec.md
```diff
@@ -3772,16 +3814,14 @@ and the following Resource level attributes:
 - Type: URL
 - Description: a URL to the default Version of the Resource.
 
-  The Version URL path MUST include `$details` if the request asked for
-  the serialization of the Resource metadata. This would happen when
-  `$details` was used in the request, or when the Resource is included in the
-  serialization of a Group, such as when the `?inline` feature is used.
+  If the Resource's `hasdocument` aspect set to `true`, then the Version URL
+  path MUST include the `$details` suffix, unless `?doc` is enabled and
+  the Version is included in the output, then `$details` is left off.
```
jskeet, line 3819:
```suggestion
  the Version is included in the output, then `$details` is omitted.
```

(Just a tone suggestion.)

core/spec.md
```diff
@@ -3808,22 +3848,33 @@ and the following Resource level attributes:
...
-      MUST be ignored. The newest Version MUST be the default Version.
-    - If the resulting value of this attribute is `true` then the sticky
-      aspect MUST be turned on, and any `defaultversionid` attribute from
-      the request is applied - where a value of `null` means "newest".
-      If the request was a `PUT` or `POST` and did not have a
-      `defaultversionid` then the implicit value of `null` MUST be used,
-      resulting in "newest". If the request was a `PATCH` and did not have a
-      `defaultversionid` then the current default Version MUST be used.
-      A reference to a Version that does not exist MUST generate an error.
+  - Since this attribute and `defaultversionid` can impact each other, their
```
jskeet, line 3851: Just a meta comment really... if I were implementing this, I'd have two simultaneous thoughts:

- "I'm really grateful this specified in such a detailed way. I can write tests based on this."
- "Wow, it's really complicated. Does it need to be this complicated?"

core/spec.md
duglin: done - went with "their"

core/spec.md
duglin: That's why I decided to "hammer the point home" with a table because it's not like normal HTML processing. And it kind of can't be... with a "/" it would mean (as you said) relative to the host/server, which isn't correct since the registry could be hosted on a path, not at the root. W/o a "/" means "relative to this spot in the doc" and in particular (w/o a bunch of ".."s) it's a child. And that doesn't work for non-children paths.

We basically landed at a spot where it's not really that different from normal HTML processing, except instead of "/..." being relative from "the host", since there is no host, it's relative from "the root of the doc". Which, if you think about it isn't too different. I think we just need to find the right words.

Notice, that all relative URLs start with a "/" so there is no "relative to this spot in the tree" concept. 

Any of this make sense?

core/spec.md
duglin: done

duglin:
Approved on the 2/6 call with the following changes:
- make relative URLs/URIs start with "#/" not "/"
- clean-up the defaultversionid text so it doesn't seem too complex

PR #253, opened 2025-02-07
missed some xid changes

duglin:
More URL->XID changes & forgot to fix-up the tooling


duglin:
Implicitly approved this past thursday - merging

PR #254, opened 2025-02-07
Add meeting time & meeting notes to README

### LINKED ISSUE #84: An outsiders view on xRegistry

manuelottlik:
Fixes #84

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,13 @@ Please see the
 [meeting minutes doc](https://docs.google.com/document/d/1OVF68rpuPK5shIHILK9JOqlZBbfe91RNzQ7u_P7YCDE/edit#)
 for the latest information on how to join the calls.
 
+The working group also meets
+* every Tuesday & Wednesday at 9AM ET (USA Eastern Time)
+* and every Friday at 8:30AM ET (USA Eastern Time)
+
+to develop xRegistry specifically. For those meetings, please see the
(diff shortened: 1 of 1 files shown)


duglin:
Approved on the 2/13 call

PR #255, opened 2025-02-07
Add RC1, RC2 and V1 to Roadmap

### LINKED ISSUE #250: Update the roadmap
It makes the project look dead even though its not – and by now we might even have a rough idea of what we plan to do in the upcoming year.

manuelottlik:
Fixes #250

### DIFF
--- a/docs/ROADMAP.md
+++ b/docs/ROADMAP.md
@@ -17,14 +17,16 @@ _Setup_ - Completed - 2023/04/20
 _Release Candidate 1_ - WIP - 2025/02/28
 
 * [Core
-  features]((https://github.com/xregistry/spec/issues?q=is%3Aissue%20state%3Aopen%20label%3Av1))
+  features](https://github.com/xregistry/spec/issues?q=is%3Aissue%20state%3Aopen%20label%3Av1)
   completed
 * Ready for implementation testing and thorough review of the spec
 
 _Release Candidate 2_ – WIP – 2025/04/30
 
 * Tested with at least one implementation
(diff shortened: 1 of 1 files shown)


docs/ROADMAP.md
```diff
@@ -14,10 +14,18 @@ _Setup_ - Completed - 2023/04/20
...
+* [Core
+  features](https://github.com/xregistry/spec/issues?q=is%3Aissue%20state%3Aopen%20label%3Av1)
+  completed
+* Ready for implementation testing and thorough review of the spec
 
-_Future_
+_Release Candidate 2_ – WIP – 2025/04/30
 
-- TBD.
+* Tested with at least one implementation
```
lailabougria, line 26: Who would be the implementing party? Inside the group or outside?

lailabougria:
Maybe it could be clarified if the current roadmap applies to the core spec or the deriving ones too?

duglin:
> Maybe it could be clarified if the current roadmap applies to the core spec or the deriving ones too?

Done

docs/ROADMAP.md
duglin: At some point we should list them - point to our reference impl and mention which companies are implementing it. But we can add that later.

lailabougria (approved):

jskeet (approved):
(End of June for GA feels a little optimistic to me given that changes have still been happening on the spec, but I'm conservative like that...)

duglin:
Approved on the 2/13 call

PR #258, opened 2025-02-11
changes for week of Feb10

### LINKED ISSUE #221: Should there be a way for a client to ask for the list of all possible capability values?
Not to know what the spec defines, but rather what this particular server is willing to support. Could be really useful to know which extensions it supports too.

e.g. `GET /capabilities?supported` returns a "capabilities" object will all possible values filled-in and all supported boolean properties set to "true" to indicate support.

```
  "capabilities": {
    "enforcecompatibility": true,
    "flags": [
      "epoch", "compact", "filter", "inline",
      "nodefaultversionid", "nodefaultversionsticky", "noepoch",
      "noreadonly",  "schema", "setdefaultversionid", "specversion"
    ],
    "mutable": [
      "capabilities", "entities", "model"
    ],
    "pagination": false,
    "schemas": [ "xRegistry-json/0.5" ], 
    "shortself": true,
    "specversions": [ "0.5" ]

    "STRING": ... *                     # Extension capabilities
  }
```

### LINKED ISSUE #223: Should we support PATCH for capabilities, model and/or extensions attributes?

### LINKED ISSUE #240: Replace URL, URI and URI-reference types with just URI
Today we define these types:
- URI - absolute URI
- URI-REFERENCE - absolute or relative URI
- URL - absolute URL

While from a spec authoring perspective there are reasons for all 3, most people tend to use one term (either URL or URI) when talking about a "reference", so asking people to choose (and think hard) about which one to use when in most cases they "just want a reference to something else", might be making things more complicated than it needs to be.

So, we're proposing this:
- have just URI as the one type name to use for references (excluding 'xid' types, which is a subset of URI)
- without any additional constraints specified, it'll be defined as either an absolute URI, absolute URL or URI-reference. User can choose what value to use
- definition of a "URI" attribute may choose to constrain allowable values if they wish (e.g. "self" MUST  be an absolute URL)

You could think of this as getting rid of URI and URL and renaming URI-reference to URI

@jskeet - any thoughts on this?

### LINKED ISSUE #248: Should we remove "server's choice" for default version algorithm?
We lean, heavily, on the idea of "default = newest" when sticky is off. Should we stop that and instead also talk about the "server's choice", or just remove "server's choice" ?

### LINKED ISSUE #251: Feedback from readthrough - Part 1
Not entirely through the spec yet, but thought to share the feedback I have already.

## Questions

- [ ] Under https://github.com/xregistry/spec/blob/main/core/spec.md#modifiedat-attribute, a question about the following section: 
>  If present in a write operation request, the following applies:
>  -  If the request value is the same as the existing value, then the current date/time MUST be used as its new value.
>  -  If the request value is different than the existing value, then the request value MUST be used as its new value.
>  -  If the request value is null then the current date/time MUST be used as the new value.

Following the above, if you set the modifiedat to a value before the createdat, it would accept that value. Is that intentional?
Is it allowed to set the modifiedat to a date in the future?

- [ ] Under https://github.com/xregistry/spec/blob/main/core/spec.md#modifiedat-attribute, a question about the following section: 
> In cases where the modifiedat attribute is set to the current date/time on multiple entities within the same operation, the same value MUST be applied to all of the entities.
  
What happens if there are differences between the dates set? Should the operation be rejected?

- [ ] Under https://github.com/xregistry/spec/blob/main/core/spec.md#registry-apis

> Unknown query parameters SHOULD be silently ignored by servers. This includes specification defined but unsupported query parameters. This includes query parameters defined in this specification, which are unsupported by the server.

What was the intent of this? If it's ignored, how can it be surfaced as an issue to the client?

- [ ] Under https://github.com/xregistry/spec/blob/main/core/spec.md#registry-model

> Implementations are REQUIRED to ensure that after any model changes are made, all of the entities in the Registry are valid with respect to the new model definition (including all Versions of Resources). How this is achieved will be implementation specific. For example, implementations can choose to automatically modify existing entities, or even to delete non-conforming entities (such as when Groups or Resource types are removed). However, it is STRONGLY RECOMMENDED that implementations not delete entities due to attribute modifications.

It sounds as if this could lead to unpredictable outcomes when updating, because implementations may be different. If I understand correctly, especially in the derived specifications, you can cross-reference entities from another registry, how would this affect such a scenario? How can the changes that will be applied be foreseen for validation?

## Nitpicks

- [ ] Under https://github.com/xregistry/spec/blob/main/core/spec.md#common-attributes the following sentence doesn't quite make sense:
> For easy reference, the JSON serialization these attributes adheres to this form:

- [x] Under https://github.com/xregistry/spec/blob/main/core/spec.md#labels-attribute, there's a typo in `**HxRegistry**-labels-verified`
- [ ] Under https://github.com/xregistry/spec/blob/main/core/spec.md#xid-attribute, there’s a reference to gid, rid and vid here that is not yet defined, would it be worth adding a note there?
- [ ] Sections like https://github.com/xregistry/spec/blob/e4772baa8fe0e068a21fe34fc371744c4941001c/core/spec.md?plain=1#L2338 could benefit from linking as it's tricky to navigate the document
- [ ] I caught typos along the way, but I think it would be more useful to do a sweep towards the end, happy to do that.

### LINKED ISSUE #256: readonly attribute question
The Resource "readonly" attribute says:
```
  - A request to update this attribute, on a non-read-only Resource, MUST be
    silently ignored.
```
why not generate an error?

- - -

readonly Attribute

Type: Boolean
Description: indicates whether this Resource is updateable by clients. This attribute is a server controlled attribute and therefore can not be modified by clients. This specification makes no statement as to when Resource are to be read-only.
Constraints:
 - When not present, the default value is false.
 - REQUIRED when true, otherwise OPTIONAL.
 - If present, it MUST be a case sensitive true or false.
 - A request to update this attribute, on a non-read-only Resource, MUST be silently ignored.
 - A request to update a read-only Resource MUST generate an error unless the ?noreadonly query parameter was used, in which case the error MUST be silently ignored. See [Registry APIs](https://github.com/xregistry/spec/blob/main/core/spec.md#registry-apis) for more information.

### LINKED ISSUE #257: reduce the "model update" section to just "can only add optional extension attributes"

duglin:
Changes for week of Feb 10
    
- Primer: explain why we ignore unknown query params instead of generate an err
- Define URI, URIABSOLUTE, URIRELATIVE, URL, URLABSOLUTE, URLRELATIVE types
  - URI-reference is now gone
- Split "API View" and "Document View" discussions to provide clarity
  - Rather than "if statements"
- Removed "server's choice" for picking the "default version". Now it's either
  "newest = default" or "client's choice/sticky"
- Make "URL*" types based off of "URI" spec plus the "URL" constraints
  mentioned in the URI spec.
- misc typos and clean-up
- reduce scope of model updates that are allow - basically just non-breaking changes
- add more links for "see ..." to make navigation easier
- support PATCH on "capabilities"
- support support for "GET /capabilities?offered"
    
Fixed #248
Fixed #240
Fixed #251 
Fixed #256 
Fixed #257 
Fixed #223 
Fixed #221

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -510,3 +510,16 @@ values need to change.
 
 Obviously, situations like these should be rare but the specification needs
 to allow for them.
+
+# Why does an unknown query parameter not generate an error?
+
+People have identified cases where query parameters might be added to request
+URLs without the client (or developer) having control to prevent it. It might
+be client-side tooling or intermediaries (e.g. proxies) that might modify
+the URL - for example to add tracking or end-to-end tracing information.
+
+To avoid these client not being able interact with an xRegistry deployment,
+the authors chose to have the server ignore unknown query parameters. It is
+worth noting that the specification says they SHOULD be ignored, not that they
+MUST be ignored. So there is room for an implementation to be very picky if
+needed, but at the risk of potentially causing interoperability problems.
--- a/core/primer.md
+++ b/core/primer.md
@@ -518,7 +518,7 @@ URLs without the client (or developer) having control to prevent it. It might
 be client-side tooling or intermediaries (e.g. proxies) that might modify
 the URL - for example to add tracking or end-to-end tracing information.
 
-To avoid these client not being able interact with an xRegistry deployment,
+To avoid these clients not being able interact with an xRegistry deployment,
 the authors chose to have the server ignore unknown query parameters. It is
 worth noting that the specification says they SHOULD be ignored, not that they
 MUST be ignored. So there is room for an implementation to be very picky if
--- a/core/spec.md
+++ b/core/spec.md
@@ -321,9 +321,12 @@ information about each data type):
 - `TIMESTAMP`
 - `UINTEGER`
 - `URI`
-- `URIREFERENCE`
+- `URIABSOLUTE`
+- `URIRELATIVE`
 - `URITEMPLATE`
 - `URL`
+- `URLABSOLUTE`
+- `URLRELATIVE`
 - `XID`
 - `TYPE` - one of the allowable data type names (MUST be in lower case) listed
   in [Attributes and Extensions](#attributes-and-extensions)
@@ -436,40 +439,37 @@ be one of the following data types:
 - `string` - sequence of Unicode characters.
 - `timestamp` - an [RFC3339](https://tools.ietf.org/html/rfc3339) timestamp.
 - `uinteger` - unsigned integer.
-- `uri` - absolute URI as defined in [RFC 3986 Section
+- `uri` - a URI as defined in [RFC 3986](https://tools.ietf.org/html/rfc3986).
+   Note that it can be absolute or relative.
+- `uriabsolute` - absolute URI as defined in [RFC 3986 Section
   4.3](https://tools.ietf.org/html/rfc3986#section-4.3).
-- `urireference` - URI-reference as defined in [RFC 3986
-  Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1).
+- `urirelative` - relative URI as defined in [RFC 3986 Section
+  4.2](https://tools.ietf.org/html/rfc3986#section-4.2).
 - `uritemplate` - URI Template as defined in
   [RFC 6570 Section 3.2.1](https://tools.ietf.org/html/rfc6570#section-3.2.1).
-- `url` - URL as defined in
-  [RFC 1738](https://datatracker.ietf.org/doc/html/rfc1738).
+- `url` - an absolute URL (`urlabsolute`) or relative URL (`urlrelative`).
+- `urlabsolute` - an absolute URI as defined in [RFC 3986 Section
+  4.3](https://datatracker.ietf.org/doc/html/rfc3986#section-4.3) with the
+  added "URL" constraints mentioned in [RFC 3986 Section
+  1.1.3](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.3).
+- `urlrelative` - a relative URI as defined in [RFC 3986 Section
+  4.2](https://datatracker.ietf.org/doc/html/rfc3986#section-4.2) with the
+  added "URL" constraints mentioned in [RFC 3986 Section
+  1.1.3](https://datatracker.ietf.org/doc/html/rfc3986#section-1.1.3).
 
-All relative URIs and URLs that reference entities within the Registry MUST
-begin with `#/` followed by a [JSON Pointer] reference to the entity. With the
-exception of `xid` values, relative URIs/URLs that reference entities that are
-part of a document view `GET` response (`?doc` is enabled), MUST be relative
-to the path specified on the request. Relative `xid` values MUST be relative
-to the root of Registry regardless of the path specified on the request.
+The 6 variants of URI/URL are provided to allow for strict type adherence
+when needed. However, for attributes that are simply "pointers" that might
+in practice be any of those 6 types it is RECOMMENDED that `uri` be used.
 
-For clarity, if a Registry has a Schema Resource at
-`/schemagroups/g1/schemas/s1`, then this entity's `self` URL (when serialized
-in document view) would change based on the path specified on the `GET`
-request:
-
-| GET Path | `self` URL |
-| --- | --- |
-| `http://example.com/myreg` | `#/schemagroups/g1/schemas/s1` |
-| `http://example.com/myreg/schemagroups` | `#/g1/schemas/s1` |
-| `http://example.com/myreg/schemagroups/g1/ ` | `#/schemas/s1` |
-| `http://example.com/myreg/schemagroups/g1/schemas ` | `#/s1` |
-| `http://example.com/myreg/schemagroups/g1/schemas/s1` | `#/` |
+Attributes that are defined to be relative URIs or URLs MUST state what they
+are relative to and any constraints on their values, if any.
 
 The root path of a Registry service MAY be at the root of a host or have a
 `PATH` portion in its URL (e.g. `http://example.com/myregistry`).
 
 The "scalar" data types are: `boolean`, `decimal`, `integer`, `string`,
-`timestamp`, `uinteger`, `uri`, `urireference`, `uritemplate`, `url`, `xid`.
+`timestamp`, `uinteger`, `uri`, `uriabsolute`, `urirelative`, `uritemplate`,
+`url`, `urlabsolute`, `urlrelative`, `xid`.
 Note that `any` is not a "scalar" type as its runtime value could be a complex
 type such as `object`.
 
@@ -537,7 +537,7 @@ The following attributes are used by one or more entities defined by this
 specification. They are defined here once rather than repeating them
 throughout the specification.
 
-For easy reference, the JSON serialization these attributes adheres to this
+For easy reference, the JSON of serialization these attributes adheres to this
 form:
 - `"SINGULARid": "STRING"`
 - `"self": "URL"`
@@ -557,20 +557,17 @@ The definition of each attribute is defined below:
 
 - Type: String
 - Description: An immutable unique identifier of the Registry, Group, Resource
-  or Version. The actual name of this attribute will vary based on the entity it
-  identifies. For example, a `schema` Resource would use an attribute name of
-  `schemaid`. This attribute MUST be named `registryid` for the Registry
+  or Version. The actual name of this attribute will vary based on the entity
+  it identifies. For example, a `schema` Resource would use an attribute name
+  of `schemaid`. This attribute MUST be named `registryid` for the Registry
   itself, and MUST be named `versionid` for all Version entities.
+
 - Constraints:
   - MUST be a non-empty string consisting of [RFC3986 `unreserved`
     characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)
     (ALPHA / DIGIT / `-` / `.` / `_` / `~`) and `@`, MUST start with
     ALPHA, DIGIT or `_` and MUST be between 1 and 128 characters in length.
   - MUST be case insensitive unique within the scope of the entity's parent.
-    In the case of the `registryid` for the Registry itself, the uniqueness
-    scope will be based on where the Registry is used. For example, a publicly
-    accessible Registry might want to consider using a globally unique value,
-    while a private Registry does not need to be so widely unique.
   - This attribute MUST be treated as case sensitive for look-up purposes.
     This means that an HTTP request to an entity with the wrong case for its
     `SINGULARid` MUST be treated as "not found".
@@ -582,6 +579,7 @@ The definition of each attribute is defined below:
     `SINGULARid` if one exists, otherwise an HTTP `400 Bad Request` error MUST
     be generated.
   - MUST be immutable.
+
 - Examples:
   - `a183e0a9-abf8-4763-99bc-e6b7fcc9544b`
   - `myEntity`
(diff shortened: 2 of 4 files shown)


core/spec.md
```diff
@@ -537,7 +543,7 @@ The following attributes are used by one or more entities defined by this
 specification. They are defined here once rather than repeating them
 throughout the specification.
 
-For easy reference, the JSON serialization these attributes adheres to this
+For easy reference, the JSON of serialization these attributes adheres to this
```
lailabougria, line 546:
```suggestion
For easy reference, the JSON serialization of these attributes adheres to this
```

core/spec.md
```diff
@@ -595,38 +599,45 @@ the display name.
...
-  when `$details` was used in the request, or when the Resource (or Version)
-  is included in the serialization of a parent entity. If `?doc` mode
-  is enabled, then `$details` MUST appear when this URL is an absolute URL,
-  otherwise for relative URLs it MUST NOT appear.
-
-  If the Resource or Version has the `hasdocument` model aspect set to `false`,
-  then this URL MAY be appended with `$details` only if the request was
-  directed at the Resource or Version and included `$details`. Otherwise in
-  all other situations `$details` MUST NOT be included.
+- Description: A server generated unique URL referencing the current entity.
```
lailabougria, line 607:
```suggestion
- Description: A server-generated unique URL referencing the current entity.
```

core/spec.md
```diff
@@ -595,38 +599,45 @@ the display name.
...
+  - Each entity in the Registry MUST have a unique `self` URL value that
+    locates where in the Registry hierarchy the entity exists and from where
```
lailabougria, line 609:
```suggestion
    locates the entity in the Registry hierarchy and from where
```
I think this reads a bit easier, feel free to ignore if you disagree

core/spec.md
```diff
@@ -595,38 +599,45 @@ the display name.
...
+    the entity can be retrieved.
+  - When specified as an absolute URL, It MUST be based on the URL of the
```
lailabougria, line 611:
```suggestion
  - When specified as an absolute URL, it MUST be based on the URL of the
```

core/spec.md
```diff
@@ -595,38 +599,45 @@ the display name.
...
+    Registry root appended with the hierarchy path of the Registry
+    entities/collections leading to the entity.
+
+    In the case of pointing to an entity that has `SINGULARid` attribute,
```
lailabougria, line 615:
```suggestion
    In the case of pointing to an entity that has a `SINGULARid` attribute,
```

core/spec.md
```diff
@@ -595,38 +599,45 @@ the display name.
...
+    the URL MUST be a combination of the URL used to retrieve its parent
+    appended with its `SINGULARid` value.
+
+- API View Constraints:
+  - MUST be a non-empty absolute URL based on the URL of the Registry.
+  - When serializing Resources or Versions, if the `hasdocument` aspect set to
```
lailabougria, line 621:
```suggestion
  - When serializing Resources or Versions, if the `hasdocument` aspect is set to
```

core/spec.md
```diff
@@ -1477,52 +1502,58 @@ The serialization of the Registry entity adheres to this form:
...
 - Type: Registry Model
 - Description: A description of the features, extension attributes, Groups and
   Resources supported by this Registry. See [Registry Model](#registry-model)
+
 - Constraints:
-  - OPTIONAL.
-  - MUST NOT be included in responses unless requested.
-  - MUST be included in responses if requested.
-  - SHOULD be included in document view when the model is not known in advance.
+  - MUST NOT be included in API and document views  unless requested.
```
lailabougria, line 1546:
```suggestion
  - MUST NOT be included in API and document views unless requested.
```

core/spec.md
```diff
@@ -1477,52 +1502,58 @@ The serialization of the Registry entity adheres to this form:
...
+  - MUST be included in API and document views  if requested.
```
lailabougria, line 1547:
```suggestion
  - MUST be included in API and document views if requested.
```

core/spec.md
```diff
@@ -2031,45 +2070,31 @@ new attributes within the model definitions themselves), but only if
...
-
-If a backwards incompatible change is needed, and the existing entities need
-to be preserved, then it is RECOMMENDED that users define a new Group or
-Resource types for future instances of those entities.
+Once a Registry has been created, changes to the model are limited to:
+- Adding new Groups or Resources.
+- Adding new OPTIONAL attributes.
+
+Any model change that might result in existing data within the Registry
+becoming non-compliant MUST generate an error, even if it does not do so. For
```
lailabougria, line 2078: What is the intent of the following?

>, even if it does not do so.

core/spec.md
```diff
@@ -2031,45 +2070,31 @@ new attributes within the model definitions themselves), but only if
...
+- Deleting an attribute, Group or Resource.
+- Changing the `name` or `type` of an attribute.
+- Removing a value from the enumeration list of an attribute.
+- Enabling/disabling a feature that might impact storing data
+  (e.g. `hasdocument`).
+
+Some example changes that are allowed:
+- Changing the `description` of an attribute.
+- Adding an enumeration value to an attribute.
+- Changing the `readonly` aspect or `default` value of an attribute.
```
lailabougria, line 2089: Wouldn't changing the `readonly` aspect from false to true possibly break something?

core/spec.md
```diff
@@ -3697,8 +3728,8 @@ Note that the `meta` and `versions` attributes MUST only appear when
 requested by the client - for example, via the `?inline` flag.
 
 When the Resource is serialized with its domain specific document in the
```
lailabougria, line 3730:
```suggestion
When the Resource is serialized with its domain-specific document in the
```

core/primer.md
```diff
@@ -510,3 +510,16 @@ values need to change.
...
 to allow for them.
+
+# Why does an unknown query parameter not generate an error?
+
+People have identified cases where query parameters might be added to request
+URLs without the client (or developer) having control to prevent it. It might
+be client-side tooling or intermediaries (e.g. proxies) that might modify
+the URL - for example to add tracking or end-to-end tracing information.
+
+To avoid these client not being able interact with an xRegistry deployment,
```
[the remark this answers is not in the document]
duglin: Done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: Done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: even if the data doesn't become invalidated because they got lucky "this time", doesn't mean it's an ok change to make. I'm not happy with that phrasing I used, so I'm open to an alternative.

core/spec.md
duglin: hmm would it? I agree it would change what future API calls might do to the data, but would any existing data become invalid?

core/spec.md
duglin: done and got all of them in the spec

core/spec.md
duglin: e.g. changing an optional attr from string to int would be safe if that attr isn't used yet. But it's not a valid change to make in general.

core/spec.md
```diff
@@ -850,17 +883,16 @@ of the existing entity. Then the existing entity can be deleted.
...
     - If the request value is different than the existing value, then the
       request value MUST be used as its new value.
-    - If the request value is `null` then the current date/time MUST be used
-      as the new value.
   - When absent in a write operation request, it MUST be set to the current
     date/time.
-  - In cases where the `modifiedat` attribute is set to the current date/time
-    on multiple entities within the same operation, the same value MUST be
-    applied to all of the entities.
+  - During the processing of a single request, all entities that have their
```
jskeet, line 892: Same question as earlier about how this could be enforced.

core/spec.md
```diff
@@ -906,14 +927,14 @@ incorporate authentication and/or authorization mechanisms for the APIs.
 If an OPTIONAL HTTP path is not supported by an implementation, then any
 use of that API MUST generate a `404 Not Found` error.
 
-If a HTTP method is not supported for a defined HTTP path then a
+If a HTTP method is not supported for a supported HTTP path then a
```
jskeet, line 930: a => an?

core/spec.md
```diff
@@ -1998,7 +2043,73 @@ response.
...
+GET /capabilities?offered
+
+{
+  "STRING": [ ALLOWED-VALUES ] *
+}
+```
+
+Where:
+- `"STRING"` is the capability name.
+- `ALLOW_VALUES` is an array of the valid values for that capability. The
```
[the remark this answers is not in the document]
jskeet: (Would "SUPPORTED_VALUES" be clearer though?)

core/spec.md
```diff
@@ -1998,7 +2043,73 @@ response.
...
+  "pagination": [ true, false ],
+  "schemas": [ "xRegistry-json/0.5" ],
+  "shortself": [ true, false ],
+  "specversions": [ "0.5" ]
+}
+```
+
+The array of values allows for some special cases:
+- `"*"` mean that all possible values of that data type are allowed.
+- Numeric capabilities MAY express ranges via `"#-#"` (min-max),
```
jskeet, line 2084: I suspect that using `[x, y]`, `[x, y)`, `(x, y]` or `(x, y)` would be more flexible here, to allow for closed and open bounds.

core/spec.md
```diff
@@ -2010,18 +2121,20 @@ the Registry in advance and therefore will need to dynamically discover it.
 To enable support for a wide range of use cases, but to also ensure
 interoperability across implementations, the following rules have been defined
 with respect to how models are defined or updated:
+- Changes to specification defined attributes MAY be included in the model but
```
jskeet, line 2124:
```suggestion
- Changes to specification-defined attributes MAY be included in the model but
```

core/spec.md
```diff
@@ -2031,45 +2144,31 @@ new attributes within the model definitions themselves), but only if
...
-Resource from `false` to `true` if there are attributes defined on that
-Resource whose names conflict with the `RESOURCE*` attributes. Any attempt do
-to so MUST generate an error.
-
-If a backwards incompatible change is needed, and the existing entities need
-to be preserved, then it is RECOMMENDED that users define a new Group or
-Resource types for future instances of those entities.
+Once a Registry has been created, changes to the model are limited to:
+- Adding new Groups or Resources.
+- Adding new OPTIONAL attributes.
```
jskeet, line 2149: What about making previously-required attributes optional? That wouldn't break any existing data - but *might* break clients which expect the previously-required attributes to always be present.

core/spec.md
```diff
@@ -3846,7 +3965,7 @@ and the following Resource level attributes:
   - When not present, the default value is `false`.
   - REQUIRED when `true`, otherwise OPTIONAL.
   - If present, it MUST be a case sensitive `true` or `false`.
-  - If present in a request, a value of `null` has the same meaning as
+  - If present in a request, a value of `null` MUST have the same meaning as
```
jskeet, line 3968: Do we have to support null as a valid value at all? Do we have the same null==false equivalence for other Boolean attributes?

core/spec.md
```diff
@@ -5252,33 +5415,26 @@ as the semantic minor version identifier.
...
+1. Newest = Default. The newest Version created (based on `createdat` timestamp)
    is always the "default" Version. This is the default choice. If more than
    one Version has the same "newest" `createdat` timestamp, then the choice
    the Version with the highest alphabetical `versionid` value using case
    insensitive compares MUST be chosen. For example, if there are 3 Versions
    with `versionid` values of `v10`, `z1` and `V2`, then ordering of the
    Versions from oldest to newest would be: `v10`, `V2`, `z1.
 
-3. Client explicitly chooses the "default". In this option, a client has
+2. Client explicitly chooses the "default". In this option, a client has
```
jskeet, line 5429: Suggest using "1" for all numbers in numbered lists, and Markdown will just do the right thing. (I believe, anyway. You *might* need to remove the blank line above.)

core/spec.md
```diff
@@ -5252,33 +5415,26 @@ as the semantic minor version identifier.
...
    explicitly chosen which Version is the "default" and it will not change
    until a client chooses another Version, or that Version is deleted (in
-   which case the server MUST revert back to option 2 (newest = default)).
-   This is referred to as the default Version being "sticky" as it will not
-   change until explicitly requested by a client.
+   which case the server MUST revert back to option 2 (newest = default), if
```
jskeet, line 5432:
```suggestion
   which case the server MUST revert back to option 1 (newest = default), if
```

core/spec.md
duglin: [x,] for x or higher

or ".." instead of ","

core/spec.md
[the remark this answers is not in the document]
duglin: make it "OFFERED_VALUES"

check on "_" vs "-"

core/spec.md
```diff
@@ -1947,30 +1984,37 @@ extension values.
...
+missing capability MUST be interpreted as a request to reset it to its default
+value. If a `PATCH` is used then each capability included MUST be fully
+specified, and fully replaced by the incoming value. In other words, `PATCH`
+is done at a capability level not any deeper within the JSON structure.
 
 The request to the `/capabilities` API MUST be of the form:
 
 ```yaml
 PUT /capabilities
+PATCH /capabilities
```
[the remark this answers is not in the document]
duglin: just split into 2 examples

core/spec.md
```diff
@@ -817,29 +841,38 @@ of the existing entity. Then the existing entity can be deleted.
...
     existing value, however a value of `null` MUST use the current date/time
     as the new value.
   - When absent in a write operation request, any existing value MUST remain
-    unchanged, or if not present, set to the current date/time
-  - In cases where the `createdat` attribute is set to the current date/time
-    on multiple entities within the same operation, the same value MUST be
-    applied to all of the entities.
+    unchanged, or if not already set, set to the current date/time
+  - During the processing of a single request, all entities that have their
+    `createdat` or `modifiedat` attributes set to the current date/time
```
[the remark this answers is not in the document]
duglin: reword this to say "null" instead of "current date/time"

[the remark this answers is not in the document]
duglin: ok done. See if this is any better:

  - During the processing of a single request, all entities that have their
    `createdat` or `modifiedat` attributes set to the current date/time due
    to the processing rules above MUST use the same value in all cases.

Can edit more in the next PR if needed.

core/spec.md
duglin: same change as above

jskeet:
Yup, that change for the current date/time sounds good.

core/spec.md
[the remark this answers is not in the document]
duglin: done

core/spec.md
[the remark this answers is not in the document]
duglin: done

core/spec.md
duglin: Will address in https://github.com/xregistry/spec/issues/262

core/spec.md
duglin: will address via: https://github.com/xregistry/spec/issues/261#issuecomment-2657673120

core/spec.md
```diff
@@ -3965,10 +4105,15 @@ Resource's xRegistry metadata. To support this, some of the Resource's
...
-MUST use `Content-Type`, not `xRegistry-contenttype`).
+Resource SHOULD appear as HTTP headers where the header name is the name of the
+attribute prefixed with `xRegistry-`. Note, the optionality of this requirement
+is not to allow for servers to decide whether or to do so, rather it is to
+allow for [No-Code Servers](#no-code-servers) servers than might not be able
+to control the HTTP response headers.
+
+Certain attributes do not follow this rule if a standard HTTP header name is
+to be used instead (e.g. `contenttype` MUST use `Content-Type`, not
+`xRegistry-contenttype`).
```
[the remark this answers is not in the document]
duglin: done

duglin:
approved on the 2/13 call - will address some of @jskeet's comments in follow-on PRs

core/spec.md
lailabougria: How about:
> , even if the existing data doesn't result in non-compliance.

PR #263, opened 2025-02-14
Changes for week of Feb17

duglin:
- minor text/tool tweaks
- fix-up OWNERS and contributors pages
- add info about running our tests locally

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -119,3 +119,16 @@ Documents in this repository will adhere to the following rules:
 - Lines are wrapped at 80 columns (when possible)
 - Specifications will use [RFC2119](https://tools.ietf.org/html/rfc2119)
   keywords to indicate normative requirements
+
+## Testing Changes Locally
+
+Each PR will run through some checks to verify everything looks ok. Checks
+include:
+- all links resolve successfully
+- no spelling errors are found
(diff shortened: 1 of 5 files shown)


Fannon (approved):

docs/CONTRIBUTING.md
```diff
@@ -119,3 +119,16 @@ Documents in this repository will adhere to the following rules:
...
 - Specifications will use [RFC2119](https://tools.ietf.org/html/rfc2119)
   keywords to indicate normative requirements
+
+## Testing Changes Locally
+
+Each PR will run through some checks to verify everything looks ok. Checks
+include:
+- all links resolve successfully
+- no spelling errors are found
+- no tabs in *md files
```
jskeet, line 129: I get nervous with `*` in a Markdown file - in this case it *doesn't* mean italics because there's no matching one, but if we ended up with "\*md files or \*xyz files" then I suspect it would be broken.

How about just "no tabs in Markdown files"?

jskeet (approved):
One tiny suggestion, but generally fine.

docs/CONTRIBUTING.md
duglin: done

duglin:
Approved on the 2/20 call

PR #266, opened 2025-02-25
Compatibility

lailabougria:
Fixes #

## Why keep compatibility in v1 of the spec

Compatibility levels up the value of a registry. Apart from it being a resource I can use to discover and query the current schemas, it becomes a resource that can enforce a compatibility strategy. Looking at the existing schema registries out there, most (if not all) of them support compatibility enforcement (albeit not for all schemas):
- [Confluent Schema Registry](https://docs.confluent.io/platform/current/schema-registry/fundamentals/schema-evolution.html#schema-evolution)
- [Event Hubs Schema Registry](https://learn.microsoft.com/en-us/azure/event-hubs/schema-registry-concepts#schema-evolution)
- [Apicurio](https://www.apicur.io/registry/docs/apicurio-registry/3.0.x/getting-started/assembly-rule-reference.html)
- [AWS Glue](https://docs.aws.amazon.com/glue/latest/dg/schema-registry.html#schema-registry-compatibility)

### `enforcecompatibility`

The registry-wide `enforcecompatibility` setting can't be changed once set to `true`.

#### Reasoning

- If it was previously set to false, previous versions aren't reevaluated, only future versions are. This makes it behave much more like a setting or flag, not causing any validation to happen when modified. Instead, compatbility will be enforced when changes are made.
- Changing from `true` back to `false` is not allowed as it's a fundamental change in how the registry behaves.

### Resource-level `compatibility`

The resource-level `compatibility` setting can be changed from `none` to another value, but can't be changed to any other value. The new value will only be vallidated for future changes.

#### Reasoning

It's unlikely that a resource that adhered to one compatibility strategy can adhere to another compatibility strategy without changing previous versions, unless it was always fully compatible. In addition, a change in the compatibility strategy is very much like a contract change which should be clearly visible, as users of the registry may be making assumptions based on the previously defined strategy.

### `setdefaultversionsticky`

The `setdefaultversionsticky` attribute can't be set to `true` when the registry has `enforcecompatibility` set to `true`. This is to avoid having to keep track of the order of versions as they were set according to this attribute. 
Any request to set the `setdefaultversionsticky` to `true` when `enforcecompatibility` is turned on should result in a failure.

#### Reasoning

This allows us to defer the inclusion of version-ordering as indicated by the sticky flag in the spec, and instead allows us to rely on the version's creation date. This may however implicate that we need a server creation date time which can't be set or altered by the client.

**Release Note**

```release-note

```

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -3910,6 +3910,9 @@ and the following Resource level attributes:
   - If present, it MUST be a case-sensitive value from the model defined
     enumeration range.
   - When not present, the implied default value is `none`.
+  - The value may change from `none` to any other defined value. However, once
+    set to a non-`none` value, it MUST NOT be changed. Instead, a new
+    resource should be defined with the new compatibility value.
 
 ##### `defaultversionid` Attribute
 - Type: String
--- a/core/spec.md
+++ b/core/spec.md
@@ -3978,6 +3978,8 @@ and the following Resource level attributes:
 - Constraints:
   - When not present, the default value is `false`.
   - REQUIRED when `true`, otherwise OPTIONAL.
+  - If `enforcecompatibility` is set to `true`, this attribute MUST always
+    be set to `false`.
   - If present, it MUST be a case-sensitive `true` or `false`.
   - If present in a request, a value of `null` MUST have the same meaning as
     deleting the attribute, implicitly setting it to `false`.
--- a/core/spec.md
+++ b/core/spec.md
@@ -3860,7 +3860,7 @@ and the following Resource level attributes:
   - MUST be a read-only attribute.
   - When not present, the default value is `false`.
   - REQUIRED when `true`, otherwise OPTIONAL.
-  - If present, it MUST be a case sensitive `true` or `false`.
+  - If present, it MUST be a case-sensitive `true` or `false`.
   - A request to update a read-only Resource MUST generate an error unless
     the `?noreadonly` query parameter was used, in which case the error MUST
     be silently ignored. See [Registry APIs](#registry-apis) for more
@@ -3907,7 +3907,7 @@ and the following Resource level attributes:
   - `none` - No compatibility checking is performed.
 
 - Constraints:
-  - If present, it MUST be a case sensitive value from the model defined
+  - If present, it MUST be a case-sensitive value from the model defined
     enumeration range.
   - When not present, the implied default value is `none`.
 
@@ -3975,7 +3975,7 @@ and the following Resource level attributes:
 - Constraints:
   - When not present, the default value is `false`.
   - REQUIRED when `true`, otherwise OPTIONAL.
-  - If present, it MUST be a case sensitive `true` or `false`.
+  - If present, it MUST be a case-sensitive `true` or `false`.
   - If present in a request, a value of `null` MUST have the same meaning as
     deleting the attribute, implicitly setting it to `false`.
   - The processing of the `defaultversionsticky` and `defaultversionid`
--- a/core/spec.md
+++ b/core/spec.md
@@ -1900,16 +1900,18 @@ The following defines the specification-defined capabilities:
   reject all attempts to create/update a Resource (or its Versions) that would
   result in those entities violating the stated compatibility rules.
 
-  This includes the server rejecting requests to update the `compatibility`
-  attribute's value if any of the Resource's Versions would violate the
-  compatibility rules.
-
   A value of `false` indicates that the server MUST NOT perform any
   compatibility checking.
 
-  Attempts to change this value from `false` to `true` MUST fail if doing so
-  would result in any existing Version violating the `compatibility` rules
-  defined for the owning Resource.
+  Attempts to change this value from `false` to `true` SHOULD NOT result in
+  any validation of existing versions. Instead, compatibility will be
+  enforced for any new versions that are added.
+
+  This value can be set to `true` at any time, but can't be reset to`false`.
+  Once compatibility is enforced, it must remain enforced. Changing this
+  requires a new registry to be created as it's a fundamental change for
+  clients of the registry, which should be signaled accordingly.
+
 - If not specified, the default value is `false`.
 
 #### `flags`


duglin:
A few things popped in my head as I was reading this:
- while the sticky flag is problematic, the ability for clients to set the versionid is as well because people can create then in non-linear order (eg v1, v1.2, v1.1), and this might be ok for usecases where people never use the "default" version. It seems like we'd need to remove the ability to use both features.
- having a global flag that bans the use of other features will probably be too restrictive because we should probably allow two different GroupTypes to have different semantics w.r.t. things like compatibility and versioning. So, we may want to consider moving the global flag down to the (at least) the Group level. This is especially true when you consider cases where a single registry has schemas, messages and endpoints - where the semantic rules for things like stickiness and versioning could vary a lot.
- changing "compatibility" from "non-none" to "none" (or absent) should be ok, no?
- I agree that changing it from non-none to another non-none is probably something we should ban
- going from "none"/absent to "non-none" should probably be ok as long as there's not more than one version. I think we need this for cases where people create a Resource w/o knowing for sure what "compat" mode they want yet, but they better know before they create the 2nd version. This allows a dev to not be blocked to setup things, while waiting for someone higher-up to decide what kind of policy they'll want to turn on.

core/spec.md
```diff
@@ -1951,9 +1951,15 @@ The following defines the specification-defined capabilities:
...
   compatibility checking.
 
-  Attempts to change this value from `false` to `true` MUST fail if doing so
-  would result in any existing Version violating the `compatibility` rules
-  defined for the owning Resource.
+  Attempts to change this value from `false` to `true` SHOULD NOT result in
+  any validation of existing versions. Instead, compatibility will be
+  enforced for any new versions that are added.
+
+  This value can be set to `true` at any time, but can't be reset to`false`.
```
duglin, line 1958: s/can't/MUST NOT/

core/spec.md
```diff
@@ -1951,9 +1951,15 @@ The following defines the specification-defined capabilities:
...
+  Once compatibility is enforced, it MUST remain enforced. Changing this
+  requires a new registry to be created as it's a fundamental change for
+  clients of the registry, which MUST be signaled accordingly.
```
duglin, line 1961: If we talked about this and I'm just forgetting, sorry....
why did you want this restriction? If changing this flag only impacts future versions, then what's the harm in turning it off? An impl can choose to ban this change if they want, and how they notify users, if it does change, is up to them... but from a spec perspective I'm not sure about banning it for all usecases.

lailabougria: @duglin This is an angle that I don't think we fully agreed upon, so I left it in here purposefully to have the discussion. I can see it both ways:
- On the one hand, it's a very big restriction as you mentioned
- On the other hand, if compatibility strategies are set and the registry promises to enforce those strategies and later "changes its mind", this is a breaking change. Both producers and consumers are operating based on the assumption that compatibility is enforced which may cause ripple effects in the system. I'm leaning towards the stricter end but that may be due to my messaging-hat-bias. If the enforcement can be turned off at any time it loses much of its value imho. The way I see it is that this would likely mean that users will be much more careful in turning it on?

core/primer.md
```diff
@@ -674,3 +674,18 @@ in the entity in order to be model compliant.
 
 Likewise, implementations of the server must validate the entire entity
 against the new model, not just a subset of the entity's attributes.
+
+# Why are unset `ancestor` attributes allowed?
+
+The [`ancestor` attribute](spec.md#ancestor-attribute) is used to build a
+hierarchy of Versions aiding in compatibility checking when the
+[`compatibility`-attribute](spec.md#compatibility-attribute) is set.
```
duglin, line 682: let's remove the dash here and below

core/spec.md
```diff
@@ -5528,6 +5549,30 @@ as defined below:
...
 
+##### `ancestor` Attribute
+- Type: String
+- Description: A string value that points to another version based on the
+  `versionid`-attribute.
+
+  When creating a version without explicitly setting the `ancestor`
+  attribute, the server MUST set the `ancestor` to the `versionid` of the
+  current default Version of the Resource. The `ancestor` attribute MAY also
+  be explicitly set to NULL, to indicate to the server that this Version has
```
duglin, line 5560: s/NULL/\`null\`/

duglin: Also, we need to think more about this because today when someone sets an attribute "null" we never tell the impl to save it as "null". Meaning, a GET of that attribute later will most likely be absent in the response. If we did that for "ancestor" then on an import/put we can't tell the diff between "I'm a root" and "I come after the current default version".  We may need to require this attribute to ALWAYS be serialized to remove the ambiguity. Or, I guess, define a special versionid to mean "root" - but I'm not thrilled with that (despite that we do reserve some other version IDs - like "null" - for other edge cases).

I think the simple/default case should be "missing ancestor == I come after current default".

core/spec.md
```diff
@@ -5528,6 +5549,30 @@ as defined below:
...
+  based the `versionid` in alphabetical order, where the first version
+  created has the `ancestor`-attribute pointed to the Resource's current
+  default Version.
+
+- Constraints: the `ancestor` attribute MUST NOT be set to a value that
+  creates circular references between versions. For example, an operation that
+  makes Version A's ancestor B, and version B's ancestor A, MUST be refused
+  by the server.
+  If the `ancestor` attribute is set to a `versionid` that does not exist
+  the request SHOULD be refused by the server.
```
duglin, line 5574: I think we need to remove this otherwise we're basically saying "a server SHOULD reject an import of a valid existing registry if it has dangling pointers" - which is just weird. OR, we remove the idea of dangling pointers and switch this from a SHOULD to a MUST.

I prefer the latter because while a dangling pointer has some interesting info, in the end it doesn't really help the user because they can't find that missing version; the server can't do anything with that info; and what if someone creates a new version with that versionid? Is it safe to have this dangling pointer no longer be dangling? It could be what they wanted, but we don't know. I'd prefer that they explicitly set up the pointer if they're trying to recreate the missing part of the tree.  Especially, since after an import into a new Registry the server might start at "1" for new versionids and could accidentally create a version that matches a dangling pointer.

core/spec.md
```diff
@@ -5368,7 +5386,8 @@ When serialized as a JSON object, the Version entity adheres to this form:
 
   "RESOURCEurl": "URL", ?                  # If not local
   "RESOURCE": ... Resource document ..., ? # If inlined & JSON
-  "RESOURCEbase64": "STRING" ?             # If inlined & ~JSON
+  "RESOURCEbase64": "STRING", ?             # If inlined & ~JSON
```
duglin, line 5389: extra space before "#"

core/spec.md
```diff
@@ -5528,6 +5549,30 @@ as defined below:
   - MUST NOT be present if the Resource's `hasdocument` model attribute is
     set to `false.
 
+##### `ancestor` Attribute
+- Type: String
+- Description: A string value that points to another version based on the
+  `versionid`-attribute.
+
+  When creating a version without explicitly setting the `ancestor`
+  attribute, the server MUST set the `ancestor` to the `versionid` of the
```
lailabougria, line 5558: The risk here is that if later the user changes the value of the `versionid` attribute to a new value, we'll have to enforce implementations to update the `ancestor` attribute of all versions that point to the changed `versionid` to ensure accidentally creating dangling versions that don't point to a valid version anymore.

core/spec.md
lailabougria: My intent here was to remove the `enforcecompatibility` flag, as we agreed upon in previous meetings. This would mean that if you set the `compatibility` attribute of a Resource to a value, this is by default enforced (assuming the registry is able to enforce it). If you don't want the enforcement, this attribute should be set to `None`. 

However, thinking through this, I did have the following concern:
What if the registry, for whatever reason, cannot enforce compatibility, but still, you want to centrally register what the agreement was (even though it can't be enforced)? Now, you could only really do that by defining an extension that allows you to define that. Are we okay with that?

duglin: It seems to me that we may really confuse people with fields that "sometimes" are verified/used by the server and "sometimes" not. And taken to this logical conclusion... if a server doesn't support any compat checking then the "compat" field should not be available at all. Likewise, setting it to a valid that the server can't support (even if it does support others) should not be a valid state. 

If we don't take a firm position on this then how does a client know when the server is enforcing something or not (ie. lying or not)?

If we don't do this then we wouldn't just need a "enforcecompat" flag for the client to enable/disable, but also some kind of "yes, I did check it" flag for the server to let clients know if it did it.

core/spec.md
duglin: yes.

I've been trying to find a nice solution for how to know when "null" (on a request) means "I come after the current default" vs "I'm a new root" and I only see two solutions as of now:
1 - require versions that are root to ALWAYS include `"ancestor": null` instead of it being absent.
2 - define a new versionid (e.g. `none`) to mean "I'm a root". We'd probably need to change absent/null to "none" in some cases for consistency (e..g new resource with 1 version). I think an absent/null attribute may need to be banned, and the server would need to fill it in appropriately before committing the tx.

First option feels cleaner, but it's annoying that we can't follow our normal rule that "null" and "absent" mean the same thing.
Second option makes it very clear what's going on, but it's more verbose in some cases. 

So, my answer is still "yes" but instead of "null" it would be "none" if we did option 2.

duglin:
Scenario:
- create a resource+version(v1), set sticky=true
- create a new version(vFOO) - server assigns it `ancestor:v1`. Note that default version does not change, it's sticky.
- create a new version - what is its ancestor? "v1" or "vFOO"? If "v1", we forked

Scenario:
- create a resource+version(v1), set sticky=true
- create a new version(vFOO) - server assigns it `ancestor:v1`. Note that default version does not change, it's sticky.
- create a new version(vBAR) where request says `ancestor:v1`, we forked on purpose
- create a new version - what is its ancestor? "v1" (fork) or "vFOO" or "vBAR" ?

The idea of the server being smart and auto-setting a new version's "ancestor" AND moving the "default version" pointer may only be valid when sticky is "false". Which may surprise some people if they think of "sticky" as something that just influences what people see on a GET, not what happens on a PUT/POST.

When we think about why git doesn't have this problem, I think it's because git clients are always working within the context of a commit that's based on another commit that the user explicitly picked. I think the equivalent of this would be to require clients to provide "ancestor" when sticky is "true" if they don't want the new version's ancestor to be the current default version.

lailabougria:
- If sticky = OFF, when you create a new version, the ancestor is set to the default version which is **the latest version that was created**
- If sticky = ON, when you create a new version, the ancestor is set to **the latest version that was created**

duglin:
My notes:
- creating a version w/o "ancestor" -> "ancestor" defaults to latest version (based on newest `createdat` and then highest alpha)
- creating a version w/ "ancestor":null -> becomes a root (ancestor is missing from GET responses)
- creating a version w/ "ancestor" != null -> ok, allows for dangling references   **?????**
- creating multiple versions:
  - create all versions
  - iterate over all new versions, oldest per createdat first, then lowest alpha first
    - absent "ancestor" (or `&orphan`) -> set to current latest version, this becomes the latest
    - "ancestor": null -> becomes a root
    - "ancestor" != null -> ok, skip
- sticky has no influence over the above logic
- maxversions delete -> find all roots, delete oldest / lowest alpha  **dangling???**
- delete middle version -> 2 trees   **dangling child???**

lailabougria:
Regarding `enforcecompatibility`:
- Even if compatibility is set and enforced, not all servers will be able to enforce compatibility
- Instead of `enforcecompatibility`, we can use "compatibility":"backward;nocheck" which is setting both flags at once, allowing for scenarios in which you want to communicate the compat strategy without enforcement.

core/primer.md
```diff
@@ -674,3 +674,18 @@ in the entity in order to be model compliant.
...
+# Why are `ancestor` attributes that point to self allowed?
+
+The [`ancestor` attribute](spec.md#ancestor-attribute) is used to build a
+hierarchy of Versions aiding in compatibility checking when the
+[`compatibility` attribute](spec.md#compatibility-attribute) is set.
+Generally speaking, implementations are not expected to allow users to
+explicitly create multiple roots for a single Version. However, the
+specification does not mandate this as there may be cases where it is useful
+to allow for multiple roots. In some cases it may even be unavoidable, for
+example when the `maxversions`-attribute (See `groups.resources.maxversions`
```
duglin, line 687: let's remove the dash

core/spec.md
```diff
@@ -2338,6 +2345,7 @@ Regardless of how the model is retrieved, the overall format is as follows:
           "setversionid": BOOLEAN, ?   # vid settable? Default=true
           "setdefaultversionsticky": BOOLEAN, ? # sticky settable? Default=true
           "hasdocument": BOOLEAN, ?     # Has separate document. Default=true
+          "singleversionroot": BOOLEAN, ? # allow single root. Default=false
```
duglin, line 2348: s/allow/enforce/

core/spec.md
```diff
@@ -2677,11 +2685,16 @@ The following describes the attributes of Registry model:
...
   - When the limit is exceeded, implementations MUST prune Versions by
-    deleting the oldest Version (based on creation times) first, skipping the
-    Version marked as "default". An exception to this pruning rule is if
-    `maxversions` value is one (`1`) then the newest Version of the Resource
-    MUST always be the "default" and the `setdefaultversionsticky` aspect
-    MUST be `false`.
+    deleting the oldest Version first, skipping the
+    Version marked as "default". The strategy for finding the oldest Version
+    to delete is as follows:
+    1. Find all Versions that do not have an existing `ancestor`
```
duglin, line 2691: I'm not sure ancestor will ever be null given what we talked about today, right? They would always point to themselves, I think.
So this could be either:
- find all versions that are roots
or
- find all versions that have an ancestor value references itself

core/spec.md
```diff
@@ -2677,11 +2685,16 @@ The following describes the attributes of Registry model:
...
+    1. If many exist, find all the Version with the oldest `created` date
```
duglin, line 2692: s/Version/Versions/

core/spec.md
```diff
@@ -2677,11 +2685,16 @@ The following describes the attributes of Registry model:
...
+    1. If many exist, find the first Version according to alphabetical order
+       of the `versionid`-attribute and delete it
```
duglin, line 2694: s/-/ /
I'd stop after "attribute".
Then on the next line, outside the numbered list say: Once the single oldest Version is determined, delete it.

core/spec.md
```diff
@@ -2677,11 +2685,16 @@ The following describes the attributes of Registry model:
     it is valid for an implementation to only support one (`1`) Version when
     `maxversions` is set to `0`.
   - When the limit is exceeded, implementations MUST prune Versions by
-    deleting the oldest Version (based on creation times) first, skipping the
-    Version marked as "default". An exception to this pruning rule is if
-    `maxversions` value is one (`1`) then the newest Version of the Resource
-    MUST always be the "default" and the `setdefaultversionsticky` aspect
-    MUST be `false`.
+    deleting the oldest Version first, skipping the
+    Version marked as "default". The strategy for finding the oldest Version
```
duglin, line 2689: I think it might be good to add a section to the primer to talk about how "oldest" doesn't necessarily mean "the one with the oldest createdat timestamp". Normally, that's what we'd expect the main algorithm to focus on, but  "ancestors" take higher priority because they more accurately describe an ordered lineage than timestamps do.

core/spec.md
```diff
@@ -2728,6 +2741,18 @@ The following describes the attributes of Registry model:
...
+- `groups.resources.singleversionroot`
+    - Type: Boolean (`true` or `false`, case-sensitive).
+    - OPTIONAL.
+    - Indicates whether Resources of this type can have multiple Versions
+      that have their `ancestor` attribute set to itself.
+    - When not specified, the default value MUST be `false`.
+    - A value of `true` indicates that only one Version of the Resource can
+      have its `ancestor` attribute set to itself. This is useful to avoid
+      creating multiple roots. When this attribute is set to `true`, the
+      server MUST reject the deletion of any Versions that would cause multiple
```
duglin, line 2753: server MUST generate an error if any request results in a
state where more than one Version of a Resource is a
root of an ancestor tree.
- - -
This will cover ALL cases of multiple roots being created, not just deletes.  E.g. an import that tries to create a 2nd one.

core/spec.md
```diff
@@ -2728,6 +2741,18 @@ The following describes the attributes of Registry model:
   - A value of `true` indicates that Resource of this type supports a separate
     document to be associated with it.
 
+- `groups.resources.singleversionroot`
+    - Type: Boolean (`true` or `false`, case-sensitive).
+    - OPTIONAL.
+    - Indicates whether Resources of this type can have multiple Versions
+      that have their `ancestor` attribute set to itself.
```
duglin, line 2748: ...can have multiple Versions
that represent roots of an ancestor tree, as indicated by
the Version's `ancestor` attribute value being the same as
its `versionid` attribute.
- - -
I think it's better to refer to this as "having multiple roots" instead of "things pointing to themselves", because 1) it'll align with the name of attribute and 2) I think the mental model of what "pointing to itself" means is an important think to convey... meaning, the key point is "multiple roots", and the self-pointer is just how it's expressed.  IOW we want people think "multi-roots is bad", not "more than one self-pointer is bad".

core/spec.md
```diff
@@ -4017,6 +4045,10 @@ and the following Resource level attributes:
   rules defined by the current value of this attribute. This includes rejecting
   requests to add/delete/modify Versions or update this attribute in such a
   way as to make the existing Versions incompatible due to the new value.
+  For `compatibility` strategies that require understanding the sequence in
+  which versions need to be compatible, the server MUST use the [`ancestor`]
+  (#ancestor-attribute) to determine the sequence of Versions. For Versions
+  that have no `ancestor` value set, compatibility cannot be enforced.
```
duglin, line 4051: I think ancestor=null (or absent) is no longer possible. 
So even an Resource with one Version should probably point to itself, for consistency.

core/spec.md
```diff
@@ -4035,9 +4067,12 @@ and the following Resource level attributes:
   - `none` - No compatibility checking is performed.
 
 - Constraints:
-  - If present, it MUST be a case sensitive value from the model defined
+  - If present, it MUST be a case-sensitive value from the model defined
     enumeration range.
   - When not specified, the default value MUST be `none`.
+  - The value MAY change only when the new `compatibility` value can be
+    enforced across all existing versions. If that's not the case, the
```
duglin, line 4074: s/versions/Versions/

s/refuse the change/generate an error/   <- I'm trying to use this phrase everywhere so I can easily add a reference to the error we expect people to use.

tools/dict
```diff
@@ -1,6 +1,7 @@
 acks
 amqp
 amqps
+ancestor
```
duglin, line 4: you shouldn't need this one

core/spec.md
```diff
@@ -5528,6 +5566,35 @@ as defined below:
   - MUST NOT be present if the Resource's `hasdocument` model attribute is
     set to `false.
 
+##### `ancestor` Attribute
+- Type: String
+- Description: A string value that points to another version based on the
```
duglin, line 5571: I'd just say:` the `versionid` of this Version's ancestor.`

core/spec.md
```diff
@@ -5528,6 +5566,35 @@ as defined below:
...
+  The `ancestor` attribute MAY be set to any existing Version's `versionid`,
+  or to the `versionid` of the version itself, to indicate to the server that
+  this Version has no `ancestor`.
+
+  When creating a version without explicitly setting the `ancestor`
+  attribute, the server MUST set the `ancestor` to the most recent Version's
+  `versionid` attribute, based on the `createdat` attribute. If multiple
+  exist, the first one will be used in alphabetical order.
+
+  If an operation contains multiple versions with the `ancestor` attribute
```
duglin, line 5583: if a write operation...

core/spec.md
```diff
@@ -5528,6 +5566,35 @@ as defined below:
...
+  omitted, the server MUST order all versions based on the `createdat`
+  attribute and then alphabetically. The first Version will have the most
+  recent Version's `versionid` as its `ancestor` as clarified above.
+
+  When deleting a Version, the server MUST update the `ancestor` attribute
+  of any Version that points to the deleted Version to point to itself.
+
+- Constraints: the `ancestor` attribute MUST NOT be set to a value that
+  creates circular references between versions. For example, an operation that
+  makes Version A's ancestor B, and version B's ancestor A, MUST be refused
```
duglin, line 5593: MUST generate an error.

core/spec.md
```diff
@@ -5528,6 +5566,35 @@ as defined below:
...
+  by the server.
+  If the `ancestor` attribute is set to a `versionid` that does not exist
```
duglin, line 5595: Any attempt to set an `ancestor` attribute to a non-existing `versionid` MUST generate an error.

core/spec.md
duglin: Another situation that I think we need to mention in the spec, or at least the primer:
- singleversionroot=true
- maxversions limit is reached
- user tries to create a new version

The create needs to fail (assuming it'll delete the current root) because the net result is 2 roots/tree.

core/spec.md
lailabougria: Agreed, I missed this. I also added a sentence to the ancestor attribute:
>  When creating a version without explicitly setting the `ancestor`
  attribute, the server MUST set the `ancestor` to the most recent Version's
  `versionid` attribute, based on the `createdat` attribute. If multiple
  exist, the first one will be used in alphabetical order. **If no Versions 
  exist, the `ancestor` attribute MUST be set to the `versionid` of the 
  Version being created.**

core/spec.md
lailabougria: I added a section, it's possibly a bit too detailed in the sense that it repeats some of the info below, but it felt a bit odd to link to the maxversion part here.

core/spec.md
```diff
@@ -2677,11 +2685,18 @@ The following describes the attributes of Registry model:
...
   - When the limit is exceeded, implementations MUST prune Versions by
-    deleting the oldest Version (based on creation times) first, skipping the
-    Version marked as "default". An exception to this pruning rule is if
-    `maxversions` value is one (`1`) then the newest Version of the Resource
-    MUST always be the "default" and the `setdefaultversionsticky` aspect
-    MUST be `false`.
+    deleting the oldest Version first, skipping the
+    Version marked as "default". The strategy for finding the oldest Version
+    to delete is as follows:
+    1. Find all Versions that are roots (i.e. `ancestor` is the same as
```
lailabougria, line 2691: A structure observation: `ancestor` is referenced before it's described. We can link to it to fill that gap but I was wondering whether this is a concern to consider from other angles.

core/spec.md
```diff
@@ -5528,6 +5571,36 @@ as defined below:
...
+  The `ancestor` attribute MAY be set to any existing Version's `versionid`,
+  or to the `versionid` of the Version itself, to indicate to the server that
+  this Version has no `ancestor`.
+
+  When creating a version without explicitly setting the `ancestor`
+  attribute, the server MUST set the `ancestor` to the most recent Version's
+  `versionid` attribute, based on the `createdat` attribute. If multiple
+  exist, the first one will be used in alphabetical order. If no Versions
+  exist, the `ancestor` attribute MUST be set to the `versionid` of the
+  Version being created, making it a root.
```
lailabougria, line 5587:
```suggestion
  exist, the first one will be used in alphabetical order.
```

core/primer.md
```diff
@@ -674,3 +674,50 @@ in the entity in order to be model compliant.
 
 Likewise, implementations of the server must validate the entire entity
 against the new model, not just a subset of the entity's attributes.
+
+# Why are `ancestor` attributes that point to self allowed?
+
+The [`ancestor` attribute](spec.md#ancestor-attribute) is used to build a
+hierarchy of Versions aiding in compatibility checking when the
+[`compatibility` attribute](spec.md#compatibility-attribute) is set.
```
duglin, line 682: Technically, they build a tree whether compat is set or not.

core/primer.md
```diff
@@ -674,3 +674,50 @@ in the entity in order to be model compliant.
...
+# What's the oldest Version of a Resource?
+
+The oldest Version of a Resource isn't necessarily the one with the oldest
+`createdat` timestamp, because the `ancestor` attribute takes precedence
+over the `createdat` attribute. This is because the ancestor tree more
+accurately describes an ordered lineage of the Versions than timestamps. In
+the case where `groups.resources.singleversionroot` is set to `true`, the oldest
+Version of a Resource (or the first one ordered alphabetically should
+multiple Versions with the same `createdat` timestamp exist) will always be
+the oldest one in practice as only a single root exists.
```
duglin, line 702: I'm not sure we need the previous sentence - I think the stuff before that covers it and this sentence is introducing stuff (dup createdat timestamp) that might just confuse people.

core/primer.md
```diff
@@ -674,3 +674,50 @@ in the entity in order to be model compliant.
...
+To accurately determine the oldest Version of a Resource, independent of the
```
duglin, line 703: I think you can remove the rest of this since it's part of the spec itself already

core/spec.md
```diff
@@ -2677,11 +2654,18 @@ The following describes the attributes of Registry model:
...
-    deleting the oldest Version (based on creation times) first, skipping the
-    Version marked as "default". An exception to this pruning rule is if
-    `maxversions` value is one (`1`) then the newest Version of the Resource
-    MUST always be the "default" and the `setdefaultversionsticky` aspect
-    MUST be `false`.
+    deleting the oldest Version first, skipping the
+    Version marked as "default". The strategy for finding the oldest Version
+    to delete is as follows:
+    1. Find all Versions that are roots (i.e. `ancestor` is the same as
+       `versionid`)
```
duglin, line 2661: its `versionid`

core/spec.md
```diff
@@ -2677,11 +2654,18 @@ The following describes the attributes of Registry model:
...
+    1. If many exist, find the Version with the oldest `created` date
```
duglin, line 2662: s/created/createat/

core/spec.md
```diff
@@ -2677,11 +2654,18 @@ The following describes the attributes of Registry model:
...
+    1. If many exist, find the first Version according to alphabetical order
+       of the `versionid`-attribute
```
duglin, line 2664: s/-/ /
need to add in "lowest" in there someplace

core/spec.md
```diff
@@ -4035,9 +4039,34 @@ and the following Resource level attributes:
...
+    server MUST generate an error.
+
+#### `enforcecompatibility`
+- Name: `enforcecompatibility`
+- Type: Boolean
+- Description: Indicates a request to the server to enforce the `compatibility`
+  value defined by the owning Resource.
+  When set to `true`:
+- 1. The server MUST reject the request if the server cannot enforce the
+     compatibility for this type of Resource.
```
duglin, line 4056: Perhaps just say: ...cannot enforce the desired compatibility for the data to be validated for the Resource's Versions.

1) I wasn't sure what "type of Resource" meant since we never really talk about "contenttype" in all of this.
2) technically, someone could define a compat mode that checks more than the doc, it could look at the xreg data too. In fact, it should to some extent to check the "contenttype", but in general if hasdoc=false then "compat" for that kind of resource might involve checking the xReg metadata across versions.

So I think the more generic statement covers all future complex cases too

core/spec.md
```diff
@@ -5528,6 +5565,36 @@ as defined below:
   - MUST NOT be present if the Resource's `hasdocument` model attribute is
     set to `false.
 
+##### `ancestor` Attribute
+- Type: String
+- Description: The `versionid` of this Version's ancestor.
+
+  The `ancestor` attribute MAY be set to any existing Version's `versionid`,
```
duglin, line 5572: s/MAY/MUST/

core/spec.md
```diff
@@ -5528,6 +5565,36 @@ as defined below:
...
+  or to the `versionid` of the Version itself, to indicate to the server that
```
duglin, line 5573: `to indicate that this Version has no "ancestor"`.
The "to the server" implies it's some kind of signal from the client to the server, which it is sometimes, but in general we should talk about the meaning of the attribute when it appears in the datastore.
```suggestion
  or to the `versionid` of the Version itself, to indicate to the server that
```

core/spec.md
```diff
@@ -5528,6 +5565,36 @@ as defined below:
...
+  this Version has no `ancestor`.
+
+  When creating a version without explicitly setting the `ancestor`
```
duglin, line 5576: s/version/Version/

core/spec.md
```diff
@@ -5528,6 +5565,36 @@ as defined below:
...
+  attribute, the server MUST set the `ancestor` to the most recent Version's
+  `versionid` attribute, based on the `createdat` attribute. If multiple
+  exist, the first one will be used in alphabetical order. If no Versions
+  exist, the `ancestor` attribute MUST be set to the `versionid` of the
+  Version being created, making it a root.
+
+  If a write operation contains multiple versions with the `ancestor` attribute
```
duglin, line 5583: s/versions/Versions/

core/spec.md
```diff
@@ -5528,6 +5565,36 @@ as defined below:
...
+  omitted, the server MUST order all versions based on the `createdat`
```
duglin, line 5584: s/versions/Versions/

core/spec.md
```diff
@@ -5528,6 +5565,36 @@ as defined below:
...
+  attribute and then alphabetically. The first Version will have the most
+  recent Version's `versionid` as its `ancestor` as clarified above.
+
+  When deleting a Version, the server MUST update the `ancestor` attribute
+  of any Version that points to the deleted Version to point to itself.
+
+- Constraints: the `ancestor` attribute MUST NOT be set to a value that
+  creates circular references between versions. For example, an operation that
```
duglin, line 5592: s/versions/Versions/

core/spec.md
```diff
@@ -5528,6 +5565,36 @@ as defined below:
...
+  makes Version A's ancestor B, and version B's ancestor A, MUST generate an
```
duglin, line 5593: v->V

duglin:
I'll reread it again later, but it looks right based on our chat today! thank

core/spec.md
```diff
@@ -4035,9 +4039,34 @@ and the following Resource level attributes:
...
+- 1. The server MUST reject the request if the server cannot enforce the
+     compatibility for this type of Resource.
+  1. The server MUST reject all attempts to create/update a Resource (or its
+     Versions) that would result in those entities violating the stated
+     compatibility statement.
+  1. The server MUST reject requests to update the `compatibility`
+  attribute value if any of the Resource's Versions would violate the
+  compatibility rules.
+
+  A value of `false` indicates that the server MUST NOT perform any
```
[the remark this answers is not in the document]
lailabougria: Isn't that covered by what I added to the primer? Or do you want to make it extra explicit here?

> The `compatibility` attribute is a statement made by the authority managing the
registry about the compatibility guarantees of the Resource. The authority
is expected to guarantee the configured `compatibility`.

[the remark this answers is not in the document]
duglin: oops, forgot about that.  Can't decide if it's important enough to be in the spec or not.

however, I was wondering if we should revisit your previous suggestion of calling it something like "compatauthority" (with "server" or "external" as values). What's nice about this is we avoid the potential confusion around "enforcecompat:false" being interpreted as "the compat mode isn't valid". "compatauthority" (I think) will encourage people to think that "compat" is still valid, but "who" is offering the guarantee changes. More thinking needed...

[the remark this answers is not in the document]
lailabougria: The main challenge in such a solution is that we can't in any way ensure that that value will be true in all cases, especially in the context of export-import.
So these are the scenarios, I think:
- if you have a server that can enforce and is enforcing, this will be set to server. 
- If you have a server that can enforce and is **not** enforcing, this will be set to external.
- if you have a server that can't enforce, this will be set to external.
- if you don't have a server (e.g., file server), it has to be set by the registry's authority, and it could be anything (still ok IMHO, as this will be the case for other thing,s too)
- if you export from a server that can and did enforce, and you import to a server that can't enforce, the import will be rejected
- if you export from a server that can and did enforce, and you import to a server that can enforce, the import will be invalidated and only accepted if compatibility is valid
- if you export from a server that can and did enforce, and you import into a file server, ideally, the value would be changed to external, but we can't enforce that.

If we're ok with the above scenarios, we can introduce it. However, there are two cases in which the registry may not be conveying what actually happened, and I mostly feel a bit icky about the last one. In that scenario, that registry will convey that:
- it has a server that can and will enforce
- it has a resource with versions that were enforced when created
- however, any changes or additions will not be enforced; at this point, the registry is "lying".

But again, this is a scenario that can't be solved, and maybe this is something we need to accept and where the responsibility is at the registry's authority, not in the spec to address. The upside would be as you mentioned, that the user has a better understanding of who did the validation, at least in the majority of cases.

Should we include this, it would be nice to have it extensible, just like the compatibility field, so that individual registries can add more options for their users (e.g., client-side, manual, etc.).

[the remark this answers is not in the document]
duglin: I think I'm comfortable with those situations because in the end the boolean values have the exact same meaning as external/server, just w/o the worry about people misinterpreting what "false" means.

As for making it extensible, no real objection because anything except "server" would have the same semantics/meaning and any extension value that might have any addition meaning to it would be up for that implementation (or client) to explain to their users. So, someone choose to put the userid in there of the person who uploaded it.... cool or risky? dunno.

core/primer.md
```diff
@@ -674,3 +674,64 @@ in the entity in order to be model compliant.
 
 Likewise, implementations of the server must validate the entire entity
 against the new model, not just a subset of the entity's attributes.
+
+# Why are `ancestor` attributes that point to themselves allowed?
+
+The [`ancestor` attribute](spec.md#ancestor-attribute) is used to build a
+hierarchy of Versions to facilitate compatibility checking when the
+[`compatibility` attribute](spec.md#compatibility-attribute) is set.
+Generally speaking, implementations are not expected to allow users to
```
duglin, line 683: This comes really close to saying people SHOULD NOT have multiple roots, and yet we default singleversionroot to 'false'. I'd prefer if we remove this "editorial" comment since it feels inconsistent and it kind of bugs me that we're suggesting people not use it. I don't think we lose anything by dropping it (and the next sentence).

core/primer.md
```diff
@@ -674,3 +674,64 @@ in the entity in order to be model compliant.
...
+explicitly create multiple roots for a single Version. However, the
+specification does not mandate this as there may be cases where it is useful
+to allow for multiple roots. In some cases it may even be unavoidable, for
+example when the `maxversions` attribute (See `groups.resources.maxversions`
+under [Resource Model](spec.md#registry-model)) is set to a value that forces
+pruning of the Version tree. In such cases, when deleting the oldest version,
```
duglin, line 689: v->V

core/primer.md
```diff
@@ -674,3 +674,64 @@ in the entity in order to be model compliant.
...
+this could result in a new root being created when there are multiple
+decedents of the deleted Version.
+
+# Creating versions with `singleversionroot` and a `maxversions` limit
```
duglin, line 693: v->V

core/primer.md
```diff
@@ -674,3 +674,64 @@ in the entity in order to be model compliant.
...
+
+When the `groups.resources.maxversions` attribute of the [Resource
+Model](spec.md#registry-model) is set to a specified value for a Resource, the
+server will prune versions when attempting to create a new version that would
+exceed this value. However, there may be cases in which the creation of new
+versions may be blocked, due to server being unable to prune versions.
+This may occur when the `groups.resources.singleversionroot` attribute of
+the [Resource Model](spec.md#registry-model) is set to `true`.
+
+Consider a scenario in which 3 versions exist: v1 is the root (and therefore
```
duglin, line 703: v->V above
and you can just say `v1` instead of "its versionid" in the next line. To be extra clear.

core/primer.md
```diff
@@ -674,3 +674,64 @@ in the entity in order to be model compliant.
...
+pruning of the Version tree. In such cases, when deleting the oldest version,
+this could result in a new root being created when there are multiple
+decedents of the deleted Version.
+
+# Creating versions with `singleversionroot` and a `maxversions` limit
+
+When the `groups.resources.maxversions` attribute of the [Resource
+Model](spec.md#registry-model) is set to a specified value for a Resource, the
+server will prune versions when attempting to create a new version that would
+exceed this value. However, there may be cases in which the creation of new
```
duglin, line 698: This paragraph might need to be tweaked because even if maxversions=0 the server could still delete Versions.

Perhaps:
There may be situations where the server will need to prune Versions automatically,
for example to enforce the `maxversions` semantics. In these case if `singleversionroot`
is set to `true` and deleting a Version would result in multiple Versions being roots
then the operation that initiated this situation must be rejected.

core/primer.md
```diff
@@ -674,3 +674,64 @@ in the entity in order to be model compliant.
...
+their `ancestor` attribute set to v1. In addition, the `groups.resources.
+maxversions` is set to 3. When creating a new version, the server will find
+the oldest version (v1) and attempt to prune it. However, deleting v1 would
+mean that v2 and v3 would become roots, as both of them would need to point
+to themselves. This is exactly the behavior that the
+`groups.resources.singleversionroot` attribute prevents when set to `true`.
+Therefore, the server is unable to prune versions and will block the
+creation of a new version. To resolve this, the user will have to manually
+delete v2 or v3 to allow the server to prune the oldest version (v1) before
+creating a new version.
```
duglin, line 714: several v->V above

core/primer.md
```diff
@@ -674,3 +674,64 @@ in the entity in order to be model compliant.
...
+statement made by the authority managing the registry about the
+compatibility guarantees of the Resource. The authority is expected to
+guarantee the configured `compatibility`. The
+[`compatibilityauthority` attribute](spec.md#compatibilityauthority-attribute)
+represents who the enforcing authority is. Any requests to set the authority
+to the server when the server cannot perform compatibility checking will be
+refused. However, in cases where the registry is hosted on a file-server or
+blob storage, there is no real server that has the ability to validate such
+requests. In such cases, the `compatibilityauthority` attribute could be set
+to `server` while the server has no ability to enforce compatibility. It's
```
duglin, line 735: This "In such cases" sentence, while correct feels like it's missing something... like an explanation for why it's ok.
In thinking about how to fix it, I wonder if it would be better to combine these 3 sentences to say:
- in cases where the hosting service isn't backed by an xReg implementation (e.g. blob-store), it is recommended that "compatauthority" is set to "external" as that accurately reflects the situation
- however, there may be cases where it is appropriate to set it to "server". For example, if the file-server based xRegistry is an alternative mechanism to retrieve data from another xRegistry service that did validate the data (ie. it was exported to the file-server), and there's a desire to not expose this implementation/deployment choice to the end-user. In other words, the use of the file-server as a caching service should be transparent to their users.

or something like that. WDYT?

lailabougria: Yes that would work really well!

core/primer.md
```diff
@@ -674,3 +674,72 @@ in the entity in order to be model compliant.
...
 against the new model, not just a subset of the entity's attributes.
+
+# Multi-root ancestor hierarchies
+
+The [`ancestor` attribute](spec.md#ancestor-attribute) is used to build a
+hierarchy of Versions to facilitate compatibility checking when the
+[`compatibility` attribute](spec.md#compatibility-attribute) is set. There
+are cases in which it's desirable to create multiple roots. In some
+cases it may even be unavoidable, for example when the `maxversions`
+attribute (See `groups.resources.maxversions` under
```
duglin, line 685: s/See/see/

core/primer.md
```diff
@@ -674,3 +674,72 @@ in the entity in order to be model compliant.
...
+makes the ancestor explicit, and avoids ambiguity.
+
+# Pruning Versions with `singleversionroot` enabled
+
+There are cases in which the server will need to prune Versions. For
+example, this can happen when attempting to create a new Version that would
+exceed the value set on the `groups.resources.maxversions` attribute of the
+[Resource Model](spec.md#registry-model), or when adjusting this attribute's
+value that is smaller than the number of existing Versions. In such
+scenarios, the server may be unable to prune Versions, when the `groups.
```
duglin, line 702: group.   needs to be moved to the next line. The newline char cause a space to appear after the period and before the "resources...." on the next line.

core/primer.md
```diff
@@ -674,3 +674,72 @@ in the entity in order to be model compliant.
...
+resources.singleversionroot` attribute of the
+[Resource Model](spec.md#registry-model) is set to `true` and the request
+must be rejected.
+
+Consider a scenario in which 3 Versions exist: v1 is the root (and therefore
+has its `ancestor` attribute set to v1), and v2 and v3 both have
+their `ancestor` attribute set to v1. In addition, the `groups.resources.
```
duglin, line 709: as above, move "groups.resources." to the next line

core/primer.md
```diff
@@ -674,3 +674,72 @@ in the entity in order to be model compliant.
...
+maxversions` is set to 3. When creating a new Version, the server will find
+the oldest Version (v1) and attempt to prune it. However, deleting v1 would
+mean that v2 and v3 would become roots, as both of them would need to point
+to themselves. This is exactly the behavior that the
+`groups.resources.singleversionroot` attribute prevents when set to `true`.
+Therefore, the server is unable to prune Versions and will block the
+creation of a new Version. To resolve this, the user will have to manually
+delete v2 or v3 to allow the server to prune the oldest Version (v1) before
+creating a new Version.
```
duglin, line 718: I like it!  ^^

core/primer.md
```diff
@@ -674,3 +674,72 @@ in the entity in order to be model compliant.
...
+The oldest Version of a Resource isn't necessarily the one with the oldest
+`createdat` timestamp, because the `ancestor` attribute takes precedence
+over the `createdat` attribute. This is because the ancestor tree more
+accurately describes an ordered lineage of the Versions than timestamps.
+
+# What does the `compatibilityauthority` attribute convey?
+
+The [`compatibility` attribute](spec.md#compatibility-attribute) is a
+statement made by the authority managing the registry about the
+compatibility guarantees of the Resource. The authority is expected to
```
duglin, line 731: I wonder if we should say:   ... compatibility guarantees between the Versions of the Resource.

Otherwise it might sound like it's possible to check compat across Resources. Or it could just be me

core/spec.md
```diff
@@ -2698,11 +2674,18 @@ The following describes the attributes of Registry model:
     it is valid for an implementation to only support one (`1`) Version when
     `maxversions` is set to `0`.
   - When the limit is exceeded, implementations MUST prune Versions by
-    deleting the oldest Version (based on creation times) first, skipping the
-    Version marked as "default". An exception to this pruning rule is if
-    `maxversions` value is one (`1`) then the newest Version of the Resource
-    MUST always be the "default" and the `setdefaultversionsticky` aspect
-    MUST be `false`.
+    deleting the oldest Version first, skipping the
+    Version marked as "default". The strategy for finding the oldest Version
```
duglin, line 2678: hmmm I wonder if our algorithm to determine the "latest" needs to change from "latest createdat timestamp + highest alpha" to be:
- find all leaves
- if more than one: find newest
- if more than one: find highest alpha

Otherwise I think we might be inconsistent.  Thoughts?

core/spec.md
```diff
@@ -2698,11 +2674,18 @@ The following describes the attributes of Registry model:
...
+    to delete is as follows:
+    1. Find all Versions that are roots (i.e. `ancestor` is the same as its
+       `versionid`)
+    1. If many exist, find the Version with the oldest `createdat` date
+    1. If many exist, find the lowest Version according to alphabetical order
+       of the `versionid` attribute
+    Once the single oldest Version is determined, delete it.
+    An exception to this pruning rule is if `maxversions` value is one (`1`)
```
duglin, line 2686: I know this was here before your PR, but it doesn't read correctly to me now...
What do you think about:
A special case for the pruning rules is that if `maxversions` value is one (`1`), then the "default" Version is not skipped, which means it'll be deleted and the new Version will become "default".

Or...

above change: ...skipping the Version marked as "default".
to: ...skipping the Version marked as "default" if `maxversions` is not set to one (`1`).

Thoughts?

core/spec.md
```diff
@@ -4035,13 +4035,15 @@ and the following Resource level attributes:
...
+  If the `compatibilityauthority` attribute is set to `server` then
   implementations of this specification are REQUIRED to perform the proper
   compatibility checks to ensure that all Versions of a Resource adhere to the
-  rules defined by the current value of this attribute. This includes
-  generating an error in response to requests to add/delete/modify Versions or
-  update this attribute in such a way as to make the existing Versions
-  incompatible due to the new value.
+  rules defined by the current value of this attribute.
+  For `compatibility` strategies that require understanding the sequence in
+  which versions need to be compatible, the server MUST use the [`ancestor`]
```
duglin, line 4043: v->V
I was wondering if you got them all :-)

core/spec.md
```diff
@@ -4061,10 +4063,42 @@ and the following Resource level attributes:
...
     enumeration range.
   - When not specified, the default value MUST be `none`.
   - The enumeration range MUST include `none` as a valid value.
+  - The value MAY change only when the new `compatibility` value can be
+    enforced across all existing Versions. If that's not the case, the
+    server MUST generate an error.
+
+#### `compatibilityauthority` Attribute
+- Name: `compatibilityauthority`
+- Type: Boolean
```
duglin, line 4076: string

core/spec.md
```diff
@@ -4061,10 +4063,42 @@ and the following Resource level attributes:
...
+- Description: Indicates the authority that enforces the `compatibility`
+  value defined by the owning Resource.
+
+  This specification defines the following enumeration values. Implementations
+  MAY choose to extend this list.
+  - `external` - The compatibility is enforced by an external authority.
+  - `server` - The compatibility is enforced by the server.
+
+  When set to `server`:
+  1. The server MUST reject the request if the server cannot enforce the
```
duglin, line 4086: s/reject the/reject any/

otherwise I wasn't sure which request you were referring to.

duglin: actually, which request were you thinking of that isn't covered by the next 2 bullets?

core/spec.md
```diff
@@ -4061,10 +4063,42 @@ and the following Resource level attributes:
...
+     compatibility for the data to be validated for the Resource's Versions.
+  1. The server MUST reject all attempts to create/update a Resource (or its
+     Versions) that would result in those entities violating the stated
+     compatibility statement.
+  1. The server MUST reject requests to update the `compatibility`
```
duglin, line 4091: Isn't this covered by the last sentence in the previous attribute?

core/spec.md
```diff
@@ -4532,7 +4567,8 @@ then the resulting serialization of the source Resource would be:
     "createdat": "2024-01-01-T12:00:00",
     "modifiedat": "2024-01-01-T12:01:00",
     "readonly": false,
-    "compatibility": "none"
+    "compatibility": "none",
+    "compatibilityauthority": "external",
```
duglin, line 4571: remove trailing comma

core/spec.md
```diff
@@ -5387,7 +5426,8 @@ When serialized as a JSON object, the Version entity adheres to this form:
 
   "RESOURCEurl": "URL", ?                  # If not local
   "RESOURCE": ... Resource document ..., ? # If inlined & JSON
-  "RESOURCEbase64": "STRING" ?             # If inlined & ~JSON
+  "RESOURCEbase64": "STRING", ?            # If inlined & ~JSON
+  "ancestor": "STRING"
```
duglin, line 5430: let's move this into the previous section since that's where all of the "normal" attributes appear.
I like it between "modifiedat" and "contenttype" since it's kind of related to timestamps.

duglin: I think there are other pseudo schema chunks that need to be touched by this PR to add the attribute - eg the main one at the top of the doc.

core/spec.md
```diff
@@ -5427,6 +5467,8 @@ and the following Version level attributes:
 - [`RESOURCEurl`](#resourceurl-attribute) - OPTIONAL.
 - [`RESOURCE`](#resource-attribute) - OPTIONAL.
 - [`RESOURCEbase64`](#resourcebase64-attribute) - OPTIONAL.
+- [`ancestor`](#ancestor-attribute) - REQUIRED in API and document views.
```
duglin, line 5470: move up here too

core/spec.md
```diff
@@ -5546,6 +5588,42 @@ as defined below:
   - MUST NOT be present if the Resource's `hasdocument` model attribute is
     set to `false.
 
+##### `ancestor` Attribute
```
duglin, line 5591: will need to move this up too

core/spec.md
```diff
@@ -5546,6 +5588,42 @@ as defined below:
...
+- Type: String
+- Description: The `versionid` of this Version's ancestor.
+
+  The `ancestor` attribute MUST be set to any existing Version's `versionid`,
```
duglin, line 5595: ...MUST be set to the `versionid` of this Version's ancestor (parent). If this Version is a root of an ancestor hierarchy tree (i.e. it has no parent) then it MUST be set to its own `versionid` value.

core/spec.md
```diff
@@ -5546,6 +5588,42 @@ as defined below:
...
+  or to the `versionid` of the Version itself, to indicate that this Version
+  has no `ancestor`.
+
+  When creating a Version without explicitly setting the `ancestor`
+  attribute, the server MUST set the `ancestor` to the most recent Version's
+  `versionid` attribute, based on the `createdat` attribute. If multiple
+  exist, the lowest Version will be determined alphabetically. If no
```
duglin, line 5602: s/lowest/highest/   no?

And probably: If multiple exist, then the Version with the highest alphabetical `versionid` MUST be chosen.

core/spec.md
```diff
@@ -5546,6 +5588,42 @@ as defined below:
...
+  Versions exist, the `ancestor` attribute MUST be set to the `versionid` of the
+  Version being created, making it a root.
+
+  If a write operation contains multiple Versions with the `ancestor` attribute
+  omitted, the server MUST order all Versions based on the `createdat`
```
duglin, line 5607: s/all Versions/all of those Versions/

core/spec.md
```diff
@@ -5546,6 +5588,42 @@ as defined below:
...
+  attribute and then alphabetically. The first Version will have the most
```
duglin, line 5608: and then alphabetically (lowest first).

core/spec.md
```diff
@@ -5546,6 +5588,42 @@ as defined below:
...
+  recent Version's `versionid` as its `ancestor` as clarified above.
+
+  When deleting a Version, the server MUST update the `ancestor` attribute
+  of any Version that points to the deleted Version to point to itself,
+  making it a new root.
+
+  If a create operation asks the server to choose the `vID` when creating a
```
duglin, line 5615: s/vID/versionid/

core/spec.md
```diff
@@ -5546,6 +5588,42 @@ as defined below:
...
+  root Version, the `versionid` is not yet known and therefore cannot be
+  assigned a value. In those cases a value of `request` MAY be used as a way
```
duglin, line 5617: assigned a value -> used in the `ancestor` attribute

core/primer.md
lailabougria: Can't hurt to be extra explicit!

core/spec.md
lailabougria: Inconsistent with what? 
I don't really see it, what would it mean to find all leaves? The top of the leaf? What is your concern with finding all roots? When does it become inconsistent? Even when singleversionroot is set to true, it's still "find all roots", there will just be one.

core/spec.md
lailabougria: I was thinking of a request to set the compatibilityauthority to server. I'll clarify this.

core/spec.md
lailabougria: image

core/spec.md
lailabougria: Adding the first suggestion, it's a bit more verbose but I think it will be more clear to readers.

core/spec.md
lailabougria: I do see them as different:
1. In this case you're creating versions that may or may not be compatible, which is checked
2. In this case you're updating the compatbility setting, causing revalidation and rejecting if it can't

The intent of the user is different which I think is worth making explicit

core/spec.md
lailabougria: Let's say there is 1 root, with the 2 oldest versions that share the same date, one is A and one B. We'd want to remove A right? A is "lower" than B the way I see it, that's why I said "lowest"

core/spec.md
```diff
@@ -5546,6 +5590,43 @@ as defined below:
   - MUST NOT be present if the Resource's `hasdocument` model attribute is
     set to `false.
 
+##### `ancestor` Attribute
+- Type: String
+- Description: The `versionid` of this Version's ancestor.
+
+  The `ancestor` attribute MUST be set to the `versionid` of this Version's
+  ancestor (parent). If this Version is a root of an ancestor hierarchy tree
+  (i.e. it has no parent) then it MUST be set to its own `versionid` value.
```
lailabougria, line 5599: I adjusted this based on your feedback, but now I'm realizing this is the only place where we mention the place parent. Should we do that? Isn't ancestor clear enough?

I suggest omitting (parent) and (i.e. it has no parent).

core/spec.md
lailabougria: I hope I got them all, please take a look

core/spec.md
```diff
@@ -5468,6 +5514,43 @@ as defined below:
   - `true`
   - `false`
 
+##### `ancestor` Attribute
+- Type: String
+- Description: The `versionid` of this Version's ancestor.
+
+  The `ancestor` attribute MUST be set to the `versionid` of this Version's
+  ancestor (parent). If this Version is a root of an ancestor hierarchy tree
+  (i.e. it has no parent) then it MUST be set to its own `versionid` value.
```
lailabougria, line 5523: See https://github.com/xregistry/spec/pull/266#discussion_r2008319000

core/spec.md
```diff
@@ -5468,6 +5514,43 @@ as defined below:
...
+
+  When creating a Version without explicitly setting the `ancestor`
```
lailabougria, line 5525: See https://github.com/xregistry/spec/pull/266#discussion_r2008059251

core/spec.md
duglin: But you're not deleting, you're trying to find the newest so you can set the new version's ancestor, no?

core/spec.md
duglin: what I mean is that when we need to "find the oldest" (e.g. for deletion) we find all roots, find oldest, then lowest alpha.
So, when we want to find the latest (to become the ancestor for a new version) we need to do the opposite, find all leaves, find the newest, then highest alpha. If ancestor is of higher precedence for finding "oldest" then it needs to take precedence for finding newest. 

Take the extreme case when we have:
```
A<-B
 \-C
```
A's createdat=12:03pm. (root)
B's createdat=12:02pm. (branch 1)
C's createdat=12:02pm. (branch 2)
No matter the timestamps, I would always want a new version's ancestor to default to either B or C. So between those 2 which is the newest? Look at highest timestamp and if they're the same we need to find the opposite of the lowest alpha.... which means the highest alpha.  C in this case.

Sorry if it wasn't clear because I added this comment during the "delete" case... it's just when I realized we were missing the opposite usecase.

core/spec.md
duglin: The sentence in the previous attribute ("compatibility") says: "The value MAY change only when the new `compatibility` value can be enforced across all existing Versions", is basically the same as:   `The server MUST reject requests to update the `compatibility` attribute value if any of the Resource's Versions would violate the compatibility rules`, isn't it?

And actually, I prefer the "The server MUST..." because the use of MAY isn't really correct because it's awkward to say "MAY only" instead of "MUST ...something...". "MAY" is for optional stuff.

core/spec.md
duglin: yeah, that's fine

core/spec.md
duglin: "find all leaves" == 'find all versions where no other version points to it via "ancestor"'

core/primer.md
```diff
@@ -674,3 +674,72 @@ in the entity in order to be model compliant.
...
+cases it may even be unavoidable, for example when the `maxversions`
+attribute (see `groups.resources.maxversions` under
+[Resource Model](spec.md#registry-model)) is set to a value that forces
+pruning of the Version tree. In such cases, when deleting the oldest Version,
+this could result in a new root being created when there are multiple
+decedents of the deleted Version.
+
+To signal that a Version represents a root of a hierarchy, the `ancestor`
+attribute has its value set to the Version's `versionid` attribute. This
+makes the ancestor explicit, and avoids ambiguity.
```
duglin, line 693: ... and avoids the possible ambiguity of using another value such as `null` which, based on the scenario, could mean "no ancestor" or "default to the latest".

w/o expanding on the thought I'm not sure a new reader would understand where the ambiguity comes from.

core/primer.md
```diff
@@ -674,3 +674,72 @@ in the entity in order to be model compliant.
...
+creation of a new Version. To resolve this, the user will have to manually
+delete v2 or v3 to allow the server to prune the oldest Version (v1) before
+creating a new Version.
+
+# What's the oldest Version of a Resource?
+
+The oldest Version of a Resource isn't necessarily the one with the oldest
+`createdat` timestamp, because the `ancestor` attribute takes precedence
+over the `createdat` attribute. This is because the ancestor tree more
+accurately describes an ordered lineage of the Versions than timestamps.
```
duglin, line 725: Perhaps add something likle:
- - -
Likewise, the newest Version of a Resource is found by examining all of the leaves of the hierarchy (ie. all Versions that are not used as ancestors for any other Versions), and then finding the one with the newest `createdat` timestamp.
- - -
We should probably mention the alpha stuff for both of these cases too.

Maybe change the title to: What's the oldest/newest Version of a Resource
?

core/spec.md
```diff
@@ -2698,11 +2675,18 @@ The following describes the attributes of Registry model:
...
-    MUST always be the "default" and the `setdefaultversionsticky` aspect
-    MUST be `false`.
+    deleting the oldest Version first, skipping the
+    Version marked as "default". The strategy for finding the oldest Version
+    to delete is as follows:
+    1. Find all Versions that are roots (i.e. `ancestor` is the same as its
+       `versionid`)
+    1. If many exist, find the Version with the oldest `createdat` date
+    1. If many exist, select the first Version sorting all Versions in
+       ascending alphabetical order based on the `versionid` attribute
```
duglin, line 2685: let's add periods to the end of each bullet point.

core/spec.md
```diff
@@ -4035,13 +4036,15 @@ and the following Resource level attributes:
...
-  rules defined by the current value of this attribute. This includes
-  generating an error in response to requests to add/delete/modify Versions or
-  update this attribute in such a way as to make the existing Versions
-  incompatible due to the new value.
+  rules defined by the current value of this attribute.
+  For `compatibility` strategies that require understanding the sequence in
+  which Versions need to be compatible, the server MUST use the [`ancestor`]
+  (#ancestor-attribute) to determine the sequence of Versions. For Versions
+  that have the `ancestor` value set to its `versionid` attribute,
+  compatibility cannot be enforced.
```
duglin, line 4047: This last sentence is missing something... 1) it only focuses on **backwards** and leaves **forwards** out of the picture, 2) "cannot be enforced" sounds like some kind of bad situation, so perhaps it might be better to say something like "implicitly passes the checks due to there not being an ancestor". Of course, that again only focuses on backwards.

Seems to me we either need to enhance that sentence to be more abstract:
- when the previous, or next, Version in the hierarchy is needed to determine adherence to the compatibility rules, if none exists then the implied result of the validation is "true"

or, we can just drop the entire sentence because it's kind of obvious and the definition of the rules SHOULD tell people that "no parent/child means it passes".

core/spec.md
```diff
@@ -4061,10 +4064,41 @@ and the following Resource level attributes:
 
 - Constraints:
   - REQUIRED.
-  - If present, it MUST be a case sensitive value from the model defined
```
duglin, line 4064: I wonder if the definitions above need to take into account "ancestor"? To us it's obvious what we mean, but when we only talk about "next oldest" I wonder if someone well read that to only mean "look at timestamps".

core/spec.md
```diff
@@ -5468,6 +5511,48 @@ as defined below:
...
+  attribute, the server MUST set the `ancestor` to the most recent Version's
+  `versionid` attribute. If no Versions exist, the `ancestor` attribute MUST
+  be set to the `versionid` of the Version being created, making it a root.
+  If there are existing Versions, the strategy for finding the most recent
+  Version is as follows:
+  1. Find all Versions that are not referenced as an `ancestor` in other
+     Versions
+  1. If many exist, find the Version with the most recent `createdat` date
+  1. If many exist, select the first Version sorting all Versions in
+     descending alphabetical order based on the `versionid` attribute
```
duglin, line 5532: add periods to end of each

core/spec.md
lailabougria: Good catch. I tried multiple options and ended up dropping the last sentence, mostly because it gets too much into the crux of the individual compatibility strategies where above we mention that the spec doesn't really dictate what those mean. Although it's unlikely that backward and forward change, I think it might read a bit clunky in the context of their transitive options as they mean compatibly across all parents/descendants and then only the next version is not sufficient.

core/spec.md
lailabougria: Wouldn't the above sentence cover that?
> For `compatibility` strategies that require understanding the sequence in
  which Versions need to be compatible, the server MUST use the [`ancestor`]
  (#ancestor-attribute) to determine the sequence of Versions.

We are clearly saying that any sequencing should be based on the ancestor. Would it be more clear if we move that sentence down, right under the enumeration options?

duglin: it's probably fine as-is. let's see if anyone else from the team raises a concern

core/primer.md
```diff
@@ -674,3 +674,83 @@ in the entity in order to be model compliant.
...
+# What's the oldest/newest Version of a Resource?
+
+The oldest Version of a Resource isn't necessarily the one with the oldest
+`createdat` timestamp, because the `ancestor` attribute takes precedence
+over the `createdat` attribute. This is because the ancestor tree more
+accurately describes an ordered lineage of the Versions than timestamps. In
+the case multiple Versions exist with the same `createdat` timestamp, those
+Versions are sorted in a descending alphabetical order to select the oldest
+Version.
+The newest Version of a Resource is selected by searching in the opposite
```
duglin, line 731: Can you add a blank line before 731?

duglin:
You don't show "ancestor" on the Resource. As of now all Version attributes are replicated into the Resource. I think we need to add it to each serialization of the Resource where the Version attributes are included.

duglin:
@lailabougria we got pre-approval for this on today's weekly call, so if you can wrap up your last edits before tomorrow's call (or before tuesday's) then we should be able to approve v1.0-rc1 in time for @manuelottlik to talk about it at KubeCon (on wed)

core/spec.md
```diff
@@ -5472,6 +5524,48 @@ as defined below:
...
+  the most recent Version's `versionid` as its `ancestor` as clarified above.
+
+  When deleting a Version, the server MUST update the `ancestor` attribute
+  of any Version that points to the deleted Version to point to itself,
+  making it a new root.
+
+  If a create operation asks the server to choose the `versionid` when
+  creating a root Version, the `versionid` is not yet known and therefore
+  cannot be assigned the value in the `ancestor` attribute. In those cases a
+  value of `request` MAY be used as a way to reference the Version being
```
duglin, line 5561: I think this needs to be a MUST since we're already within the scope of "creating a root with an unknown id", so I don't think there are alternative available... they MUST use "request"

core/spec.md
```diff
@@ -5472,6 +5524,48 @@ as defined below:
...
+  processed in the current request.
+
+- Constraints: the `ancestor` attribute MUST NOT be set to a value that
```
duglin, line 5564: let's make these sub-bullets (2 of them)

core/spec.md
```diff
@@ -5812,6 +5910,9 @@ xRegistry-xid: /endpoints/ep1/messages/msg1/versions/1.0
 xRegistry-epoch: 2
 xRegistry-name: Blob Created
 xRegistry-isdefault: true
+xRegistry-createdat: TIME
+xRegistry-modifiedat: TIME
```
duglin, line 5916: use real timestamps here since it's meant to be an real example with real data

core/spec.md
```diff
@@ -6453,6 +6558,15 @@ as appropriate, including being specified in a language other than English.
 
 <!-- start-err-def -->
 
+#### ancestor_circular_reference
+
+* Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#ancestor_circular_reference`
+* Code: `400 Bad Request`
+* Instance: The request URL
```
duglin, line 6567: Let's make this a URL to the Version we're touching

duglin: in a large/tree update, it might hard to know which version we're talking about if more than one is trying to use the same ancestor value

schema/spec.md
```diff
@@ -414,6 +415,7 @@ Versions for a schema named `com.example.telemetrydata`:
             "1": {
               "schemaid": "com.example.telemetrydata",
               "versionid": "1",
+              "ancestor": "1",
```
duglin, line 418: I think it needs to move down after "description"

duglin: right before "format"

duglin: 2 more below

duglin:
Per 3/27 call - this was preemptively approved

PR #279, opened 2025-03-12
Week of March 10

duglin:
- more clean-up (try to make it clearer which Resource attributes are inherited from the default Version in pseudo schema)
- start to define our errors
- add checker to make sure all our defined errors are actually referenced
- s/relaxednames/namecharset/ - "strict" (default) or "extended"


duglin:
Approved on the 3/13 call

PR #281, opened 2025-03-19
Make compatiblewith an URI and improve description

### LINKED ISSUE #276: Clarification: Should `compatiblewith` just be a label or a top-level property in the meta-model?
The `compatiblewith` feature is currently just a label in the metamodel and not a top-level property of the metamodel. 

The information that a model is compatible to a shared model understanding is really important to get interoperability beyond the own xRegistry scope and make a shared ecosystem around the model standard possible. 

I wonder if this is obvious enough, as understanding that one model in xRegistry A is compatible with another Model in xRegistry B (usually via the same shared model contract, e.g. the CNCF Domain Specific Specifications) is crucial. 

It would be helpful if a model export / import can carry this information and that the implications of it are clear to the xRegistry admins. Not sure if the current solution via labels is good enough for this or not, @duglin could judge better. 

What I worry about: Let's say a company hosts their own xRegistry server and imports the message definitions model and customize it. How would an external xRegistry or consumer know that the model is compatible with their own understanding? It is possible to customize models in a way that they are not interoperable anymore. In that case, it would not be possible to use a shared ecosystem around that model (like we can do for OpenAPI / AsyncAPI).

Fannon:
Fixes #276 

•	make it a URI
•	MUST reference/represent an xRegistry model definition

## TODO

- [ ] How to make it a spec-defined optional attribute?
- [ ] Can we deep-link to a model group like "endpoints"? 
- [ ] What would be the final links to the official xRegistry standardized models?


endpoint/spec.md
```diff
@@ -36,7 +36,7 @@ this form:
   "documentation": "URL", ?
   "labels": {
     "modelversion": "1.0", ?
-    "compatiblewith": "https://github.com/xregistry/spec/blob/main/endpoint/spec.md", ?
+    "compatiblewith": "https://xregistry.soaphub.org/model/groups/endpoints", ? # TODO: Doesn't work, URL needs to change
```
duglin, line 39: would pointing to: https://github.com/xregistry/spec/blob/main/endpoint/model.json work?

duglin:
didn't you want to make this a top-level optional attribute instead of a label?

Fannon:
> didn't you want to make this a top-level optional attribute instead of a label?

I did the change now. Also added it to the `model.json` files. Is it correct in that way? Looks like I broke some python script.

As URI, I went with the github raw file links

endpoint/spec.md
Fannon: I'll use the raw link, so it's a JSON download, not HTML

endpoint/model.json
```diff
@@ -3,6 +3,8 @@
     "endpoints": {
       "singular": "endpoint",
       "plural": "endpoints",
+      "modelversion": "1.0",
```
Fannon, line 6:
```suggestion
      "modelversion": "0.5",
```

tools/dict
```diff
@@ -205,6 +206,7 @@ shortself
 shorturl
 siblingattributes
 singularid
+soaphub
```
Fannon, line 209:
```suggestion
```

core/spec.md
```diff
@@ -945,6 +949,35 @@ of the existing entity. Then the existing entity would be deleted.
...
 
+#### modelversion
+
+- Type: String
+- Description: The version of the model. It is common to use a combination
+  of major and minor version numbers.
+
+- Constraints:
+  - OPTIONAL.
+  - Only applicable on model definition.
```
duglin, line 960: you can delete the above line

core/spec.md
```diff
@@ -945,6 +949,35 @@ of the existing entity. Then the existing entity would be deleted.
...
+
+- Type: URI
+- Description: References / represents an xRegistry model definition that
+  the current model is compatible with. This is meant to express
+  interoperability between models in different xRegistries via using a
+  shared compatible model.
+
+- Constraints:
+  - OPTIONAL.
+  - Only applicable on model definition.
```
duglin, line 975: you can delete the above line

core/spec.md
```diff
@@ -945,6 +949,35 @@ of the existing entity. Then the existing entity would be deleted.
...
+  - Does not imply runtime validation of the claim.
+
+- examples:
```
duglin, line 978: Examples:
(cap E)

schema/spec.md
```diff
@@ -35,12 +35,12 @@ this form:
   "description": "STRING", ?
   "documentation": "URL", ?
   "labels": {
-    "modelversion": "1.0", ?
-    "compatiblewith": "https://github.com/xregistry/spec/blob/main/schema/spec.md", ?
-    "STRING": "STRING" *
+      "STRING": "STRING" *
```
duglin, line 38: outdent 2 spaces

core/spec.md
```diff
@@ -597,6 +599,8 @@ form:
 - `"labels": { "STRING": "STRING" * }`
 - `"createdat": "TIMESTAMP"`
 - `"modifiedat": "TIMESTAMP"`
+- `"modelversion": "STRING"`
+- `"compatiblewith": "URI"`
```
duglin, line 603: let's move these to where the other model attributes are defined: https://github.com/xregistry/spec/blob/dfde83550d8662beca8e89766e4be4e6d81d7816/core/spec.md#registry-model

This section is just for xreg metadata (normal user facing) attributes, not model attributes

core/spec.md
```diff
@@ -945,6 +949,35 @@ of the existing entity. Then the existing entity would be deleted.
 - Examples:
   - `2030-12-19T06:00:00Z`
 
+#### modelversion
```
duglin, line 952: this section and the next one should be moved under: https://github.com/xregistry/spec/blob/dfde83550d8662beca8e89766e4be4e6d81d7816/core/spec.md#registry-model

core/spec.md
```diff
@@ -138,6 +138,8 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
 
   "model": {                            # Only if inlined
     "labels": { "STRING": "STRING" * }, ?
+    "modelversion": "STRING", ?         # Version of the model
+    "compatiblewith": "URI", ?          # Statement of compatibility with model spec
```
duglin, line 142: Minor thing, but move these 2 to be before "labels", I think from an "importance" perspective, if they're specified I think they're more important than "labels", and I try to put the most important stuff first.

I tend to view "labels" as a bag for extensions... and so I try to keep them near the end of the list of attributes.

message/model.json
```diff
@@ -3,6 +3,8 @@
     "messagegroups": {
       "singular": "messagegroup",
       "plural": "messagegroups",
+      "modelversion": "1.0",
+      "compatiblewith": "https://raw.githubusercontent.com/xregistry/spec/refs/heads/main/message/model.json",
```
duglin, line 7: these should be move to the top now, right?

message/spec.md
```diff
@@ -40,12 +40,12 @@ this form:
   "description": "STRING", ?
   "documentation": "URL", ?
   "labels": {
-    "modelversion": "1.0", ?
-    "compatiblewith": "https://github.com/xregistry/spec/blob/main/message/spec.md", ?
-    "STRING": "STRING" *
+      "STRING": "STRING" *
```
duglin, line 43: 2 space indent, as before

message/spec.md
```diff
@@ -40,12 +40,12 @@ this form:
...
   }, ?
   "createdat": "TIMESTAMP",
   "modifiedat": "TIMESTAMP",
+  "modelversion": "1.0", ?
+  "compatiblewith": "https://raw.githubusercontent.com/xregistry/spec/refs/heads/main/message/model.json", ?
```
duglin, line 48: per our chat last week, these can go away

schema/model.json
```diff
@@ -3,14 +3,14 @@
     "schemagroups": {
       "singular": "schemagroup",
       "plural": "schemagroups",
-
+      "modelversion": "1.0",
+      "compatiblewith": "https://raw.githubusercontent.com/xregistry/spec/refs/heads/main/schema/model.json",
```
duglin, line 7: move to top level

schema/spec.md
```diff
@@ -35,12 +35,12 @@ this form:
...
   }, ?
   "createdat": "TIMESTAMP",
   "modifiedat": "TIMESTAMP",
+  "modelversion": "1.0", ?
+  "compatiblewith": "https://raw.githubusercontent.com/xregistry/spec/refs/heads/main/schema/model.json", ?
```
duglin, line 43: per our chat last week, these can go away

core/spec.md
duglin, line 142: in all cases, let's move these 2 up before "labels"

message/spec.md
```diff
@@ -188,8 +186,8 @@ All message definitions MUST defined inside message groups.
 The formal xRegistry extension model of the Message Definitions Registry
 resides in the [model.json](model.json) file.
 
-By importing and keeping the `compatiblewith` label, interoperability on the
-CNCF defined endpoint model is stated.
+By importing and keeping the `compatiblewith` label and its value,
+interoperability on the CNCF defined endpoint model is stated.
```
duglin, line 190: same comment as above, I think we can remove this

schema/spec.md
```diff
@@ -161,8 +159,8 @@ version" and the **schema Versions** as "minor versions".
 The formal xRegistry extension model of the Schema Registry resides in the
 [model.json](model.json) file.
 
-By importing and keeping the `compatiblewith` label, interoperability on the
-CNCF defined endpoint model is stated.
+By importing and keeping the `compatiblewith` label and its value,
+interoperability on the CNCF defined endpoint model is stated.
```
duglin, line 163: same comment as above, I think we can remove this

message/model.json
duglin: yes!

core/spec.md
```diff
@@ -2295,6 +2297,8 @@ Regardless of how the model is retrieved, the overall format is as follows:
 
 ```yaml
 {
+  "modelversion": "STRING"             # Version of the model
+  "compatiblewith": "URI"`             # Statement of compatibility with model spec
```
duglin, line 2301: theses need ? since they're optional

Fannon:
@duglin - I think I'm done now, did some final cleanup and consistency things. 

Now it's down to the python error, my guess is that your scripts have problems with the two new attributes in the mode.json files?

core/spec.md
```diff
@@ -2353,6 +2361,25 @@ Regardless of how the model is retrieved, the overall format is as follows:
 ```
 
 The following describes the attributes of Registry model:
+
+- `modelversion` <span id="model.modelversion"></span>
+  - Type: String.
+  - OPTIONAL.
```
duglin, line 2367: let's remove the periods on these 2 and on the next attribute - just for consistency with the other attributes

schema/model.json
```diff
@@ -12,6 +14,8 @@
       },
 
       "resources": {
+        "modelversion": "0.5",
+        "compatiblewith": "https://raw.githubusercontent.com/xregistry/spec/refs/heads/main/schema/model.json",
```
duglin, line 18: these need to be under a resource (e.g. "schemas") not the "resources" map.
This will fix the validation error you're seeing

core/spec.md
Fannon: Will do. I added them because the previous location had the periods consistently, then I moved it. 

I noticed also some other differences between those two places, but I don't think anyone would notice if they don't compare 1:1

duglin: crap - add them back in - I just looked and you're right. I'll add then in other spots in the spec to be consistent

duglin: done

schema/model.json
Fannon: Fixed it, thanks!

Fannon:
@duglin - PR is green now :)

core/spec.md
```diff
@@ -2353,6 +2361,25 @@ Regardless of how the model is retrieved, the overall format is as follows:
...
+  - The version of the local model of a group or resource.
+  - It is common to use a combination of major and minor version numbers.
+  - Examples: `1.2`
```
duglin, line 2370: minor things but let's move `1.2` to a new sub-bullet like the next one has it

message/model.json
```diff
@@ -1,8 +1,11 @@
 {
+
```
duglin, line 2: other json files don't have this blank line - any reason you added it?

Fannon: No, it just came through rearranging, I'll remove it tomorrow.

core/spec.md
Fannon: Will do tomorrow

Fannon: I looked below and I think the consistent form is: 

- Example: `1.2`

Will do the same with compatiblewith.

duglin:
@Fannon I rebased and squashed it down to one commit - hope that's ok.
I had to resolve a few minor conflicts

duglin:
approved on the 3/27 call

duglin:
It's live:
image

Fannon:
Thanks @duglin  for helping me through it :)

PR #284, opened 2025-03-21
Message Review

clemensv:

tightening of various definitions in the messaging spec and aligning the model with the spec in several places

### DIFF
--- a/message/model.json
+++ b/message/model.json
@@ -9,16 +9,16 @@
         "envelope": {
           "name": "envelope",
           "type": "string",
-          "description": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted. The predefined envelope format is 'CloudEvents/1.0'",
+          "description": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted. The predefined envelope format is CloudEvents/1.0"
         },
         "protocol": {
           "name": "protocol",
           "type": "string",
-          "description": "Protocol identifier that defines the transport message protocol. All definitions in this group share this protocol type. Mixed-protocol groups are not permitted. Predefined protocol formats are 'AMQP/1.0', 'MQTT/3.1.1', 'MQTT/5.0', 'KAFKA', 'HTTP', and 'NATS'",
+          "description": "Protocol identifier that defines the transport message protocol. All definitions in this group share this protocol type. Mixed-protocol groups are not permitted. Predefined protocol formats are 'AMQP/1.0', 'MQTT/3.1.1', 'MQTT/5.0', 'KAFKA', 'HTTP', and 'NATS'"
         },
         "*": {
           "name": "*",
-          "type": "var",
+          "type": "any",
           "description": "Any other attribute that is common to all definitions in this group"
         }
       },
@@ -924,7 +924,7 @@
                             "ttl": {
                               "name": "ttl",
                               "description": "AMQP time-to-live (milliseconds)",
-                              "type": "number"
+                              "type": "integer"
                             },
                             "first-acquirer": {
                               "name": "first-acquirer",
--- a/cloudevents/schemas/document-schema.avsc
+++ b/cloudevents/schemas/document-schema.avsc
@@ -313,11 +313,6 @@
                       "type": [
                         {
                           "type": "record",
-                          "name": "EnvelopeNoneType",
-                          "fields": []
-                        },
-                        {
-                          "type": "record",
                           "name": "EnvelopeCloudevents10Type",
                           "fields": [
                             {
@@ -359,7 +354,7 @@
                                 ]
                               },
                               "name": "envelopemetadata",
-                              "doc": "CloudEvents metadata constraints"
+                              "doc": "CloudEvents attribute declarations, adding constraints to the CloudEvents envelope not yet covered by the CloudEvents spec https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md"
                             },
                             {
                               "type": "GenericRecord",
@@ -369,16 +364,11 @@
                           ]
                         }
                       ],
-                      "doc": "Message envelope format identifier. This attribute MUST be the same as the 'envelope' attribute of the definition group, but is not automatically inherited because definitions may be cross-referenced across groups"
+                      "doc": "Message envelope format identifier. This attribute MUST be the same as the 'envelope' attribute of the definition group, but is not automatically inherited because definitions may be cross-referenced across groups. The predefined envelope format is 'CloudEvents/1.0'"
                     },
                     {
                       "name": "protocol",
                       "type": [
-                        {
-                          "type": "record",
-                          "name": "ProtocolNoneType",
-                          "fields": []
-                        },
                         {
                           "type": "record",
                           "name": "ProtocolAmqp10Type",
@@ -419,7 +409,7 @@
                             {
                               "type": "GenericRecord",
                               "name": "protocoloptions",
-                              "doc": "The Apache Kafka message metadata constraints"
+                              "doc": "Apache Kafka message metadata constraints"
                             }
                           ]
                         },
@@ -431,6 +421,17 @@
                               "type": "GenericRecord",
                               "name": "protocoloptions",
                               "doc": "The HTTP message metadata constraints"
+                            }
+                          ]
+                        },
+                        {
+                          "type": "record",
+                          "name": "ProtocolNATSType",
+                          "fields": [
+                            {
+                              "type": "GenericRecord",
+                              "name": "protocoloptions",
+                              "doc": "NATS message metadata constraints"
                             }
                           ]
                         }
@@ -483,7 +484,7 @@
                     {
                       "type": "string",
                       "name": "dataschemauri",
-                      "doc": "The URI of the schema for the message payload, equivalent to the 'schemaurl' attribute of the schema registry"
+                      "doc": "The URI of the schema for the message payload, equivalent to the 'schemauri' attribute of the schema registry"
                     },
                     {
                       "type": "string",
@@ -589,12 +590,12 @@
             {
               "type": "string",
               "name": "envelope",
-              "doc": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted"
+              "doc": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted. The predefined envelope format is CloudEvents/1.0"
             },
             {
               "type": "string",
               "name": "protocol",
-              "doc": "Protocol identifier that defines the transport message protocol. All definitions in this group share this protocol type. Mixed-protocol groups are not permitted"
+              "doc": "Protocol identifier that defines the transport message protocol. All definitions in this group share this protocol type. Mixed-protocol groups are not permitted. Predefined protocol formats are 'AMQP/1.0', 'MQTT/3.1.1', 'MQTT/5.0', 'KAFKA', 'HTTP', and 'NATS'"
             },
             {
               "name": "Extensions",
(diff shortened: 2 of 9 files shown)


duglin:
You can add the new words to `tools/dict`

message/model.json
```diff
@@ -228,19 +275,48 @@
...
+                          "attributes": {
+                            "description": {
+                              "name": "description",
+                              "type": "string"
+                            },
+                            "type": {
+                              "name": "type",
+                              "type": "string",
+                              "default": "string",
+                              "required": false
```
duglin, line 308: just curious, is there a reason to have: required:false  in some but not others?

duglin:
If you use the `xr` exe from the ref impl:
```
$ xr model verify https://raw.githubusercontent.com/xregistry/spec/1965907af1a2eadb1ebae4259187d4d35972d98d/message/model.json
```
you'll see if it adheres to the spec. Right now I get this:
```
$ xr model verify https://raw.githubusercontent.com/xregistry/spec/1965907af1a2eadb1ebae4259187d4d35972d98d/message/model.json
https://raw.githubusercontent.com/xregistry/spec/1965907af1a2eadb1ebae4259187d4d35972d98d/message/model.json: unknown field "description"
```

I'll look into seeing if we can get some indication of where it is in the json, but it might be awhile.  Golang doesn't help much with that.

duglin:
After a rebase we need to make sure all of the core-level attributes appear in the pseudo-schema snippets - especially the recently added ones.

message/spec.md
```diff
@@ -176,13 +192,38 @@ describe constraints for the metadata of messages and events, for instance the
...
+and which which data types, values, or patterns are expected to produce a
+conformant message. A message consumer can use the definitions to validate
+incoming messages and to extract the metadata for routing or processing.
+
 A message group is a collection of message definitions that are related to
 each other in some application-specific way. For instance, a message group
 can be used to group all events raised by a particular application module or by
 a particular role of an application protocol exchange pattern.
 
 All message definitions MUST defined inside message groups.
```
duglin, line 206: missing "be"

... MUST **be** defined inside...

message/spec.md
```diff
@@ -176,13 +192,38 @@ describe constraints for the metadata of messages and events, for instance the
...
 
+A message processor for a messaging or eventing channel can use a message group
+and its contained message definitions to match incoming messages to the declared
+message definitions and determine whether an incoming message is conformant to
+the expected metadata constraints. If a conformant message has been identified,
```
lailabougria, line 211:
```suggestion
the expected metadata constraints. If a conforming message has been identified,
```

message/spec.md
```diff
@@ -176,13 +192,38 @@ describe constraints for the metadata of messages and events, for instance the
...
+the processor might then use the linked schema to handle the message body. This
```
lailabougria, line 212: Should this be MAY?

message/spec.md
```diff
@@ -176,13 +192,38 @@ describe constraints for the metadata of messages and events, for instance the
...
+is especially useful in scenarios where the message itself does not contain a
+schema hint or even content type information as it is the case, for instance, in
+MQTT 3.1.1.
+
+Whether a message is conformant to a message definition is determined by the
```
lailabougria, line 217:
```suggestion
Whether a message conforms to a message definition is determined by the
```

message/spec.md
```diff
@@ -176,13 +192,38 @@ describe constraints for the metadata of messages and events, for instance the
...
+message processor and its implementation specifici rules. Conformance rules are
```
lailabougria, line 218:
```suggestion
message processor and its implementation-specific rules. Conformance rules are
```

message/spec.md
```diff
@@ -870,17 +913,22 @@ following properties are defined, with type constraints:
 | `group-sequence`       | `integer`     | position of this message within its group                                        |
 | `reply-to-group-id`    | `uritemplate` | group-id to which the receiver of this message ought to send replies to          |
 
+The `message-id` permits the types `ulong`, `uuid`, `binary`, `string`, and
+`uritemplate`. A `value` constraint for `message-id` property SHOULD NOT be defined in
+the message definition except for the case where the message-id is a `uritemplate`.
```
lailabougria, line 918:
```suggestion
the message definition except for the case where the `message-id` is a `uritemplate`.
```

lailabougria (commented):
The "required" attribute is false by default, so it could be omitted everywhere. Also, if the attribute is not required, shouldn't the default also be removed everywhere?

duglin:
- "envelopeoptions" map needs to be tagged as optional with "?" 

- the following line needs to be modified since they're not labels any more:
  - By importing and keeping the compatiblewith label, interoperability on the CNCF defined endpoint model is stated.

- s/basic attributes/xRegistry-defined core attributes/

- when {VERSION} is optional, do we need to say it defaults to the latest version?

- s/about likely alternative message/about likely alternative messages/   in the "deprecated" section

- basemessagurl should probably say something more about how the attributes are merged. For example, do local attributes fully replace the basemessageurl attributes or is there some kind of merge that goes on?

- datacontenttype has a typo:
  - `Type: Type: String per [RFC 2046](https://tools.ietf.org/html/rfc2046)` - double "Type"

- alphabetize the "common properties" section.

- `Placeholders that are used multiple times in the same message definition are assumed to represent identical values.` - should it be "MUST" instead of "assumed" ?

- for "type", move `Default value MUST be "string".` to its own sub-bullet.

- for "type", s/var/any/

- for "type" should we reference the xreg spec type system instead of CE's - where possible? While we all love CE, for someone who doesn't need it, it seems more appropriate to lean on xReg instead of CE.

- `The dataschema attribute's value is inferred from the [dataschemauri](https://github.com/xregistry/spec/blob/1965907af1a2eadb1ebae4259187d4d35972d98d/message/spec.md#dataschemauri) attribute or [dataschema](https://github.com/xregistry/spec/blob/1965907af1a2eadb1ebae4259187d4d35972d98d/message/spec.md#dataschema) attribute of the message definition if absent.` - how can CE's "dataschema", which is a URI, align with "dataschema" of the message which is an inline schema?

- this look odd to have `required: BOOLEAN`. I'm not sure unknown extensions can be required:
```
  "*": {
    "value": ANY, ?
    "type": "TYPE", ?
    "required": BOOLEAN ?
  }
```
in fact, the xreg spec says: `When the attribute name is * then required MUST NOT be set to true.`

- what does ...`implied dataschema is https://example.com/schemas/com.example.myevent.json` mean? Is "implied" the right word? oh wait, I think my confusion would be removed if it said "the CE 'dataschema'" so people don't think you're talking about the msg spec's "dataschema" attribute.

- For consistency, things like "The http ..." should be "The HTTP". e.g. see table in: https://github.com/xregistry/spec/blob/1965907af1a2eadb1ebae4259187d4d35972d98d/message/spec.md#message-protocols

- https://github.com/xregistry/spec/blob/1965907af1a2eadb1ebae4259187d4d35972d98d/message/spec.md#properties-amqp-10   has "*" as a type , to make it clear that people need to see the text after the table perhaps change it to "see note below" ?

-

duglin:
Not all attributes (e.g. "envelope" have OPTIONAL vs REQUIRED in their "Constraints" sections. We should add them even if some text (in other specs) hints at it. Then people have a consistent way of find out that aspect.

message/model.json
```diff
@@ -3,20 +3,26 @@
...
           "type": "string",
-          "description": "Protocol identifier that defines the transport message protocol. All definitions in this group share this protocol type. Mixed-protocol groups are not permitted"
+          "description": "Protocol identifier that defines the transport message protocol. All definitions in this group share this protocol type. Mixed-protocol groups are not permitted. Predefined protocol formats are 'None', 'AMQP/1.0', 'MQTT/3.1.1', 'MQTT/5.0', 'KAFKA', 'HTTP', and 'NATS'",
+          "default": "None",
+          "required": false
         },
         "*": {
           "name": "*",
-          "type": "any"
+          "type": "var",
```
duglin, line 24: just a reminder to use "any"

message/model.json
```diff
@@ -35,126 +41,155 @@
             "envelope": {
               "name": "envelope",
               "type": "string",
-              "description": "Message envelope format identifier. This attribute MUST be the same as the 'envelope' attribute of the definition group, but is not automatically inherited because definitions may be cross-referenced across groups",
+              "description": "Message envelope format identifier. This attribute MUST be the same as the 'envelope' attribute of the definition group, but is not automatically inherited because definitions may be cross-referenced across groups. Predefined envelope formats are 'None' and 'CloudEvents/1.0'",
               "ifvalues": {
                 "None": {},
```
duglin, line 46: once we remove "none"  (see comment above) we can remove these too

message/model.json
```diff
@@ -35,126 +41,155 @@
...
                           "name": "specversion",
-                          "description": "CloudEvents spec version",
+                          "description": "CloudEvents specversion https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#specversion",
                           "type": "object",
                           "attributes": {
                             "description": {
                               "name": "description",
-                              "type": "string"
+                              "type": "string",
+                              "required": false
```
duglin, line 62: I think it would be better to remove the "required:false" lines to reduce the noise and to not imply that w/o it the default is "required:true".

message/model.json
```diff
@@ -3,20 +3,26 @@
     "messagegroups": {
       "singular": "messagegroup",
       "plural": "messagegroups",
+      "description": "A group of message definitions that are semantically related, such as a set of messages that are part of the same event flow or that share a common protocol or envelope format. One or both of the envelope and protocol definitions must be set to a value other than `None`.",
```
duglin, line 6: we actually don't have a "description" field here in our meta-model. I've open an issue to discuss adding it

message/model.json
```diff
@@ -1229,12 +1346,68 @@
...
+                      }
+                    }
+                  }
+                },
+                "NATS": {
+                  "siblingattributes": {
+                    "protocoloptions": {
+                      "name": "protocoloptions",
+                      "description": "NATS message metadata constraints",
+                      "type": "object",
```
duglin, line 1367: need: charset:extended

message/model.json
```diff
@@ -720,88 +953,123 @@
...
-                              },
-                              "required": {
-                                "name": "required",
-                                "description": "AMQP header required",
-                                "type": "boolean",
-                                "required": true,
-                                "default": false
-                              }
+                          "type": "object",
+                          "description": "AMQP header section https://docs.oasis-open.org/amqp/core/v1.0/os/amqp-core-messaging-v1.0-os.html#type-header",
```
duglin, line 998: need: namecharset:extended

message/spec.md
clemensv: conformant (adjective)
(especially of technology) compatible or conforming with appropriate standards:
"vendors are using the test tool and expect to announce conformant products soon" · "using XSLT will guarantee that your output will be conformant to rules for generating XML" · "our text encoding policy is now fully conformant with the guidelines"

message/model.json
clemensv: There are no fields in NATS that require it

duglin: "reply-to"

message/model.json
clemensv: I'm removing them.

message/model.json
```diff
@@ -3,20 +3,26 @@
...
       "singular": "messagegroup",
       "plural": "messagegroups",
+      "description": "A group of message definitions that are semantically related, such as a set of messages that are part of the same event flow or that share a common protocol or envelope format. One or both of the envelope and protocol definitions must be set to a value other than `None`.",
       "attributes": {
         "envelope": {
           "name": "envelope",
           "type": "string",
-          "description": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted"
+          "description": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted. Predefined envelope formats are 'None' and 'CloudEvents/1.0'",
+          "default": "None",
```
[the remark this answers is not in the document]
clemensv: taking this out (for now)

duglin:
`xr model verify http...` will help in your reviews. I just updated it with the latest stuff... like "description" on group and resource model types

duglin:
or just `xr model verify FILE`

PR #286, opened 2025-03-21
Endpoint review

clemensv:

Tweaks for the endpoints model

### DIFF
--- a/message/model.json
+++ b/message/model.json
@@ -270,8 +270,8 @@
                           }
                         },
                         "datacontenttype": {
-                          "name": "CloudEvents datacontenttype https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#datacontenttype.",
-                          "description": "The content type of the data attribute value. ",
+                          "name": "datacontenttype",
+                          "description": "CloudEvents datacontenttype https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#datacontenttype.",
                           "type": "object",
                           "attributes": {
                             "description": {
@@ -1240,6 +1240,7 @@
                       "name": "protocoloptions",
                       "description": "NATS message metadata constraints",
                       "type": "object",
+                      "namecharset": "extended",
                       "attributes": {
                         "subject": {
                           "name": "subject",


endpoint/model.json
```diff
@@ -199,9 +201,25 @@
                         "copy"
                       ],
                       "description": "The AMQP distribution mode for receivers. Can be 'move' or 'copy'.  A value of 'move' indicates an exclusive lock on the message. A value of 'copy' indicates a non-exclusive lock on the message",
-                      "required": true,
+                      "required": false,
                       "default": "move"
                     },
+                    "connection-capabilities": {
+                      "name": "capabilities",
```
duglin, line 208: bad name

endpoint/model.json
```diff
@@ -225,10 +243,10 @@
                         "type": "object",
                         "attributes": {
                           "uri": {
-                            "name": "uri",
-                            "type": "uri",
-                            "description": "Network accessible location",
-                            "required": true
+                            "name": "url",
```
duglin, line 246: name doesn't match key (uri)

endpoint/model.json
```diff
@@ -279,45 +297,52 @@
                       "name": "deployed",
                       "type": "boolean",
                       "description": "If `true`, the endpoint metadata represents a public, live endpoint that is available for communication and a strict validator MAY test the liveness of the endpoint",
-                      "required": true,
+                      "required": false,
                       "default": false
                     },
-
                     "topic": {
                       "name": "topic",
```
duglin, line 304: I wonder if we should allow "name" to be optional on requests but then require the server to fill it in from the key? Just to make life easier for people. We do something similar for IDs already.

duglin: https://github.com/xregistry/spec/issues/304

endpoint/model.json
```diff
@@ -225,10 +243,10 @@
                         "type": "object",
                         "attributes": {
                           "uri": {
-                            "name": "uri",
-                            "type": "uri",
-                            "description": "Network accessible location",
-                            "required": true
+                            "name": "url",
+                            "type": "url",
```
duglin, line 247: "type" should match the name too

duglin:
for rc1

Issue #304, opened 2025-03-30
Should we allow "name" to be optional for requests in the model but then require server to fill it in from the "key"?


duglin:
Just to make life easier for people. We do this for xReg entity IDs already

duglin:
Approved 4/8

Issue #303, opened 2025-03-29
when a default value is provided, should "required" be "true"?


duglin:
As a reminder, we require attributes with a default value to always be serialized so the clients know what value was used. And because the default value might change over time, but existing instances of that attribute MUST NOT change their value as a result. Basically, make it "implicitly required".

duglin:
Note, this would change a lot of the samples that we have

duglin:
Approved 4/8

PR #315, opened 2025-04-03
Week of Apr7

duglin:
- Add some samples to the 'core' dir
- Run tooling on more md/json files
- make it clear that ancestor ordering is case-insensitive w.r.t vIDs
- while "strongly recommended", it's no required to detect circular references
  w.r.t. version's ancestor attribute
- make it clear the changing 'ancestor' means epoch/modifiedat are also changed
- in model, an attr with a default value MUST also be tagged as 'required'
  - fix some models
- add text to primer to make it clear the not all required fields are
  required in all requests - e.g. map keys->IDs
- point to xregistry.io static file hosting of our model.json docs for
  the "compatiblewith" attributes

Fixes: #320
Fixes: #319 
Fixes: #304 
Fixes: #303

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -756,3 +756,60 @@ the data (i.e. it was exported to the file-server), and there's a desire to
 not expose this implementation/deployment choice to the end-user. In other
 words, the use of the file-server as a caching service should be transparent
 to their users.
+
+# Detecting circular references between Versions
+
+The specification mentions that the Version's `ancestor` attribute should not
+be set in such a way as to cause a circular dependency list between Versions.
+However, when talking about the server generating an error in such conditions
+it says it is "STRONGLY RECOMMENDED" that an error is generated, not that it
+"MUST" be generated. This is not meant to be used by implementations as a way
+to avoid doing the check in general. However, it is recognized that there
+might be situations (such as when the number of Versions is very large) that
+performing this check upon every change could become a performance concern.
+
+Rather than mandating a potentially burdensome requirement on implementations,
+the specification allows for this check to be skipped if there is no
+reasonable way to do it. However, if an implementation chooses not to perform
+the check then it becomes incumbent upon them to ensure that there are no
+negative ramifications. For example, when performing backwards compatibility
+checking, they may need to add extra logic to ensure they don't go caught in
+an infinite loop. Additionally, implementations should also document that they
+do not enforce this check to set the proper expectation levels for their users.
+
+# Optional required fields in requests
+
+One of the design principles of the specification was to try to make each
+object be fully self-describing, within reason. For example, this is why when
+objects appear within maps the map key is often replicated within the object
+itself. For example, when looking at a Versions collection it might look like
+this:
+
+```yaml
+"version": {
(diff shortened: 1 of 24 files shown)


core/primer.md
```diff
@@ -756,3 +756,59 @@ the data (i.e. it was exported to the file-server), and there's a desire to
 not expose this implementation/deployment choice to the end-user. In other
 words, the use of the file-server as a caching service should be transparent
 to their users.
+
+# Detecting circular references between Versions
+
+The specification mentions that the Version's `ancestor` attribute should not
+be set in such a way as to cause a circular dependency list between Versions.
+However, when talking about the server generating an error in such conditions
+it says it is "STRONGLY RECOMMENDED" that an error is generate, not that it
```
jskeet, line 765:
```suggestion
it says it is "STRONGLY RECOMMENDED" that an error is generated, not that it
```

core/primer.md
```diff
@@ -756,3 +756,59 @@ the data (i.e. it was exported to the file-server), and there's a desire to
...
+"MUST" be generated. This is not meant to be used by implementations as a way
+to avoid doing the check in general. However, it is recognized that there
+might be situations (such as when the number of Versions is very large) that
+performing this check upon every change could become a performance concern.
+
+Rather than mandating a potentially burdensome requirement on implementations,
+the specification allows for this check to be skipped if there is no
+reasonable way to do it. However, if an implementation chooses to not perform
```
jskeet, line 773:
```suggestion
reasonable way to do it. However, if an implementation chooses not to perform
```

core/primer.md
```diff
@@ -756,3 +756,59 @@ the data (i.e. it was exported to the file-server), and there's a desire to
...
+the check then it becomes incumbent upon them to ensure that there are no
+negative ramifications. For example, when performing backwards compatibility
+checking, they may need to add extra logic to ensure they don't go caught in
+an infinite loop. Additionally, Implementation should also document that they
```
jskeet, line 777:
```suggestion
an infinite loop. Additionally, implementations should also document that they
```

core/primer.md
```diff
@@ -756,3 +756,59 @@ the data (i.e. it was exported to the file-server), and there's a desire to
...
+However, when those attributes are defined as "required", that does not
+necessarily mean that they are required to appear on all requests if the
+server can determine the proper value via some other means, such as the owning
+map key, or if the request URL includes the information.
+
+From specification perspective, all that's required is that after any
+operation all required fields have valid values. Whether it is provided by
+the client or calculated/inferred by the server is an implementation detail.
+
+However, this just refers to incoming messages, all server generated messages
```
jskeet, line 812:
```suggestion
This accommodation only applies to incoming messages. All server generated messages
```

jskeet (approved):
Just a few suggestions, but basically fine.

core/primer.md
duglin: fixed

core/primer.md
duglin: done

Issue #321, opened 2025-04-07
check domain spec Resource aspects


duglin:
e.g. some have `"setdefaultversionsticky": false` and I'm not sure why.

We should check all apsects.

```
GROUP: endpoints / endpoint
  ATTRIBUTES:   TYPE   REQ RO EDIT DEFAULT
  channel       string n   n  y    
  deprecated    object n   n  y    
  envelope      string n   n  y    
  messagegroups array  n   n  y    
  protocol      string n   n  y    
  usage         string n   n  y    
  *             any    n   n  y    

  RESOURCE: messages/ message
    Max versions      : 1      XXXXX
    Set version id    : true
    Set version sticky: false    XXXXXX
    Has document      : false
    
    ATTRIBUTES:      TYPE   REQ RO EDIT DEFAULT
    basemessageurl   uri    n   n  y    
    datacontenttype  string n   n  y    
    dataschema       any    n   n  y    
    dataschemaformat string n   n  y    
    dataschemauri    uri    n   n  y    
    envelope         string n   n  y    
    protocol         string n   n  y    

GROUP: messagegroups / messagegroup
  ATTRIBUTES: TYPE   REQ RO EDIT DEFAULT
  envelope    string n   n  y    
  protocol    string n   n  y    
  *           any    n   n  y    

  RESOURCE: messages/ message
    Max versions      : 1           XXXXXX
    Set version id    : true
    Set version sticky: false     XXXXX
    Has document      : false
    
    ATTRIBUTES:      TYPE   REQ RO EDIT DEFAULT
    basemessageurl   uri    n   n  y    
    datacontenttype  string n   n  y    
    dataschema       any    n   n  y    
    dataschemaformat string n   n  y    
    dataschemauri    uri    n   n  y    
    envelope         string n   n  y    
    protocol         string n   n  y    

GROUP: schemagroups / schemagroup
  ATTRIBUTES: TYPE REQ RO EDIT DEFAULT
  *           any  n   n  y    

  RESOURCE: schemas/ schema
    Max versions      : 0
    Set version id    : true
    Set version sticky: true
    Has document      : true
    
    ATTRIBUTES: TYPE   REQ RO EDIT DEFAULT
    format      string n   n  y    
    *           any    n   n  y    
    
    META ATTRIBUTES: TYPE    REQ RO EDIT DEFAULT
    validation       boolean y   n  y    true     XXXXX
```

duglin:
I made "validation" default to "false" but it's still "required" as part of PR #322

Issue #318, opened 2025-04-04
make $schema officially allowed in incoming JSON snippets


duglin:
Is it as simple as just saying something like:

- - -
Any xRegistry JSON metadata message that represents a single entity (i.e. not a map) MAY include a top-level "$schema" attribute that points to a JsonSchema document that describes the message contents. These notations can be used or ignored by receivers of these messages. There is no requirement for implementation of this specification to persist these values, to include them in responses or to use this information.
- - -

btw - support for allowing "$schema" to appear (even to ignore it) could be costly. For example, a receiver may not be able to just stream its processing without reading the entire json into memory so that it can remove the "$schema" attribute before it passes the json along to a deserializer that might then look for a "$schema" attribute in the target data structure. 

While I understand the tooling value of this information, it doesn't seem cost-free to support in practice for non-schema-aware code.

duglin:
AI: Add a pointer to the xregistry.io/.... location for the various schema docs we generate.

Issue #307, opened 2025-04-01
Do we need a new "export" type of API?


duglin:
In https://github.com/xregistry/spec/pull/285 @clemensv has some sample Groups/Resources that can be added to an existing Registry via a `POST /`. However, there's no way to generate that same type of output from a Registry with our APIs. There's no way to ask for just the children collections w/o also getting the parent's attributes (the Registry itself in this case).

Do we need a new API to make it easier to get just the children collection for import/export usecases?

I have a similar question for exporting Resources from Groups too.

duglin:
Viewed as "I want just the children collections":
```
GET /?collections
GET /schemagroups/mygroup?children
```

Viewed as "I want to exclude the request root's attributes":
```
GET /?noroot
GET /schemagroups/mygroup?noattributes
GET /schemagroups/mygroup?noattrs
```

In both cases the user can choose which collections, or entries in the collections, are returned via `?filter`. Note that the "excluded" collection will still appear, just with no entries in them.

First ones sound a bit too much like asking for "inline", which could be confusing.
I'm leaning towards `?noroot`.

Not supported on .../rID because people can ask for the verions via `GET .../rID/versions`, no need for a special/new API. Note that `POST .../rID` is used to get ONE new version, not a collection of versions.

duglin:
approved - use "collections"

Issue #302, opened 2025-03-29
should we move "resource.deprecated" to core?


duglin:
It's already in endpoint, message and schema.

```
          "deprecated": {
            "effective": "TIMESTAMP", ?
            "removal": "TIMESTAMP", ?
            "alternative": "URL", ?
            "docs": "URL"?
          }, ?
```

duglin:
Approved on 4/15

PR #322, opened 2025-04-10
Week of Apr 14 changes

duglin:
Week of Apr 14 changes

- minor wording fixes
- move "deprecated" into core spec. Removed from message and schema. Endpoint points to core, on endpoint/group.
- treat a missing 'ancestor' on PUT/POST as a request to leave it unchanged
- allow for `$schema` to appear in request xReg metadata messages
- add a `?collection` flag for GET / and GET /GROUPS/gID (turns on `?inline=*`)
- remove unnecessary stuff from our model and samples - to reduce noise

Fixes: #297
Fixes: #302
Fixes: #307
Fixes: #318
Fixes: #321
Fixes: #325

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -813,3 +813,18 @@ This accommodation only applies to incoming messages. All server generated
 messages (aside from when special flags, like the `?doc` flag, is enabled)
 must include all required attributes, even if they are duplicated elsewhere
 in the message.
+
+# Deprecation of entities in an xRegistry
+
+The core specification defines a `deprecated` attribute that may appear
+under a Resource's `meta` sub-object. This attribute was added to the Resource
+itself rather than to the Version because it was determined that the most
+likely usage of this feature is to express the intent to deprecate the entire
+Resource rather than just one Version (or subset of Versions). This is not say
+that the use of this feature might not be useful at the Version-level or even
+at the Group-level. However, for those cases, custom models may define
+an extension at the appropriate location in the model to meet their needs.
+When doing so it is recommended to use the same attribute definition as
+defined in the core specification for consistency. It is worth noting that
+the Endpoint [specification](../endpoint/spec.md) does exactly this to
+indicate when an Endpoint (i.e. a Group) is deprecated.
--- a/core/spec.md
+++ b/core/spec.md
@@ -42,9 +42,10 @@ automation and tooling.
     - [Retrieving a Version](#retrieving-a-version)
     - [Deleting Versions](#deleting-versions)
   - [Configuring Responses](#configuring-responses)
-    - [Inline](#inline)
-    - [Filter](#filter)
-    - [Doc](#doc)
+    - [Collections Flag](#collections-flag)
+    - [Doc Flag](#doc-flag)
+    - [Filter Flag](#filter-flag)
+    - [Inline Flag](#inline-flag)
   - [HTTP Header Values](#http-header-values)
   - [Error Processing](#error-processing)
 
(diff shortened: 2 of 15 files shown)


core/primer.md
```diff
@@ -813,3 +813,18 @@ This accommodation only applies to incoming messages. All server generated
...
+
+The core specification defines a `deprecated` attribute that may appear
+under a Resource's `meta` sub-object. This attribute was added to the Resource
+itself rather than to the Version because it was determined that the most
+likely usage of this feature is to express the intent to deprecate the entire
+Resource rather than just one Version (or subset of Versions). This is not say
+that the use of this feature might not be useful at the Version-level or even
+at the Group-level. However, for those cases, custom models may simply define
+an extension at the appropriate location in the model to meet their needs.
+And when doing so it is recommended to define it to look the same as the one
```
jskeet, line 827:
```suggestion
When doing so it is recommended to define it to look the same as the one
```

(I also wonder about "look the same", but I don't have any suggestions for that yet. Would "use the same schema" work?)

core/primer.md
```diff
@@ -813,3 +813,18 @@ This accommodation only applies to incoming messages. All server generated
...
+defined in the core specification. It is worth nothing the the Endpoint
```
jskeet, line 828:
```suggestion
defined in the core specification. It is worth noting the the Endpoint
```

core/spec.md
```diff
@@ -1028,6 +1036,13 @@ aspects of the processing:
 - `?noreadonly` - presence of this query parameter indicates that any attempt
   to update a read-only Resource MUST be silently ignored.
 
+Any JSON xRegistry metadata message that represents a single entity (i.e. not
+a map) MAY include a top-level "$schema" attribute that points to a JsonSchema
```
jskeet, line 1040:
```suggestion
a map) MAY include a top-level "$schema" attribute that points to a JSON Schema
```

(unless I've missed something)

core/spec.md
```diff
@@ -4168,6 +4184,50 @@ and the following Resource level attributes:
...
+
+  Note that an implementation is not mandated to use this attribute in
+  advance of removing a Resource, but is it RECOMMENDED that they do so.
+- Constraints:
+  - OPTIONAL
+- Examples:
+  - `"deprecated": {}`
+  - ```
+    "deprecated": {
+      "removal": "2030-12-19T00:00:00-00:00",
```
jskeet, line 4226: Eek, let's not use an offset of -00:00 - that indicates an "unknown local offset"; that's unlikely to be usefully represented in most software, and I doubt that many users know what it means anyway. I would strongly recommend using Z here (and everywhere).

core/spec.md
```diff
@@ -6143,107 +6211,121 @@ those cases, the client will need to query the individual inlinable attributes
...
-of the Resource, specifying `RESOURCE` in the `?inline` query parameter MUST
-be interpreted as a request for the appropriate attribute. In other words,
-`RESOURCEbase64` is not a valid inlinable attribute name.
+Note that this feature only applies to the root entity of the response and not
+to any nested entities/collections.
 
-Use of this feature is useful for cases where the contents of the Registry are
-to be represented as a single (self-contained) document.
+This feature is meant to be used when the Collections of the Registry, or
+Group, are of interest but not the top-level metadata. For example, to export
```
jskeet, line 6226:
```suggestion
Group, are of interest but not the top-level metadata. For example, this could be used to export
```

(Or something similar - otherwise it doesn't sound like a complete sentence to me.)

jskeet (approved):
Just some niggles.

core/primer.md
duglin: Done and I tweaked it to say:
When doing so it is recommended to use the same attribute definition as
defined in the core specification for consistency. 

hope that reads ok.

core/primer.md
duglin: done, and s/the the/that the/

core/spec.md
duglin: sounds right

core/spec.md
duglin: Fixed here and in endpoint's spec.
I'll work on add the "Z" in next week's PR as that's a bit of a bigger change.

core/spec.md
duglin: done

duglin:
@jskeet thank for the review.

Also, for anyone else...I just added text to say that the related "...count" and "...url" collection attributes MAY be included when ?collections is used - server's choice. And the ?collections implied ?inline=* - otherwise it's kind of silly.

core/primer.md
jskeet: Works for me :)

duglin:
Approved on the 4/17 call

PR #328, opened 2025-04-22
Removing erroneous required clauses. Fix schema generator to allow for ifvalues fields to be optional

clemensv:

## Proposed Changes

This is a fix for the models, which have some incorrect (out of spec) required fields. It also fixes the schema generator to allow for optional attributes with ifvalues.

This is an errata fix for RC1.

### DIFF
--- a/schema/schemas/openapi.json
+++ b/schema/schemas/openapi.json
@@ -1615,7 +1615,7 @@
         "schema": {
           "type": "string"
         },
-        "description": "Comma\u2011separated list of nested paths to inline (e.g. endpoints, endpoints.messages). See spec: Inline."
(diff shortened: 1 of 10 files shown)


schema/schemas/openapi.json
```diff
@@ -1615,7 +1615,7 @@
         "schema": {
           "type": "string"
         },
-        "description": "Comma\u2011separated list of nested paths to inline (e.g. endpoints, endpoints.messages). See spec: Inline."
+        "description": "Comma\u00e2\u20ac\u2018separated list of nested paths to inline (e.g. endpoints, endpoints.messages). See spec: Inline."
```
duglin, line 1618: why all of these special chars?

PR #330, opened 2025-04-22
Week of Apr 21

### LINKED ISSUE #326: Add "Z" to our tz-less timestamps per Jon's suggestion

### LINKED ISSUE #327: Which non-model APIs are supported?
- - -
Support for any particular API defined by this specification is OPTIONAL, however, it is STRONGLY RECOMMENDED that server-side implementations support at least the "read" (e.g. HTTP GET) operations.
- - -

I think it would be good to say that /model and /capabilities are STRONG RECOMMENDED in all cases to help with discovery and interop.

Add to capabilities:
```
"apis": [ "/model", "/capabilities", "/export" ]
```
Indicates which non-model APIs are supported.

duglin:
- add "Z" to timestamps
- add `apis` capability

Fixes #326
Fixes #327

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -118,6 +118,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
   "modifiedat": "TIMESTAMP",
 
   "capabilities": {                     # Supported capabilities/options
+    "apis": [ "/capabilities", "/export", "/model" ],
     "flags": [                          # Query parameters
       "collections",? "doc",? "epoch",? "filter",? "inline",?
       "nodefaultversionid",? "nodefaultversionsticky",? "noepoch",?
@@ -1907,6 +1908,7 @@ be of the form:
 
 ```
 {
+  "apis": [ "STRING" ],
   "flags": [ "STRING" * ], ?
   "mutable": [ "STRING" * ], ?
   "pagination": BOOLEAN, ?
@@ -1940,6 +1942,32 @@ Absence, presence, or configuration values of a feature in the map MAY vary
 based on the authorization level of the client making the request.
 
 The following defines the specification-defined capabilities:
+
+#### `apis`
+- Name: `apis`
+- Type: Array of strings
+- Description: The list of APIs (beyond the APIs for the data model) that
+  are supported for read (`HTTP GET`) operations. This list is meant to allow
+  for clients/tooling to easily discover which of the APIs, that are not
+  related to the data model, are supported. Whether any of the APIs listed
+  below are supported for write operations can be discovered via the
+  `mutable` capability.
+- Note that it is allowable for the data that is available via more than one
+  mechanism to not be available via all mechanisms. For example, it is
+  possible for an implementation to support `GET /model` but not
+  `GET /?inline=model`.
+- Defined values:
+  - `/capabilities`
+  - `/export`
+  - `/model`
+- Values MUST start with `/`.
+- When not specified, the default value MUST be an empty list and no APIs
+  beyond ones for data model are supported.
+- Implementations MAY define their own values but they MUST NOT conflict with
+  specification defined APIs, Registry-level attributes or Group collection
+  attribute names.
+- It is STRONGLY RECOMMENDED that implementations support at least
+  `/capabilities` and `/model`.
 
 #### `flags`
 - Name: `flags`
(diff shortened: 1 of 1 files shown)


core/spec.md
```diff
@@ -1941,6 +1943,32 @@ based on the authorization level of the client making the request.
...
+  mechanism to not be available via all mechanisms. For example, it is
+  possible for an implementation to support `GET /model` but not
+  `GET /?inline=model`.
+- Defined values:
+  - `/capabilities`
+  - `/export`
+  - `/model`
+- Values MUST start with `/`.
+- When not specified, the default value MUST be an empty list and no APIs
+  beyond ones for data model are supported.
```
lailabougria, line 1965:
```suggestion
  beyond ones for the data model are supported.
```

core/spec.md
```diff
@@ -1941,6 +1943,32 @@ based on the authorization level of the client making the request.
...
+- Implementations MAY define their own values but they MUST NOT conflict with
+  specification defined APIs, Registry-level attributes or Group collection
```
lailabougria, line 1967:
```suggestion
  specification-defined APIs, Registry-level attributes, or Group collection
```

core/spec.md
```diff
@@ -1941,6 +1943,32 @@ based on the authorization level of the client making the request.
...
+  `mutable` capability.
+- Note that it is allowable for the data that is available via more than one
+  mechanism to not be available via all mechanisms. For example, it is
+  possible for an implementation to support `GET /model` but not
+  `GET /?inline=model`.
+- Defined values:
+  - `/capabilities`
+  - `/export`
+  - `/model`
+- Values MUST start with `/`.
```
lailabougria, line 1963: What's the thought process behind requiring the slash?

core/spec.md
```diff
@@ -1907,6 +1908,7 @@ be of the form:
 
 ```
 {
+  "apis": [ "STRING" ],
```
lailabougria, line 1911: Does this need a * given that it's extensible?

core/spec.md
duglin: done

core/spec.md
duglin: done

core/spec.md
duglin: Mainly because it makes it clear that these are top-level APIs, and not available under any other path/resource.
If you think it's confusing I think it's safe to remove it.

core/spec.md
```diff
@@ -1941,6 +1946,32 @@ based on the authorization level of the client making the request.
...
+  possible for an implementation to support `GET /model` but not
+  `GET /?inline=model`.
+- Defined values:
+  - `/capabilities`
+  - `/export`
+  - `/model`
+- Values MUST start with `/`.
+- When not specified, the default value MUST be an empty list and no APIs
+  beyond ones for the data model are supported.
+- Implementations MAY define their own values but they MUST NOT conflict with
```
jskeet, line 1969: It's worth thinking about what this means for spec evolution. If you *ever* add another API to the spec, that might mean that existing implementations could be broken if they already have an API with the same name.

core/spec.md
```diff
@@ -4321,6 +4371,31 @@ and the following Resource level attributes:
...
+  - REQUIRED.
+  - If the `meta` sub-object is inlined in the document then this attribute
+    MUST be a relative URL of the form `#JSON-POINTER` where the `JSON-POINTER`
+    locates the `meta` sub-object within the current document. See
+    [Doc Flag](#doc-flag) for more information.
+  - If the `meta` sub-object is not inlined in the document then this attribute
+    MUST be an absolute URL per the API view constraints listed above.
+
+- Examples:
+  - `https://example.com/endpoints/ep1/messages/msg1/meta`
```
jskeet, line 4397: Possibly add an example with a `#JSON-POINTER`?

jskeet (approved):
A couple of things to think about, but generally fine.

Yay for adding the Z - should we add something to the description of the relevant type, to encourage the use of Z?

core/spec.md
duglin: yup - same with extensions. It's a risk both sides take.

core/spec.md
duglin: done

duglin:
Approved on the 4/24 call

PR #331, opened 2025-04-23
add model verification to checker

duglin:


duglin:
merging so we can get the checker

PR #347, opened 2025-05-14
Clarify format and how it relates to compatibility

### LINKED ISSUE #295: Should certain fields in the deriving specs be required if compatibility != none?
If `compatibility` is set to a value other than `none`, the server may need certain attributes to be set.

In the case of the message spec:
- Either [`messageenvelope`](https://github.com/xregistry/spec/blob/main/message/spec.md#envelope-message-group) or [`protocol`](https://github.com/xregistry/spec/blob/main/message/spec.md#protocol-message-group) MUST be set.

In the case of the schema spec:
- Possibly the [`format`](https://github.com/xregistry/spec/blob/main/schema/spec.md#format)-attribute must be set. However, what if the registry only defines a single format? Do we want to force users to then also set the `format` or can it be somehow implied or derived?

In the case of the endpoint spec:
- [`envelope`](https://github.com/xregistry/spec/blob/main/endpoint/spec.md#envelope) and/or [`envelopeoptions`](https://github.com/xregistry/spec/blob/main/endpoint/spec.md#envelopeoptions) and [`protocol`](https://github.com/xregistry/spec/blob/main/endpoint/spec.md#protocol).

lailabougria:
Fixes #295

## Proposed Changes

- Makes format requires when `compatibility` != `None` and `compatibilityauthority` = `server`
- Specifies that xRegistry managers can set a default value for `format`
- Clarifies the difference between Semantic Versioning and schema versioning to avoid confusion

**Release Note**

```release-note

```

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -829,13 +829,13 @@ defined in the core specification for consistency. It is worth noting that
 the Endpoint [specification](../endpoint/spec.md) does exactly this to
 indicate when an Endpoint (i.e. a Group) is deprecated.
 
-## The `format`-attribute in the Schema spec
+## The `format` attribute in the Schema spec
 
-The `format`-attribute in the Schema spec is required when the
-`compatibility`-attribute is set to a value other than `None`. This may
+The `format` attribute in the Schema spec is required when the
+`compatibility` attribute is set to a value other than `none`. This may
 appear to be overly restrictive, especially if your Registry implementation
 only supports a single format. The reason for making this field required, is
 to ease interoperability of registries through export-import functionality.
 That means that when exporting a Schema Registry that only supports a single
-format, the `format`-attribute will already be set, avoiding issues when
+format, the `format` attribute will already be set, avoiding issues when
 importing this Registry into a server that may support multiple formats.
--- a/core/primer.md
+++ b/core/primer.md
@@ -828,3 +828,14 @@ When doing so it is recommended to use the same attribute definition as
 defined in the core specification for consistency. It is worth noting that
 the Endpoint [specification](../endpoint/spec.md) does exactly this to
 indicate when an Endpoint (i.e. a Group) is deprecated.
+
+## The `format`-attribute in the Schema spec
+
+The `format`-attribute in the Schema spec is required when the
+`compatibility`-attribute is set to a value other than `None`. This may
+appear to be overly restrictive, especially if your Registry implementation
+only supports a single format. The reason for making this field required, is
+to ease interoperability of registries through export-import functionality.
+That means that when exporting a Schema Registry that only supports a single
+format, the `format`-attribute will already be set, avoiding issues when
+importing this Registry into a server that may support multiple formats.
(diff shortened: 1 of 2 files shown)


core/primer.md
```diff
@@ -828,3 +828,14 @@ When doing so it is recommended to use the same attribute definition as
 defined in the core specification for consistency. It is worth noting that
 the Endpoint [specification](../endpoint/spec.md) does exactly this to
 indicate when an Endpoint (i.e. a Group) is deprecated.
+
+## The `format`-attribute in the Schema spec
+
+The `format`-attribute in the Schema spec is required when the
```
duglin, line 834:
```suggestion
The `format` attribute in the Schema spec is required when the
```

core/primer.md
```diff
@@ -828,3 +828,14 @@ When doing so it is recommended to use the same attribute definition as
...
+`compatibility`-attribute is set to a value other than `None`. This may
```
duglin, line 835:
```suggestion
`compatibility` attribute is set to a value other than `None`. This may
```

core/primer.md
```diff
@@ -828,3 +828,14 @@ When doing so it is recommended to use the same attribute definition as
...
+appear to be overly restrictive, especially if your Registry implementation
+only supports a single format. The reason for making this field required, is
+to ease interoperability of registries through export-import functionality.
+That means that when exporting a Schema Registry that only supports a single
+format, the `format`-attribute will already be set, avoiding issues when
```
duglin, line 840:
```suggestion
format, the `format` attribute will already be set, avoiding issues when
```

schema/spec.md
```diff
@@ -317,13 +319,19 @@ the core xRegistry Resource
...
+  Managers of the xRegistry instance can set a default value for this
+  attribute, making it a REQUIRED attribute.
 - Constraints:
   - If present, MUST be a non-empty string
   - MUST follow the naming convention `{NAME}/{VERSION}`, whereby `{NAME}` is
     the name of the schema format and `{VERSION}` is the Version of the schema
     format in the format defined by the schema format itself.
   - MUST be a Version level attribute defined within the `attributes` section
     of the model.
+  - MUST be present if the `compatibility`-attribute is set to a value other
```
duglin, line 332:
```suggestion
  - MUST be present if the `compatibility` attribute is set to a value other
```

duglin (changes requested):

schema/spec.md
```diff
@@ -317,13 +319,19 @@ the core xRegistry Resource
   For many schema registry use cases this attribute is important for schema
   validation purposes, and as such implementations can choose to modify the
   model to make this attribute mandatory.
+
+  Managers of the xRegistry instance can set a default value for this
+  attribute, making it a REQUIRED attribute.
```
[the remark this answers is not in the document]
jskeet: Right. Sounds like it's consistent with the rest of the spec then, even if it's not necessarily the word I'd have used ;)

duglin:
Approved on the 5/22 call

PR #348, opened 2025-05-14
Week of May 12

### LINKED ISSUE #339: define an "xidtype" (and use it)

### LINKED ISSUE #341: ximport should be one level up & a question...
`ximport` is defined as:

```
"model": {
  "groups": {
    "messagegroups": {
      "plural": "messagegroups",
      "singular": "messagegroup",
      "resources": {
        "messages": {
          "plural": "messages",
          "singular": "message"
        }
      }
    },
    "endpoints": {
      "plural": "endpoints",
      "singular": "endpoint",
      "resources": {
        "ximport": [ "messagegroups/messages" ]
      }
    }
  }
}
```
Notice it appears in the `resources` map. This is problematic because, typically, map "values" all have the same "type". In this case it's not a "resource model definition" but instead it's an "array of strings". While we could make the map's "value" an "any", that seems wrong.

I'd like to suggest that we don't "pollute" the map with this extra info. Instead we move `ximport` up to be an attribute on the Group:

```
    "endpoints": {
      "plural": "endpoints",
      "singular": "endpoint",
      "ximport": [ "messagegroups/messages" ],
      "resources": {
        ...
      }
```

The group model schema then becomes:
```
    "groups": {
      "STRING": {                       # Key=plural name, e.g. "endpoints"
        "plural": "STRING",             # e.g. "endpoints"
        "singular": "STRING",           # e.g. "endpoint"
        "description": "STRING", ?
        "modelversion": "STRING", ?     # Version of the group model
        "compatiblewith": "URI", ?      # Statement of compatibility with model spec
        "labels": { "STRING": "STRING" * }, ?
        "attributes": { ... }, ?        # Group level attributes/extensions

        "ximport": [ STRING, * ], ?
        "resources": { ... }
    }
```

I also noticed that we only use "ximport" to share Resource models, not Group models. Does it make sense to keep the word "ximport" so generic? Plus, we're no longer forced to call it "x"-something because don't need to worry about name collisions any more. Options:
- `xresources`
- `sharedresources`
- `ximportedresources`
- `ximportresources`   <---
- `xresourcesimport`
- `ximport`

I'm leaning towards "importresources"

### LINKED ISSUE #343: Split "model" and "schema"
Today we mix the concepts of a "registry model" and "registry schema". By that I mean, a "registry schema" is what a client/tooling would find useful when trying to verify the data it gets back from the server, or describing the info their users might see (again, from the server) or the data it might send to the server on a PUT when they're trying to update instance data.

The "registry model" on the other hand is a subset of the "registry schema". It's focused just on the parts of the schema that are directly (not indirectly) definable by users. Where "indirectly" are fields in the schema that appear as a result of a "direct user definition", but not something the user should define themselves. E.g. a "direct" definition is the specification of a Group. An "indirect" definition is the creation of that group's "XXXcount" attribute at the Registry level. Another example are the RESOURCE/RESOURCEurl/RESOURCEbase64 fields.

I think we need to have a clear separation between these two concepts and allow the user to ask for each separately - where the "schema" would be read-only and the "model" would be mutable.

Here's the scenario that I ran into that cause this issue:
- I was adding "xr model group create" and "xr model group delete" commands
- "create" was easy, xr would just "GET /model", add the new group to the "groups" collection in the model JSON and "PUT /model" with that JSON
- "delete" was problematic because when I tried to do a similar flow, while deleting the group from the model.groups collection was easy, the group's "count,url,map" attributes (at the Registry level) were still part of the PUT's request body and they would then be defined as "user extensions" in the server's model

While I can certainly make "xr" remove them, I realized that in order for any client do this simple act it requires a fair amount of knowledge of which attributes are really "user controlled" vs which are "system controlled". I think this mixing of "schema" and "model" is problematic/confusing by not providing users with a clear list of "user controlled" fields.

If we had the ability for a client to ask for "just the user definable parts of the schema/model", or even better "just the exact model that the user originally provided to define the full model", then they can manipulate that w/o needing to worry about any "side effect attributes" that might appear/disappear as a result of changes to the "user definable" bits. Hence, the need for a "model" that's separate from a "schema".

I think what we return for /model today is correct w.r.t. a "schema". So I'm suggesting that we find some other api to return this information (e.g. `/schema` or `/mode/schema or `/model/full`)` so s3 can return it too), and then make `/model` be just the "user definable" parts.

Some next steps:
- define how clients ask for the schema
- define how clients ask for the schema during an export since /model in the output would be for the model not the schema
- define which attributes are part of the model vs schema

### LINKED ISSUE #344: Should we support >, <, >=, <=, <> for strings in ?filter ?
In particular, it'll make it easier to GET stuff that's old/newer than a certain date

duglin:
- typos
- add an "xidtype" to the type system (/[GROUPS[/RESOURCE[/versions]]])
- s/ximport/ximportresources/
- add <, >, <=, >=, <> for filters
- add "/modelsource"  and "modelsource" attribute to get the model definition what was used to create/update the model. "model" then becomes read-only.

Fixes #339
Fixes #341
Fixes #344
Fixes #343

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -25,7 +25,7 @@ automation and tooling.
   - [Registry Capabilities](#registry-capabilities)
   - [Registry Model](#registry-model)
     - [Retrieving the Registry Model](#retrieving-the-registry-model)
-    - [Updating the Registry Model](#updating-the-registry-model)
+    - [Creating or Updating the Registry Model](#creating-or-updating-the-registry-model)
   - [Groups APIs](#groups-apis)
     - [Retrieving a Group Collection](#retrieving-a-group-collection)
     - [Creating or Updating Groups](#creating-or-updating-groups)
@@ -134,13 +134,13 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
     "STRING": ... *                     # Extension capabilities
   }, ?
 
-  "model": {                            # Only if inlined
+  "model": {                            # Full model. Only if inlined
     "labels": { "STRING": "STRING" * }, ?
     "attributes": {                     # Registry level attributes/extensions
       "STRING": {                       # Attribute name (case-sensitive)
         "name": "STRING",               # Same as attribute's key
         "type": "TYPE",                 # string, decimal, array, object, ...
-        "target": "STRING", ?           # If "type" is "xid" or "url"
+        "target": "XIDTYPE", ?          # If "type" is "xid" or "url"
         "namecharset": "STRING", ?      # If "type" is "object"
         "description": "STRING", ?
         "enum": [ VALUE * ], ?          # Array of scalar values of type "TYPE"
@@ -153,7 +153,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
         "attributes": { ... }, ?        # If "type" above is object
         "item": {                       # If "type" above is map,array
           "type": "TYPE", ?             # map value type, or array type
-          "target": "STRING", ?         # If this item "type" is xid/url
+          "target": "XIDTYPE", ?        # If this item "type" is xid/url
           "namecharset": "STRING", ?    # If this item "type" is object
           "attributes": { ... }, ?      # If this item "type" is object
           "item": { ... } ?             # If this item "type" is map,array
@@ -166,6 +166,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
         } ?
       } *
     },
+    "modelsource": { ... }, ?
 
     "groups": {
       "STRING": {                       # Key=plural name, e.g. "endpoints"
@@ -176,6 +177,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
         "compatiblewith": "URI", ?      # Statement of compatibility with model spec
         "labels": { "STRING": "STRING" * }, ?
         "attributes": { ... }, ?        # Group level attributes/extensions
+        "ximportresources": [ "XIDTYPE", * ], ?   # Include these Resources
 
         "resources": {
           "STRING": {                   # Key=plural name, e.g. "messages"
@@ -199,6 +201,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
       } *
     } ?
   }, ?
+  "modelsource": { ... }, ?                        # Input model, if inlined
 
   # Repeat for each Group type
   "GROUPSurl": "URL",                              # e.g. "endpointsurl"
@@ -360,6 +363,7 @@ information about each data type):
 - `URLABSOLUTE`
 - `URLRELATIVE`
 - `XID`
+- `XIDTYPE`
 - `TYPE` - one of the allowable data type names (MUST be in lower case) listed
   in [Attributes and Extensions](#attributes-and-extensions)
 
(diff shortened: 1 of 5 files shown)


core/spec.md
```diff
@@ -618,6 +622,11 @@ be one of the following data types:
   value references the correct type of Registry entity. See the definition of
   the [`target` model attribute](#model.target) for more information.
   Its value MUST start with a `/`.
+- `xidtype` - MUST be a `URL` reference to an xRegistry model type. The
+   reference MUST point to oe of: the Registry itself (`/`), a Group type
```
jskeet, line 626:
```suggestion
   reference MUST point to one of: the Registry itself (`/`), a Group type
```

core/spec.md
```diff
@@ -2247,7 +2289,7 @@ Regardless of the mechanism used to update the capabilities, the Registry's
 In order for a client to discover the list of available values for each
 capability, an HTTP `GET` MAY be sent to the `/capabilities` API with the
 `?offered` query parameter and the response MUST adhere to the following
-(which borrows many of the same structure from the `/model` API):
+(which borrows many of the same structure from the model definition language):
```
jskeet, line 2292: "many of the same structures" or "much of the same structure" perhaps?

core/spec.md
```diff
@@ -3190,36 +3167,68 @@ Clarifying the  usage of the `attributes`, `resourceattributes` and
...
 
-To retrieve the Registry Model as a stand-alone entity, an HTTP `GET` MAY be
-used.
+The Registry model is available in two forms:
+- The full "model" with all possible aspects of the model defined.
+- The "modelsource" form represents just the model aspects specified when
+  the model was defined.
+
+The full "model" view can be thought of as a full schema definition of what the
+messages exchanges with the server might look like. As such, it'll include
```
jskeet, line 3176:
```suggestion
messages exchanges with the server might look like. As such, it includes
```

core/spec.md
```diff
@@ -3190,36 +3167,68 @@ Clarifying the  usage of the `attributes`, `resourceattributes` and
...
+the Groups, Resources, model-specific attributes, extension attributes,
+specification defined attributes and overrides to those specification defined
+attributes.
+
+The "modelsource" view of the model is just what was provided by the user when
+the model was defined, or last edited. It is expected that this view of the
+model will be much smaller than the full model and will only include
```
jskeet, line 3183:
```suggestion
model is much smaller than the full model and only includes
```

core/spec.md
jskeet: (Here and below, the future tense feels a bit odd.)

core/spec.md
```diff
@@ -3190,36 +3167,68 @@ Clarifying the  usage of the `attributes`, `resourceattributes` and
...
+domain-specific information. While specification defined attributes MAY appear
+in this document, they are NOT RECOMMENDED since the server will automatically
+add them and so users do not need to concern themselves with those details.
+
+The modelsource document will always be a semantic subset of the full model
```
jskeet, line 3188:
```suggestion
The modelsource document is always a semantic subset of the full model
```

core/spec.md
```diff
@@ -6710,29 +6566,39 @@ Where:
...
+    - `>=` refers to "greater than or equal to".
+    - Wildcards (`*`) (see below) MUST NOT be present in the `VALUE`.
+    - MUST only be used on numeric or string attribute types.
+
+When comparing an `ATTRIBUTE` to the specified `VALUE`, the type of the
+attribute impacts how the comparisons are done:
+- For boolean attributes, a successful `=` comparison MUST be an exact
+  case-sensitive match (`true` or `false`).
+- For numeric attributes, standard numeric comparisons rules apply.
+- For string attributes, the strings MUST be compared in a case-insensitive
```
jskeet, line 6593: String *values* can be any Unicode, I assume? In which case we need to be more specific about comparisons. Ordinal, culture-insensitive, something else?

core/spec.md
duglin: done

core/spec.md
duglin: done - both

core/spec.md
duglin: done

core/spec.md
jskeet: I'd need to know what we actually want to happen first :)

duglin:
Approved on the 5/15 call

PR #353, opened 2025-05-15
Alternating "required" fields for when ifvalues are set.

clemensv:
Picked from #328 

Alternating "required" fields for when ifvalues are set.


duglin:
Merging since it's just a tooling change

PR #354, opened 2025-05-20
Week of May 19

### LINKED ISSUE #345: model edit scenario question
Scenario:
- define a model with
  - GroupA / ResourceA
  - GroupB / (ximports GroupA/ResourceA)
- create some Resources under GroupB
- edit the model such that GroupB's definition of ResourceA is no longer using ximport, it's defined locally, but the definition of the Resource remains the same

What is the expected output?
1 - all Resources remain untouched - the "are the Resource still compliant with the model?" check returns "yes"
2 - all Resources were deleted because the old Resource type was deleted and a new one just happened to be created that looks like the old one

I'm inclined to say that the result is option 1. I'm not sure we should get into the position of trying to analyze "how" a model gets into its final state and take action based on the "how". I think it's safer to take the approach of "compare the resulting (normalized) model with the data" and then take any appropriate action. And in this case, from a "compare the runtime data against the normalized model" perspective, nothing has changed, so the data should remain untouched.

Propose spec changes:
in the section where we talk about use of "$include" and "ximportresources", I'd like to add some text along the lines of

- - -
For the purposes of validating that the existing entities in the Registry are compliant with the model, the
mechanisms used to define the model (e.g. `$include` vs `ximportresources` vs defined locally) MUST NOT
impact that analysis. In other words, model updates that have no semantic changes but rather switch between
one of those 3 mechanisms would not invalidate any existing entities in the Registry.
- - -

### LINKED ISSUE #346: sorting
Proposal:
- support a "sort" flag in general
- `?sort=propPath[=asc|desc`]   (default is 'asc')
- when pagination is turned on, and user doesn't specify a sort, the default sort is: ascending by ID (eg `?sort=versionid=asc`)
- `?sort=version=asc`  (sort on versionmode order) <- think about for later

### LINKED ISSUE #349: Unique identity of a model entity
Proposal:
- a model entity is uniquely identified by its Plural/Singular combination, despite both needing to be unique across all instances of either name.
- if a model entity (group or resource) changes it's Plural or Singular name due to a model update then the server MUST interpret it as a new model entity and any existing instances of the old model entity MUST be deleted

I fear that if we allow either to change then the usage of either value outside of something the xReg code controls could be misaligned. E.g. what if the user defines an extension that uses the Singular name as part of it's name or value. Neither of those will be automatically detected or updated by a model rename of the Singular name.

Net - an instances Plural/Singular model values are immutable.

duglin:
- Make it clear that syntax changes related to how the model is defined does not invalidate existing entities in the registyr
- Make the pural/singular names of Groups/Resources immutable
- Add support for ?sort

Fixes #349
Fixes #345
Fixes #346

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -117,7 +117,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
       "collections",? "doc",? "epoch",? "filter",? "inline",?
       "nodefaultversionid",? "nodefaultversionsticky",? "noepoch",?
       "noreadonly",?  "offered",? "schema",? "setdefaultversionid",?
-      "specversion",?
+      "sort",? "specversion",?
       "STRING" *
     ],
     "mutable": [                        # What is mutable in the Registry
@@ -760,6 +760,7 @@ The definition of each attribute is defined below:
 
 - Constraints:
   - REQUIRED.
+  - MUST be immutable.
   - MUST be a non-empty string consisting of [RFC3986 `unreserved`
     characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)
     (ALPHA / DIGIT / `-` / `.` / `_` / `~`) and `@`, MUST start with
@@ -775,7 +776,6 @@ The definition of each attribute is defined below:
     outside of the entity, and it MUST be the same as the entity's existing
     `SINGULARid` if one exists, otherwise an error
     ([mismatched_id](#mismatched_id)) MUST be generated.
-  - MUST be immutable.
 
 - Examples:
   - `a183e0a9-abf8-4763-99bc-e6b7fcc9544b`
@@ -857,12 +857,12 @@ of the existing entity. Then the existing entity would be deleted.
 
 - Constraints:
   - REQUIRED if the `shortself` capability is enabled.
+  - MUST be immutable for the lifetime of the entity.
   - MUST NOT appear in responses if the `shortself` capability is disabled.
   - MUST be unique across all entities in the Registry.
   - MUST be a non-empty absolute URL referencing the same entity as the `self`
     URL, either directly or indirectly via an HTTP redirect.
   - MUST be a read-only attribute.
-  - MUST be immutable for the lifetime of the entity.
 
 - Examples:
   - `https://tinyurl.com/xreg123` redirects to
@@ -889,11 +889,11 @@ of the existing entity. Then the existing entity would be deleted.
 
 - Constraints:
   - REQUIRED.
+  - MUST be immutable.
   - MUST be a non-empty relative URL to the current entity.
   - MUST be of the form: `/[GROUPS/gID[/RESOURCES/rID[/meta | /versions/vID]]]`.
   - MUST start with the `/` character.
   - MUST be a read-only attribute.
-  - MUST be immutable.
 
 - Examples:
   - `/endpoints/ep1`
@@ -2792,6 +2792,7 @@ The following describes the attributes of Registry model:
 - `groups."STRING".plural`
   - Type: String.
   - REQUIRED.
+  - MUST be immutable.
   - The plural name of the Group type e.g. `endpoints` (`GROUPS`).
   - MUST be unique across all Group types (plural and singular names) in the
     Registry.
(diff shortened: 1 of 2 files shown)


core/spec.md
```diff
@@ -3465,6 +3469,13 @@ the model update requests generate an error
 ([model_compliance_error](#model_compliance_error)) if existing entities are
 not compliant.
 
+For the purposes of validating that the existing entities in the Registry are
+compliant with the model, the mechanisms used to define the model (e.g.
+`$include` vs `ximportresources` vs defined locally) MUST NOT impact that
+analysis. In other words, model updates that have no semantic changes but
+rather switch between one of those 3 mechanisms would not invalidate any
```
jskeet, line 3476: "must not" or "cannot" rather than "would not" perhaps? (It's not quite clear to me at the moment whether this is stating something that is inherently true, or specifying a rule that a server must follow.)

core/spec.md
```diff
@@ -6788,6 +6800,53 @@ plural name for the collection in its defined case.
...
+the server is unable to support sorting over it, then an error
+([unsupported_sort_attribute](#unsupported_sort_attribute)) MUST be generated.
+If the specified attribute is not valid then an error
+([invalid_sort_attribute](#invalid_sort_attribute)) MUST be generated.
+
+When [pagination](../pagination/spec.md) is used to return the results, but
+the `?sort` flag is not specified, then the server MUST sort the results on the
+entities' `SINGULARid` value in ascending order.
+
+Sorting of strings MUST be done in a case-insensitive way.
```
jskeet, line 6837: Just another reminder that we need to tighten this up (and any other occurrences of "case-insensitive").

core/spec.md
```diff
@@ -6788,6 +6800,53 @@ plural name for the collection in its defined case.
...
+
+Entity's that do not have a value for the specified attribute MUST be treated
```
jskeet, line 6839:
```suggestion
Entities that do not have a value for the specified attribute MUST be treated
```

core/spec.md
```diff
@@ -6788,6 +6800,53 @@ plural name for the collection in its defined case.
...
+the same as if they had the "lowest" possible value for that attribute. If
+more than one entity shares the same attribute value then the `SINGULARid`
+MUST be used as a secondary sorting key, using the same `asc`/`desc` value
+specified for the primary sorting key.
+
+Some examples:
+- `GET /endpoints?sort=name`          # Sort (asc) on 'name', then 'endpointid'
+- `GET /endpoints/e1/messages?sort=messageid=desc` # Sort (desc) on 'messageid'
+- `GET /endpoints?sort=labels.stage`  # Sort (asc) on `labels.stage`
```
jskeet, line 6848: Then on endpointid?

core/spec.md
```diff
@@ -7193,6 +7261,15 @@ something unexpected happened in the server that caused an error condition.
 * Data: {URL to entity being processed}
 * Detail: {information specific to the processing details}
 
+#### unsupported_sort_attribute
+
+* Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#unsupported_sort_attribute`
+* Code: `400 Bad Request`
+* Instance: The request URL
+* Title: `The specified "sort" attribute is not supported`
```
jskeet, line 7269: Perhaps "Sorting on the specified attribute is not supported"? It's not that the attribute itself isn't generally supported, it's that you can't sort on it. (At least, that's what I *think* is meant.)

As a thought, is this likely to be due to the type, or the specific attribute? (e.g. because it's not indexed, even though it could be?)

core/spec.md
duglin: went with MUST NOT.

It's making sure that a server doesn't look at these types of changes as "delete group X + add group X" and in the process ends up deleting all instances of that group. Treating an edit of an existing group should be obvious that it's not, but switching from "ximport" to "inlining" the definition might be viewed as something more drastic to some people, and we need to stop that thinking.

core/spec.md
duglin: yep - thanks for the reminder: https://github.com/xregistry/spec/issues/356

core/spec.md
duglin: done

core/spec.md
duglin: yup, added that.

core/spec.md
duglin: I like that phrasing better - thanks.

It's a bit vague as to why this might happen because some implementation choices might find it hard to sort on data that's handled differently from others. For example, the "count" attributes, at least in my impl, are not stored in the DB and are only calculated during serialization due to "filtering". So sorting on them would be a real challenge for me.

duglin:
Approved on the 5/22 call

Issue #358, opened 2025-05-22
Add ":" to allowable chars for SINGULARid


duglin:
today we say:
- - -
MUST be a non-empty string consisting of [RFC3986 unreserved characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3) (ALPHA / DIGIT / - / . / _ / ~) and @, MUST start with ALPHA, DIGIT or _ and MUST be between 1 and 128 characters in length.
- - -
@clemensv is asking for ":" to be added

Issue #357, opened 2025-05-22
How to represent spec variables - e.g. GROUPS


duglin:
Today we represent variables (meant to be replaced with runtime values) as UPPERCASE words - e.g. GROUPS, RESOURCES. This might be problematic for some tooling (e.g. AI reading our specs) and people reading the spec as it's more "normal" for variables to have clear separator chars around them. For example `${VAR}`.

Current usages of variables in the specs include:
- Concrete URLs/XIDs: `/GROUPS/gID` - both `GROUP` and `gID` are variables
- Abstract URLs/XIDtypes: `/GROUPS/RESOURCES/versions`
- Text: This includes extension and GROUPS collection attribute names.
- Text: While `SINGULARid` can be something like a UUID...
- JSON: `"GROUPSurl": "URL",   ` - both `GROUPS` and `URL` are variables
- JSON comments: `"GROUPS": { GROUPS collection } ?  ` - 2nd `GROUPS` is a comment (more like spec 'text')
- Spec section headers: **`GROUPS` Collections**
- Attribute definitions: `groups."STRING"` - `STRING` is a variable

Some options:

|||||||
| --- | --- | --- | --- | --- | --- |
| GROUPS | `GROUPS` | `GROUPid` | Some `GROUPid` ... | `{ "hi": "URL"}` | `/GROUPS/gID` |
| {GROUPS} | `{GROUPS}` | `{GROUP}id` | Same `{GROUP}id` ... | `{ "hi": "{URL}"}` |  `/{GROUPS}/{gID}` |
| (GROUPS) | `(GROUPS)` | `(GROUP)id` | Same `(GROUP)id` ... | `{ "hi": "(URL)"}` | `/(GROUPS)/(gID)` |
| ${GROUPS} | `${GROUPS}` | `${GROUP}id` | Some `{GROUP}id` ... | `{ "hi": "${URL}"}` | `/{GROUPS}/{gID}` |
| $GROUPS | `$GROUPS` | `$GROUPid` | Some `$GROUPid` ... | `{ "hi": "$URL"}` | `/$GROUPS/$gID` |
| [GROUPS] | `[GROUPS]` | `[GROUP]id` | Some `[GROUP]id` ... | `{ "hi": "[URL]"}` | `/[GROUPS]/[gID]` |
| \<GROUPS> | `<GROUPS>` | `<GROUP>id` | Some `<GROUP>id`... | `{"hi": "<URL>"}` | `/<GROUPS>/<gID>` |

[...]

Issue #356, opened 2025-05-22
need to clarify how string comparisions are done in a unicode world


duglin:

duglin:
check on sorting numerics.... as numbers not strings

duglin:
Sort:
- sorting of strings is STRONGLY RECOMMENDED use Unicode collation based on En-US
- server must normalize all timestamps to UTC
- numbers are not strings
- boolean : false < true
- URI/URL variants - treat as strings

Applies to new comparison operators we added for filter

PR #360, opened 2025-05-27
Week of May 26

duglin:
- change variables from `GROUP` to `<GROUP>`
- add `:` as a valid char in `<SINGULAR>id`
- add info about how to compare the data types

Fixes: #357
Fixes: #358 
Fixes: #356

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -766,7 +766,7 @@ The definition of each attribute is defined below:
   - MUST be immutable.
   - MUST be a non-empty string consisting of [RFC3986 `unreserved`
     characters](https://datatracker.ietf.org/doc/html/rfc3986#section-2.3)
-    (ALPHA / DIGIT / `-` / `.` / `_` / `~`) and `@`, MUST start with
+    (ALPHA / DIGIT / `-` / `.` / `_` / `~`), `:` and `@`, MUST start with
     ALPHA, DIGIT or `_` and MUST be between 1 and 128 characters in length.
   - MUST be case-insensitive unique within the scope of the entity's parent.
   - This attribute MUST be treated as case-sensitive for look-up purposes.
--- a/core/spec.md
+++ b/core/spec.md
@@ -98,18 +98,18 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
 
 ```yaml
 {
-  "specversion": "STRING",
-  "registryid": "STRING",
-  "self": "URL",
-  "shortself": "URL", ?
-  "xid": "XID",
-  "epoch": UINTEGER,
-  "name": "STRING", ?
-  "description": "STRING", ?
-  "documentation": "URL", ?
-  "labels": { "STRING": "STRING" * }, ?
-  "createdat": "TIMESTAMP",
-  "modifiedat": "TIMESTAMP",
+  "specversion": "<STRING>",
+  "registryid": "<STRING>",
+  "self": "<URL>",
+  "shortself": "<URL>", ?
+  "xid": "<XID>",
+  "epoch": <UINTEGER>,
+  "name": "<STRING>", ?
+  "description": "<STRING>", ?
+  "documentation": "<URL>", ?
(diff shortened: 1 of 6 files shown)


core/spec.md
```diff
@@ -6610,20 +6612,28 @@ Where:
...
-  interpreted as a normal character and not as a wildcard. Note that a
-  `<VALUE>` of `*` MUST be equivalent to checking for the existence of the
-  attribute, with any value (even an empty string). In other words, the filter
-  will only fail if the attribute has no value at all.
+For comparing an `<ATTRIBUTE>` to the specified `<VALUE>`, and for purposes
+of sorting (see the [`?sort`](#sort-flag) flag), the type of the attribute
+impacts how the comparisons are done:
+- For boolean attributes, comparisons MUST be an exact case-sensitive match
+  (`true` or `false`). For relative comparisons, `false` is less than `true`.
+- For numeric attributes, standard numeric comparisons rules apply. Note that
```
jskeet, line 6633: Do we support floating point types? If so, we should probably document NaN comparisons (for both equality and relative ordering).

core/spec.md
```diff
@@ -6610,20 +6612,28 @@ Where:
...
+  - `<VALUE>` MAY include one or more wildcard (`*`) characters but only when
+    the comparison operator is `=`, `!=` or `<>`. The presence of a wildcard
+    indicates that any number of characters can appear at that location in the
+    `<VALUE>`. The wildcard MAY be escaped via the use of a backslash (`\\`)
+    character (e.g. `abc\*def`) to mean that the `*` is to be
+    interpreted as a normal character and not as a wildcard. Note that a
+    `<VALUE>` of `*` MUST be equivalent to checking for the existence of the
+    attribute, with any value (even an empty string). In other words, the filter
+    will only fail if the attribute has no value at all.
+  - It is STRONGLY RECOMMENDED to use Unicode collation based on en-US.
```
jskeet, line 6646: Thank you!

core/spec.md
```diff
@@ -6610,20 +6612,28 @@ Where:
...
+- For timestamp attributes, normal string comparison rules apply after
+  the values have been normalized to UTC.
+- For URI/URL variants, normal string comparison rules apply.
```
jskeet, line 6649: Do these support wildcarding?

core/spec.md
```diff
@@ -6610,20 +6612,28 @@ Where:
...
+- For boolean attributes, comparisons MUST be an exact case-sensitive match
+  (`true` or `false`). For relative comparisons, `false` is less than `true`.
+- For numeric attributes, standard numeric comparisons rules apply. Note that
+  numerics MUST NOT be treated as strings for comparison.
+- For string attributes, the following rules apply:
+  - The strings MUST be compared in a case-insensitive manner.
+  - `<VALUE>` MAY include one or more wildcard (`*`) characters but only when
+    the comparison operator is `=`, `!=` or `<>`. The presence of a wildcard
+    indicates that any number of characters can appear at that location in the
+    `<VALUE>`. The wildcard MAY be escaped via the use of a backslash (`\\`)
```
jskeet, line 6640: This "MAY" sounds like it's an optional feature. I think we should have a whole section on escaping, if there isn't one already... because this means we *at least* need to support escaping a backslash as well (otherwise how would you represent "abc backslash followed by anything"?). Are there other characters that can be escaped? Trying to cram it all into this paragraph is likely to end up being a bit much.

jskeet (commented):
I haven't reviewed the "add angle brackets" commit in detail.
Thanks for separating out the other commits!

core/spec.md
duglin: good catch - will do in a follow-on PR
https://github.com/xregistry/spec/issues/361

core/spec.md
duglin: ok - for timestamps and URLs/URIs I make it clear they're just strings and the rules above apply. See if that's ok

PR #382, opened 2025-07-01
Week of June 23 + June 30

duglin:
- Misc clean-up of docs/specs
- Changes due to review of specs


core/spec.md
```diff
@@ -400,15 +400,16 @@ could be a more manageable, and user-friendly, implementation choice.
...
+attributes.
 
 - **Resource**
 
-Resources represent the main data of interest for the Registry. In the
-filesystem analogy, these would be the "files". All Resources exist under a
-single Group and, similar to Groups, have a set of Registry metadata.
-However, unlike a Group which only has Registry metadata, each Resource can
+Resources, typically, represent the main data of interest for the Registry. In
+the filesystem analogy, these would be the "files". All Resources exist under
```
jskeet: "All Resources exist under a single Group" sounds wrong to me, unless I've misunderstood. Why would we allow multiple groups if only one of them can have resources?

I'm assuming the intention is "one Resource can't be under multiple Groups". Perhaps "Each Resource exists under a single Group"?

core/spec.md
```diff
@@ -452,13 +460,17 @@ This entity is meant to serve a few key purposes:
     within those Groups.
 
 **Groups**
-Traversing down the tree structure, below the Registry entity, there will be
+Traversing down the tree structure, below the Registry entity, there can be
```
jskeet: I wonder whether it's a simpler conceptual model to say that there is *always* a set of Groups - but that set can be empty.

core/spec.md
```diff
@@ -469,7 +481,7 @@ quite rich with respect to managing domain-specific data. See the
 **Resources**
 Below, or within, each Group can be a set of Resources. Typically, Resources
 are the main pieces of data managed by the Registry. Like the Registry
-and Group entities, that have a set of xRegistry defined "common" metadata
+and Group entities that have a set of xRegistry defined "common" metadata
```
jskeet: Again, I wonder whether it makes sense for there to always be a set of resources, but that set can be empty.

core/spec.md
```diff
@@ -805,7 +817,7 @@ The definition of each attribute is defined below:
 While `<SINGULAR>id` can be something like a UUID, when possible, it is
 RECOMMENDED that it be something human friendly as these values will often
 appear in user-facing situations such as URLs or as command-line parameters.
-And, in cases where [`name`](#name-attribute) is absent, it might be used as
+In cases where [`name`](#name-attribute) is absent, it might be used as
```
```suggestion
In cases where [`name`](#name-attribute) is absent, the `<SINGULARid>` might be used as
```

(Just to disambiguate the "it". I originally read this as "where the name is absent, the name might be used as the display name" which clearly doesn't make sense, but made me double-take...)

core/spec.md
```diff
@@ -843,7 +855,7 @@ of the existing entity. Then the existing entity would be deleted.
   - MUST be a relative URL of the form `#JSON-POINTER` where the `JSON-POINTER`
     locates this entity within the current document. See [Doc Flag](#doc-flag)
     for more information.
-  - MUST NOT include `$details` suffix on its `<SINGULAR>id`.
+  - This URL MUST NOT include `$details` suffix after its `<SINGULAR>id`.
```
jskeet: Add an article here, e.g. "the `$details` suffix" or "a `$details` suffix"?

core/spec.md
```diff
@@ -891,13 +903,13 @@ of the existing entity. Then the existing entity would be deleted.
...
 - Type: XID
-- Description: An immutable server generated unique identifier of the entity.
+- Description: An immutable server-generated unique identifier of the entity.
   Unlike `<SINGULAR>id`, which is unique within the scope of its parent, `xid`
   MUST be unique across the entire Registry, and as such is defined to be a
   relative URL from the root of the Registry. This value MUST be the same as
   the <PATH> portion of its `self` URL, after the Registry's base URL, without
   any `$` suffix (e.g. `$details`). Unlike some other relative URIs, `xid`
-  values MUST NOT be shortened based on the incoming request's URL. `xid`s
+  values MUST NOT be shortened based on the incoming request's URL, `xid`s
```
jskeet: I think this would be a great place for an elegant semi-colon. The comma feels slightly wrong to me at the moment.

core/spec.md
```diff
@@ -1063,26 +1077,26 @@ of the existing entity. Then the existing entity would be deleted.
     existing value, however a value of `null` MUST use the current date/time
     as the new value.
   - When absent in a write operation request, any existing value MUST remain
-    unchanged, or if not already set, set to the current date/time
+    unchanged, or if not already set, set to the current date/time.
   - During the processing of a single request, all entities that have their
-    `createdat` or `modifiedat` attributes set to the current date/time due
-    to the processing rules above, MUST use the same value in all cases.
+    `createdat` or `modifiedat` attributes set to the current date/time MUST
+    MUST use the same value in all cases.
```
```suggestion
    use the same value in all cases.
```

(otherwise we have MUST MUST)

core/spec.md
```diff
@@ -5071,7 +5090,7 @@ When converting a cross-reference Resource back into a "normal" Resource, the
 following MUST be adhered to:
 - The request MUST delete the `xref` attribute or set it to `null`.
 - A default Version of the Resource MUST be created.
-- Any attributes specified at the Resource level MUST be interpreted as a
+- Any attributes specified at the Resource-level MUST be interpreted as a
```
jskeet: I think this hyphen is incorrect, personally. If it were "Any Resource-level attributes" that would be okay, but with the current structure I think the space was better.

endpoint/spec.md
```diff
@@ -318,38 +318,38 @@ See the [deprecated](../core/spec.md#deprecated) attribute in the core
 - Type: String
 - Description: The name of the specification that defines the Resource
   stored in the registry. Often it is difficult to unambiguously determine
-  what a Resource is via simple inspect of its serialization. This attribute
+  what a Resource is by simply inspecting its serialization. This attribute
```
jskeet:
```suggestion
  what a Resource is by simply inspecting its serialized form. This attribute
```

(I view serialization as a process rather than a result that can be inspected.) Happy for this to be resolved with no action though.

jskeet (commented):
Just a few nits.

Personally I preferred the "being processed" wording, but I'm not going to fuss about it beyond this drive-by comment :)

core/spec.md
duglin: Fixed!

core/spec.md

core/spec.md

core/spec.md
duglin: good catch

core/spec.md

core/spec.md
duglin: done!

core/spec.md
duglin: but we really really mean it! :-)
Done

core/spec.md
duglin: got all of them - for groups and versions too

endpoint/spec.md
duglin: done

core/spec.md
```diff
@@ -407,7 +407,7 @@ attributes.
 - **Resource**
 
 Resources, typically, represent the main data of interest for the Registry. In
-the filesystem analogy, these would be the "files". All Resources exist under
+the filesystem analogy, these would be the "files". Each Resources exist under
```
jskeet: Sorry for the added nit: "Each Resource exists"

jskeet (approved):
One tiny nit, but otherwise fine. You've removed more hyphens than I'd expected, but I think that's fine too :) (It was really only "at the Resource-level" that I objected to; using "Resource-level" as a sort of compound adjective was fine. But I'm definitely not suggesting you revert things!)

core/spec.md

duglin:
Approved on the 7/3 call

PR #383, opened 2025-07-02
first pass at xReg events

### LINKED ISSUE #5: xReg: Detecting changes in the discovery service
@clemens wrote:

It would be great for the discovery service to raise CloudEvents about state changes:

    Types being added/removed/changed
    Services being added/removed/changed

Moved from: https://github.com/cloudevents/spec/issues/676

duglin:
Fixes #5 

@lailabougria

### DIFF
--- a/core/events.md
+++ b/core/events.md
@@ -18,8 +18,9 @@ This specification defines the events that an xRegistry server MAY generate.
 As updates are made to entities within an xRegistry instance, events SHOULD
 be generated to notify interested parties of those changes. This specification
 defines the metadata associated with each event as
-[CloudEvent][https://cloudevents.io) metadata. Whether CloudEvents are used in
-the generation/serialization of the events is OPTIONAL, but it is RECOMENDED.
+[CloudEvent][https://cloudevents.io) context attributes. Whether CloudEvents
+are used in the generation/serialization of the events is OPTIONAL, but it is
+RECOMENDED.
 
 This specification does not mandate the mechanisms by which events are sent
 to consumers, nor does it mandate how consumers register interest in receiving
@@ -71,8 +72,8 @@ the Registry data.
 
 A single interaction with the Registry MAY result in multiple events, however,
 within the scope of one interaction (see
-[`xregcorrelationid`](#xregcorrelationid-context-attribute)) the following
-constraints apply:
+[`xregcorrelationid`](#xregcorrelationid-extension-context-attribute)) the
+following constraints apply:
 - Only one of `created`, `updated` or `deleted` event MUST be generated per
   `subject`.
 - If the interaction involved more than one of those actions, then the single
@@ -84,13 +85,14 @@ constraints apply:
 The following sections specify the metadata defined for xRegistry events.
 Implementations MAY define additional metadata.
 
-### CloudEvent [`type`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type) Context Attribute
+### CloudEvent [`type`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type) Core Context Attribute
 
 The type of action taken on the entity. The value MUST be of the form:
     `io.xregistry.<ENTITY>.<ACTION>`
 where:
-- `<ENTITY>` is the type of xRegistry entity that was acted upon. It MUST
-  be one of:
+- `<ENTITY>` is the type of xRegistry entity
+  ([`subject`](#cloudevent-subject-core-context-attribute)) that was acted
+  upon. It MUST be one of:
   - `registry`
   - `model`
   - `modelsource`
@@ -108,26 +110,27 @@ where:
   Not all `<ACTION>` values are applicable to all entities. See the
   [Entity Events](#entity-events) section for more information.
 
-This metadata MUST be present in each event.
+This context attribute MUST be present in each event.
 
-### CloudEvent [`source`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#source-1) Context Attribute
+### CloudEvent [`source`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#source-1) Core Context Attribute
 
 The xRegistry in which the entity exists. The value MUST be an absolute URL
 to the root of the xRegistry instance.
 
-This metadata MUST be present in each event.
+This context attribute MUST be present in each event.
 
-### CloudEvent [`subject`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#subject) Context Attribute
+### CloudEvent [`subject`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#subject) Core Context Attribute
 
 The `xid` of the entity acted upon. While this attribute is OPTIONAL in the
 CloudEvents specification, it is REQUIRED to be present in an xRegistry event.
 
 Note: constructing a URL by appending the `subject` value to the `source`
-value MUST result in an absolute URL to the entity.
+value MUST result in an absolute URL to the entity (assuming any trailing `/`
+on `source` is removed since XIDs always start with `/`).
 
-This metadata MUST be present in each event.
+This context attribute MUST be present in each event.
 
-### CloudEvent [`time`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#time) Context Attribute
+### CloudEvent [`time`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#time) Core Context Attribute
 
 The time of when the interaction occurred. This value MUST be the same for
 all events generated within the same interaction.
@@ -138,9 +141,10 @@ used SHOULD match the current date/time used by the Registry during the
 processing of the interaction. Typically, this will be the `modifiedat` value
 assigned to the entities being processed.
 
-This metadata is NOT REQUIRED to be present in the event, but is RECOMMENDED.
+This context attribute is NOT REQUIRED to be present in the event, but is
+RECOMMENDED.
 
-### `xregcorrelationid` Context Attribute
+### `xregcorrelationid` Extension Context Attribute
 
 A value that uniquely identifies the interaction in which one or more events
 occured. This value has the following constraints:
@@ -163,7 +167,8 @@ xRegistry-xregcorrelationid: <STRING>
 This allows for clients to identify which events were generated as a result of
 each request-response operation.
 
-This metadata is NOT REQUIRED to be present in the event, but is RECOMMENDED.
+This context attribute is NOT REQUIRED to be present in the event, but is
+RECOMMENDED.
 
 ### CloudEvent [`data`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#event-data)
 
@@ -199,12 +204,14 @@ interaction, per the rules previously stated, only one `updated` event
 will be generated. This means the attribute names of all impacted attributes
 MUST be merged into one `changed` list.
 
-While `changed` is OPTIONAL, it is RECOMMENDED to be present due to its
-usefulness for consumers. However, if exposure of this information would be
-inappropriate for some scenarios then it MAY be excluded. For example, for
-privacy/security reasons.
+While `changed` is OPTIONAL, it is RECOMMENDED to be present on event where
+it is permitted due to its usefulness for consumers. However, if exposure of
+this information would be inappropriate for some scenarios then it MAY be
+excluded. For example, for privacy/security reasons.
 
-Implementations MAY define additional attributes as siblings to `changed`.
+While this specification only shows `data` being present when `changed` is
+permitted, implementations MAY define their own metadata to be included
+in the `data` of a CloudEvent.
 
 This metadata (`data`) is NOT REQUIRED to be present in the event, but is
 RECOMMENDED.
@@ -218,11 +225,11 @@ value.
 
 - Action: `updated`
   - MUST be generated when a Registry's attribute is updated, where each
-    modified attribute MUST be included in `changed`.
+    modified attribute MUST be included in `changed`, if present.
 
-  - MUST be generated when a Group is created or deleted, where `changed`
-    includes (at least) `epoch`, `modifiedat`, `<GROUPS>` and `<GROUPS>count`
-    attributes names.
+  - MUST be generated when a Group is created or deleted, where `changed`, if
+    present, includes (at least) `epoch`, `modifiedat`, `<GROUPS>` and
+    `<GROUPS>count` attributes names.
 
     While the `<GROUPS>` attribute is present in `changed` due to the child
     collection changing, in order to see which specific Groups were impacted
@@ -230,7 +237,7 @@ value.
     would need to be examined.
 
   - MUST be generated when `model`, `modelsource` or `capabilities` are
-    updated and the appropriate attribute MUST appear in `changed`.
+    updated and the appropriate attribute MUST appear in `changed`, if present.
 
 The `io.xregistry.registry.created` and `io.xregistry.registry.updated` events
 are not defined as part of this specification as those operations are not
@@ -246,8 +253,8 @@ performs those operations will define those events, if needed.
 
   - MUST NOT include a `changed` list.
 
-  - A `io.xregistry.registry.upated` event will also be generated where the
-    `model` attribute will be included in `changed`.
+  - A `io.xregistry.registry.updated` event will also be generated where the
+    `model` attribute will be included in `changed`, if present.
 
 Often `io.xregistry.model.updated` and `io.xregistry.modelsource.updated`
 will be generated at the same time since model updates are most likely done
@@ -263,8 +270,8 @@ to watch for `io.xregistry.model.update` events, not
 
   - MUST NOT include a `changed` list.
 
-  - A `io.xregistry.registry.upated` event will also be generated where the
-    `modelsource` attribute will be included in `changed`.
+  - A `io.xregistry.registry.updated` event will also be generated where the
+    `modelsource` attribute will be included in `changed`, if present.
 
 Often `io.xregistry.model.updated` and `io.xregistry.modelsource.updated`
 will be generated at the same time since model updates are most likely done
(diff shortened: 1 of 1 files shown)


duglin:
This includes stuff that's not agreed to yet but I added it to 1) force a discussion and 2) to see what it would look like.
Most should have a "TODO" near it.

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+- [Notations and Terminology](#notations-and-terminology)
+  - [Notational Conventions](#notational-conventions)
+- [Event Definition](#event-definition)
+- [Entity Events](#entity-events)
+- [Sample xRegistry Interactions](#sample-xregistry-interactions)
+
+## Overview
+
+As updates are made to entities within an xRegistry instance, events SHOULD
+be generated that allows for interested parties to be notified of those
```
lailabougria:
```suggestion
be generated that allow for interested parties to be notified of those
```

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+changes. This specification defines the metadata associated with each event
+as [CloudEvent][https://cloudevents.io) metadata. Whether CloudEvents is used
```
lailabougria:
```suggestion
as [CloudEvent][https://cloudevents.io) metadata. Whether CloudEvents are used
```

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+in the generation/serialization of the events is OPTIONAL, but it is
+RECOMENDED.
+
+This specification does not mandate the mechanisms by which events are sent
+to consumers, nor does it mandate how consumers registry interest in receiving
```
lailabougria:
```suggestion
to consumers, nor does it mandate how consumers register interest in receiving
```

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+the JSON snippets are used for readability and are not normative.
+
+Use of `<...>` the notation indicates a substitutable value where that is
+meant to be replaced with a runtime situational-specific value as defined by
+the word/phase in the angled brackets. For example `<NAME>` might be expected
+to be replaced by the "name" of the item being discussed.
+
+## Event Definition
+
+Notification of changes to the entities within an xRegistry instance MAY be
```
lailabougria: Should this be a SHOULD?

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+constraints apply:
+- Only one of `created`, `updated` or `deleted` event MUST be generated per
+  `subject`.
+- If the interaction involved more than one of those actions, then the single
+  event generated MUST be chosen in the following order of precedence:
+  `deleted`, `created`, `updated`.
+- Only one event with the same `type` and `subject` combination MUST be
+  generated.
+
+The following sections defines the metadata that might appear within each
```
lailabougria:
```suggestion
The following sections defines the metadata that SHOULD appear within each
```

As in, this is recommended to include, and other stuff can be added?

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+- This attribute MUST NOT appear on the `created` or `deleted` events.
+- This specification does not mandate any particular order to the attribute
+  names in the array.
+
+There are certain events where non-top-level attributes are included, and
+those will be noted in the [Entity Events](#entity-events) sections below.
+When they do appear they MUST use the JSONPath dot (`.`) notation to express
+the traversal to the attribute of interest.
+
+If a entity is updated for multiple reasons during the processing of an
```
lailabougria:
```suggestion
If an entity is updated for multiple reasons during the processing of an
```

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+value.
+
+### `registry` Events
+
+- Action: `updated`
+  - MUST be generated when a Registry's attribute is updated.
+  - MUST be generated when a Group is created or deleted, where `changed`
+    includes (at least) `epoch`, `modifiedat`, `<GROUPS>` and `<GROUPS>count`
+    attributes names.
+    TODO: GROUPS*??
```
lailabougria: I will drop the reasoning here, but it applies to every collection (resources, versions).

Whenever the collection changes (something added or removed), the parent entity changes (epoch+modified at, assuming nothing else changed), generating a changed event, and the lower-level entity will generate a created or deleted event as well. 

My main concern with adding the collection to the changed attributes array is that we're signaling the same logical operation through two different events, meaning that any subscriber who chooses to subscribe to all events, must understand that they should skip processing of xregistry collection entities as those are also signaled through dedicated events, forcing consumers to understand these event relationships as well as understanding whether the other created/deleted are also being consumed elsewhere or not (which is especially tricky when not all events are subscribed to from day 1, but there's an evolution of first subscribing to one event and later, adding another). When consumers are not aware, this could lead to duplicate processing (waste of resources as we're updating the same thing), or issues due to parallel processing of those events. The updated + created/deleted event will arrive very close to each other making it very likely that they're processed in parralel (arguably, this can still happen due to touching the same resource). One reason against this approach would be that this may cause a divergence in epoch values (process updated bumps epoch of the registry to 2, process created bumps epoch  of the registry to 3). Is that something we care about?

An alternative is to not emit a created/deleted event on the level of the collection's entity and leave it upto the updated event to identify collection changes. This has the downside that consumers who are only interested in created/deleted events have to subscripe to the higher-level updated event and introduce filtering, which depending on the broker may require additional logic. Also, it would complicate processing further as the consumer doesn't know which specific entity was added/removed from the collection, leaving it up to the consumer to figure this out. Another point against this solution is that, due the importance xregistry collection entities have in the spec, they should have their own dedicated events (as they do have dedicated APIs too).

Personally I mostly care about a single event being emitted for a single logical operation.

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+
+- Action: `deleted`
+  - MUST be generated when a Group is deleted.
+
+### `resource` Events
+
+- Action: `created`
+  - MUST be generated when a new Resource is created.
+  - Note that at least one `io.xregistry.version.created` event will also be
+    being generated since at least one Version will also be created.
```
lailabougria:
```suggestion
   generated since at least one Version will also be created.
```

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+
+- Action: `updated`
+  - MUST be generated when a Resource's (default Version) attribute is updated.
+  - MUST be generated when a Resource's `meta` attribute is updated, where
+    `changed` includes (at least) the changed top-level `meta` attributes
+    names prefixed with `meta.`. For example, `meta.defaultversionid`.
+    Note, this includes any updates to the `deprecated` sub-object, even
+    though a `deprecation` event is also generated.
+    TODO: meta.deprecated.* too??
```
lailabougria: This depends on the resolution of the collection discussion, I'd keep it consistent with that decision.

core/events.md
```diff
@@ -0,0 +1,520 @@
...
+  Note that updating a Resource's `meta.defaultversionid` attribute will
+  generate a `io.xregistry.resource.updated` event but it MUST NOT generate
+  a `io.xregistry.version.updated` event for either the prior or new default
+  Version.
+
+- Action: `deprecation`
+  - MUST be generated when a Resource's `meta.deprecated` sub-object is
+    updated, where "changed" includes the list of top-level attribute names
+    from the `deprecated` sub-object that were modified.
+
```
lailabougria, line 370:
```suggestion

Note that the same event will be generated when the `deprecation` sub-object is set, any of its attributes updated, or removed (set to NULL).
```

core/events.md
lailabougria: Another argument in favour of omitting the collection from the updated event is that these events can easily be reconciled as related through the correlationid.

core/events.md
duglin: I couldn't decide if "allow" or "allows" sounds better so I went with a co-pilot suggestion:
As updates are made to entities within an xRegistry instance, events SHOULD
be generated to notify interested parties of those changes.

core/events.md
duglin: done

core/events.md
duglin: done

core/events.md
duglin: depends on how pushy we want to be :-)
I'll go with SHOULD for now.

core/events.md
duglin: done

core/events.md
duglin: done

core/events.md
duglin: done

core/events.md
duglin: isn't this a repeat of the previous sentence?

lailabougria: I guess I wanted to be very explicit. 🫣

duglin: ok - picked up your text and merged it with the previous sentence

core/events.md
duglin: actually, I rewrote the sentence. I don't think "SHOULD" is appropriate because some attributes are required, which is why I didn't use SHOULD in this sentence when I first wrote it. I leave that normative-ness to the definition of each attribute.

core/events.md
lailabougria: Doug pointed out the concern that the updated event wouldn’t be self-contained without specifying that the collection changed, and therefore consumers would be “forced” to subscribe to updated on the parent + created/deleted on the child. This is true either way. If we assume we add the collection to the changed attributes, the consumer still needs to subscribe to the child-events because collection-changed doesn’t provide sufficient information: what changed in the collection? What was added? Or deleted? Or added and deleted simultaneously? How many collection entries were touched? The xid for the updated event would point to the parent entity. You’d need to subscribe to the child entity events to understand which collection entries changed unless you force consumers to do a full set compare.

clemensv: I think each event needs to be useful in itself. If you raise an event for a group that has multiple resources (like Doug's model for the xreg website), the group event would not even give you a hint where to navigate to look without giving the collection.

duglin:
ok made all updates based on today's conversations (on the call and off-line).
See what you think,

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+Notification of changes to the entities within an xRegistry instance SHOULD be
+exposed as events. These changes could be the result of an end-user interaction
+with the Registry, or due to some other (possibly internal) processing of
+the Registry data.
+
+A single interaction with the Registry MAY result in multiple events, however,
+within the scope of one interaction (see
+[`xregcorrelationid`](#xregcorrelationid-context-attribute)) the following
+constraints apply:
+- Only one of `created`, `updated` or `deleted` event MUST be generated per
```
jskeet, line 77: This sentence structure doesn't quite work for me at the moment. Can we simplify this to just "Only one event MUST be generated per `subject`?" (Then "If the interaction involved more than one action".)

It looks like the only other type is "deprecation" - if we can include that as well (always an update, effectively?) then I think this can be simplified (and the last bullet point in this list deleted).

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+
+The following sections specify the metadata defined for xRegistry events.
+Implementations MAY define additional metadata.
+
+### CloudEvent [`type`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type) Context Attribute
+
+The type of action taken on the entity. The value MUST be of the form:
+    `io.xregistry.<ENTITY>.<ACTION>`
+where:
+- `<ENTITY>` is the type of xRegistry entity that was acted upon. It MUST
```
jskeet: Possibly link this with `subject` explicitly?

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+to the root of the xRegistry instance.
+
+This metadata MUST be present in each event.
+
+### CloudEvent [`subject`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#subject) Context Attribute
+
+The `xid` of the entity acted upon. While this attribute is OPTIONAL in the
+CloudEvents specification, it is REQUIRED to be present in an xRegistry event.
+
+Note: constructing a URL by appending the `subject` value to the `source`
```
jskeet, line 127: This means that either the `source` must end with a `/` or the `subject` must start with a `/`. I can't remember offhand which is true - which means others may not either. This could be clarified with an example, or stated explicitly.

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+- `<ACTION>` is the operation performed on the entity. It MUST be one of:
+  - `created`
+  - `updated`
+  - `deprecration`
+  - `deleted`
+
+  Not all `<ACTION>` values are applicable to all entities. See the
+  [Entity Events](#entity-events) section for more information.
+
+This metadata MUST be present in each event.
```
jskeet: Attribute rather than metadata? (Here and elsewhere.)

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+- If the interaction involved more than one of those actions, then the single
+  event generated MUST be chosen in the following order of precedence:
+  `deleted`, `created`, `updated`.
+- Only one event with the same `type` and `subject` combination MUST be
+  generated.
+
+The following sections specify the metadata defined for xRegistry events.
+Implementations MAY define additional metadata.
+
+### CloudEvent [`type`](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#type) Context Attribute
```
jskeet: Possibly specify "Core Context Attribute" or "Extension Context Attribute" for each of these?

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+The `changed` attribute, when present, has the following constraints:
+- MUST be a list of the top-level attribute names (not values) that were
+  added, modified, or deleted for the `subject` entity.
+
+- In the case of changing a Resource's default Version pointer, the list MUST
+  include all attributes from the previous default Version that have non-`null`
+  values, and all attributes from the new default Version that have non-`null`
+  values. Any duplicate attribute names MUST be removed.
+
+- This attribute MUST NOT appear on the `created` or `deleted` events.
```
jskeet, line 192: Deprecation?

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+interaction, per the rules previously stated, only one `updated` event
+will be generated. This means the attribute names of all impacted attributes
+MUST be merged into one `changed` list.
+
+While `changed` is OPTIONAL, it is RECOMMENDED to be present due to its
+usefulness for consumers. However, if exposure of this information would be
+inappropriate for some scenarios then it MAY be excluded. For example, for
+privacy/security reasons.
+
+Implementations MAY define additional attributes as siblings to `changed`.
```
jskeet: Can implementations send data for created/deleted events, just not a "changed" attribute?

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+those will be noted in the [Entity Events](#entity-events) sections below.
+When they do appear they MUST use the JSONPath dot (`.`) notation to express
+the traversal to the attribute of interest.
+
+If an entity is updated for multiple reasons during the processing of an
+interaction, per the rules previously stated, only one `updated` event
+will be generated. This means the attribute names of all impacted attributes
+MUST be merged into one `changed` list.
+
+While `changed` is OPTIONAL, it is RECOMMENDED to be present due to its
```
jskeet: Except it's only allowed for updated, and not allowed even then for model or modelsource. This ends up being a pretty narrow recommendation - I think it's worth acknowledging that. Perhaps "it is RECOMMENDED to be present where permitted"?

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+### `group` Events
+
+- Action: `created`
+  - MUST be generated when a new Group is created.
+
+  - A `io.xregistry.registry.upated` event will also be generated where the
+    `<GROUPS>` and `<GROUPS>count` attributes will be included in `changed`.
+
+- Action: `updated`
+  - MUST be generated when a Group's attribute is updated, where each modified
```
jskeet, line 303: We've said elsewhere that `changed` is only recommended - now it's required in this specific situation?

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+
+    While the `<RESOURCES>` attribute is present due to the child
+    collection changing, in order to see which specific Resources were impacted
+    the `io.xregistry.resource.created` and `io.xregistry.resource.deleted`
+    events would need to be examined.
+
+- Action: `deleted`
+  - MUST be generated when a Group is deleted.
+
+  - A `io.xregistry.registry.upated` event will also be generated where the
```
jskeet:
```suggestion
  - A `io.xregistry.registry.updated` event will also be generated where the
```

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+   `<GROUPS>` and `<GROUPS>count` attributes will be included in `changed`.
+
+### `resource` Events
+
+- Action: `created`
+  - MUST be generated when a new Resource is created, where each modified
+    attribute MUST be included in `changed`.
```
jskeet: Ditto the above comment about being confused where `changed` is required vs recommended

core/events.md
jskeet: There are several occurrences of this, actually - just search for upated.

jskeet (commented):
Out of time/attention for reviewing this, but hopefully the comments so far are helpful.

core/events.md
lailabougria: It would appear on deprecation, and show the artributes that changed on the deprecation object

core/events.md
```diff
@@ -0,0 +1,670 @@
...
+
+- Action: `created`
+  - MUST be generated when a new Group is created.
+
+  - A `io.xregistry.registry.upated` event will also be generated where the
+    `<GROUPS>` and `<GROUPS>count` attributes will be included in `changed`.
+
+- Action: `updated`
+  - MUST be generated when a Group's attribute is updated, where each modified
+    attribute MUST be included in `changed`.
```
lailabougria: Agreed, changed isn’t required

core/events.md
duglin: the problem is that it's ok to have "deprecation" at the same time as "created" and "updated". Not sure about "deleted" though.  But for those  3 (created,updated,deleted) it made no sense to ever see more than one to me.

core/events.md
duglin: not sure how I feel about it, but I added it for now.

core/events.md
duglin: done

core/events.md
duglin: I was trying to keep it more abstract in case another serialization didn't have "attributes", but for now I'll make the changed. We can revert it when we get our 2nd serialization format.

duglin: actually, I take that back. I think "metadata" is more appropriate because these would be HTTP headers (not attributes) in the binary CE case. I guess I could say "context attribute" though... let me try that

core/events.md
duglin: XIDs are required to start with "/". I'm not sure subject is banned from ending with "/" and I'm not even sure if "//" is illegal in URL paths (RFC seems to allow it). I'll try to make it a bit clearer though

core/events.md
duglin: done

core/events.md
duglin: yes. I'll tweaks this sentence to make that clearer.

core/events.md
duglin: added "if present" to make it clearer

core/events.md
duglin: got 'em!

core/events.md
duglin: added "if present" to all mentions of "changed"

core/events.md
duglin: we agreed to include them in 'changed' but make it clear that people need to go to the other events to see which specific child entity was added/removed

PR #384, opened 2025-07-08
Changes for week of July 7

duglin:
commit 1
- more minor edits from reviews

commit 2
- no way to discover namecharset at this time, just docs
- ximportresources are now transitive

commit 3
- make endpoints.usage an array + strict enum

commit 4
- more tweaks

From last week:
- Jon: can client provide the xregcorrelationid on the request?  Decided not to support this at this time. Perhaps if we had a better idea of the usecase that need it.
- Jon: Why not just add a "deprecation" extension to data and let the subscribe() filter on it - instead of a dedicated "deprecation" action? Keep as is since we can't assume all subscription managers support filtering and we really really don't want to filter on "data"

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -635,7 +635,7 @@ sensitivity rules in the specification.
 
 Attribute names and key names are limited to 63 characters, so why are some
 Group and Resource names limited to less? Because when they appear as part of
-attribute names and they can be append with phrase like `url`, `count`
+attribute names and they can be append with phrases like `url`, `count`
 or `base64`, and they still have to fit within the 63-character limit, and so
 we need to take into account the length of those phrases.
 
--- a/core/spec.md
+++ b/core/spec.md
@@ -165,7 +165,7 @@ pseudo JSON form:
         } ?
 
         "ifvalues": {                   # If "type" is scalar
-          "<VALUE>": {                  # Possible attribute value
+          "<STRING>": {                 # Possible attribute value
             "siblingattributes": { ... } # See "attributes" above
           } *
         } ?
@@ -381,6 +381,7 @@ information about each data type):
 - `<XIDTYPE>`
 - `<TYPE>` - one of the allowable data type names (MUST be in lower case)
   listed in [Attributes and Extensions](#attributes-and-extensions)
+- `<VALUE>` - an instance of one of the above data types
 
 ### Terminology
 
@@ -2501,7 +2502,7 @@ The overall format of a model definition is as follows:
       } ?
 
       "ifvalues": {                    # If "type" is scalar
-        "<VALUE>": {
+        "<STRING>": {
           "siblingattributes": { ... } # Siblings to this "attribute"
         } *
       } ?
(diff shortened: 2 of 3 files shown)


core/primer.md
```diff
@@ -631,13 +631,13 @@ sensitivity rules in the specification.
...
+# Why the non-63 character limit on some Group and Resource names?
 
-Attribute names and key names are limited to 63 characters, so why are
-Group and Resource names limited to only 58? Because when they appear as
-Collection attributes and we append `url` or `count` to them they still have
-to fit within the 63-character limit, and so we need to take into account
-the length of the word `count`.
+Attribute names and key names are limited to 63 characters, so why are some
+Group and Resource names limited to less? Because when they appear as part of
+attribute names and they can be append with phrase like `url`, `count`
```
jskeet: phrase => phrases

core/spec.md
```diff
@@ -2897,15 +2895,15 @@ The following describes the attributes of Registry model:
...
+##### - Model: `groups."<STRING>".compatiblewith`
   - Type: URI.
   - OPTIONAL.
   - References / represents an xRegistry model definition that
     the Group type is compatible with. This is meant to express
     interoperability between models in different xRegistries via using a
     shared compatible model.
   - Does not imply runtime validation of the claim.
-  - Example: `https://raw.githubusercontent.com/xregistry/spec/refs/heads/main/schema/model.json`
+  - Example: `http://xregistry.io/xreg/xregistryspecs/schema-v1/docs/model.json`
```
jskeet: Is this deliberately http now instead of https?

core/spec.md
```diff
@@ -3659,6 +3659,9 @@ where:
   combination defined within the same Registry. It MUST NOT reference the
   same Group under which the `ximportresources` resides.
 - An empty array MAY be specified, implying no Resources are imported.
+- The use of `ximportresources` MAY be transitive as long as it does not
```
jskeet: This "MAY" feels ambiguous to me. Are we saying that the server may support transitive imports but doesn't have to, or that a particular usage may be transitive but it's valid to import a resource that doesn't import anything itself? If it's the latter, I think it would be better to say "The use of `ximportresources` MUST be handled in a transitive fashion [...]" or similar - to make it clear that this is really a requirement of the server.

core/primer.md
duglin: done

core/spec.md
duglin: nope - made it https again

core/spec.md
duglin: Is this any better?
```
- The definition of a Group MAY include an `ximportresources` directive
  that references a Resource from another Group that itself is defined
  via an `ximportresources`. However, transitive definitions of Resources
  MUST NOT result in a circular import chain.
```

jskeet: Yes, I think so. Hard to say exactly why, but it feels like it's leaving less wiggle room for servers.

jskeet: (Although maybe we should say that a server MUST spot circular import chains and report an error in that case.)

duglin: I did think about that but decided against it because we don't go out of our way to say that for all "the client MAY" type of statements. I think it should be obvious.

duglin:
Approved on 7/10 call

PR #396, opened 2025-07-25
Week of July 28

duglin:
- More tweaks
- Explain why some constraints might not be enforced when using xref or basemessage
- treat unauthorized xref references the same as "dangling" references
- add "deprecated" as a group level attribute too. Now it's a "common" attribute
- make envelope/format/... and ifvalue case-insensitive

fixes: #398 
fixes: #397

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -615,6 +615,16 @@ to be met while maintaining the interoperability goals of the specification.
 Implementations are encouraged to contact the xRegistry community if it is
 unclear if certain customizations would violate the specification.
 
+Implementations MAY (but are NOT REQUIRED) to validate cross-entity
+constraints that might be violated due to changes in a referenced entity.
+For example, [Endpoint's `envelope`](../endpoint/spec.md#envelope) attribute
+mandates that all Messages in that Endpoint use that same `envelope` value.
+One of those Messages might have a `basemessage` value that points to a
+ Message that breaks that rule. For a variety of reasons (e.g. authorization
+ constraints), server implementations might not be able to verify this
+ constraint. Likewise, the same situation might occur via the use of `xref`.
+Clients need to be aware of these possibilities.
+
 ### Attributes and Extensions
 
 Unless otherwise noted, all attributes and extensions MUST be mutable and MUST
@@ -654,9 +664,10 @@ be one of the following data types:
   information.  Its value MUST start with a `/`.
 - `xidtype` - MUST be a URL reference to an xRegistry model type. The
    reference MUST point to one of: the Registry itself (`/`), a Group type
-   (`/<GROUPS>`), a Resource type (`/<GROUPS>/<RESOURCE>`) or Version type for
+   (`/<GROUPS>`), a Resource type (`/<GROUPS>/<RESOURCES>`) or Version type for
    a Resource (`/<GROUPS>/<RESOURCES>/versions`). Its value MUST reference a
-   defined/valid type in the Registry.
+   defined/valid type in the Registry. It MUST use the plural name of the
+   referenced type, if it is a Group, Resource or Version.
 - `string` - sequence of Unicode characters.
 - `timestamp` - an [RFC3339](https://tools.ietf.org/html/rfc3339) timestamp.
   Use of a `time-zone` notation is RECOMMENDED. All timestamps returned by
--- a/core/spec.md
+++ b/core/spec.md
@@ -1524,7 +1524,7 @@ semantics defined above with the following exceptions:
     operations, any missing REQUIRED attributes MUST generate an error
     ([required_attribute_missing](#required_attribute_missing)).
 
-The `PATCH`, variant when directed at an xRegistry collection, MUST adhere to
+The `PATCH` variant when directed at an xRegistry collection, MUST adhere to
 the following:
   - The HTTP body MUST contain a JSON map where the key MUST be the
     `<SINGULAR>id` of each entity in the map. Note, that in the case of a map
--- a/core/spec.md
+++ b/core/spec.md
@@ -229,6 +229,12 @@ pseudo JSON form:
       "labels": { "<STRING>": "<STRING>" * }, ?
       "createdat": "<TIMESTAMP>",
       "modifiedat": "<TIMESTAMP>",
+      "deprecated": {
+        "effective": "<TIMESTAMP>", ?
+        "removal": "<TIMESTAMP>", ?
+        "alternative": "<URL>", ?
+        "documentation": "<URL>"?
+      }, ?
 
       # Repeat for each Resource type in the Group
       "<RESOURCES>url": "<URL>",                   # e.g. "messagesurl"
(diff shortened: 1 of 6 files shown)


core/spec.md
```diff
@@ -934,6 +939,75 @@ of the existing entity. Then the existing entity would be deleted.
...
+  The `xbase` attribute MAY only be used on Group, Resource or Version
+  instances and MUST reference an entity of the same model type as the
+  current entity.
+
+  The act of inheriting another entity entails having the referenced entity's
+  attributes visible as part of the current entity. This only applies to
+  attributes at the top-level of the entity, and each of those attributes
+  MUST be visible in its entirety.
+
+  Any attributes in the target entity that are also present in the current
```
duglin: do bottoms up instead of top down.
Follow all xbases, and then work backwards overlaying each up you go up.

core/spec.md
```diff
@@ -934,6 +939,80 @@ of the existing entity. Then the existing entity would be deleted.
...
+  instances and MUST reference an entity of the same model type as the
+  current entity.
+
+  The semantics of "inheriting" is defined as follows:
+  - Starting with an entity that has `xbase` defined, follow its value to
+    the referenced entity. Repeat as necessary until the end of the `xbase`
+    chain of entities is reached. Note that the chain MUST NOT be recursive.
+  - Using the "deepest" entity's attributes as a starting point, walk back up
+    the chain merging the next entity's attributes over the current set. Note
+    that an attribute with a value of `null`, or a scalar value, MUST overlay
```
jskeet: I think the null handling will need to be clearer here. Is this meant to say that `"a": null` effectively clears the value of `a` that would otherwise be inherited?

If so, I suspect this will be quite hard to implement in some cases - the difference between "there's no value" and "there's an explicit null value" can be hard to track. We have this problem in protobuf in various situations - basically the encoding you want to use for a sort of patch operation is not the same as the encoding you want to use for "here's the information".

Maybe it's all fine, but it makes me nervous - and it's definitely not clear to me what "MUST overlay" means for null. (Do we know what it means for arrays and dictionaries? If "merging" for an array normally means "append these values to the end" is there a way of saying "no, replace the whole list"? Again, this is territory that is making me shudder thinking about protobuf and API definitions...)

core/spec.md
```diff
@@ -934,6 +939,80 @@ of the existing entity. Then the existing entity would be deleted.
...
+    the deeper entity's value, including any nested attributes. For example,
+    `"a": 123` would replace all of `"a": { "b": c" }`, assuming it is
+    model-legal for "a" to support varying types.
+
+  If a referenced entity can not be found then an error MUST NOT be generated
```
jskeet: That's definitely not what I'd expect. If it's really the desired behavior, we should explain why. Most of the time if you reference something, and that affects other aspects, then the reference not being present would cause an error, right?

core/spec.md
```diff
@@ -934,6 +939,80 @@ of the existing entity. Then the existing entity would be deleted.
...
+  to be aware of this possibility.
+
+  Normally, write operations with attributes set to `null` are treated as a
+  request to delete those attributes from the current entity. As a result they
+  will likely not appear in the persistent store, and would not appear in the
+  serialization of that entity. However, when `xbase` is defined these `null`
+  properties MUST be processed (and likely persisted) such that the server can
+  distinguish a `null` attribute from a "missing" attribute to ensure that a
+  `null` attribute will delete any matching inherited attribute, while a
+  `missing` attribute will not.
```
```suggestion
  "missing" attribute will not.
```

core/spec.md
jskeet: (I see you go into this a bit later on. But my "here be dragons" feeling still stands.)

message/spec.md
```diff
@@ -183,14 +183,15 @@ A similar transport protocol-independent message metadata convention is, for
...
 an extension.
 
-This specification uses **protocol** to refer to a transport protocol-specific
-message metadata convention. When a known protocol is explicitly specified for
-a message definition, the "protocoloptions" section MAY contain constraints
-for the protocol-specific metadata.
+This specification uses **protocol** as a selector into the protocol-specific
+message metadata that is defined under the `protocol.ifvalues` section of the
+model. When a known protocol is explicitly specified for a message definition,
+the "protocoloptions" section MAY contain constraints for the
```
jskeet: Should protocoloptions be in backticks?

jskeet (commented):
The introduction of xbase feels like much more than a "tweak" to me.

I'm really nervous about it - particularly introducing it this close to 1.0. The value of the feature is clear - but I suspect the cost over time (both in the spec and implementations).

Perhaps xbase should at least be an optional feature for a registry?

My experience in a few other projects has been that things like this sound fine, "we'll hash out the finer details later" - and then it's a never-ending pit of awfulness, and I wish we'd stuck to a much simpler but less comprehensive feature.

That said, as everyone knows I'm really only a bystander for xRegistry. This is a warning to your future selves more than anything else. I'd love nothing more than to look back on these comments in 10 years and hear "nah, that wasn't a problem at all."

jskeet:
> @jskeet if we pushed all of this logic onto the client (ALL clients) would change change your "there be dragons" concern?

I think I'd need to think about that more - but it may well just move them rather than remove them. Will ponder.

core/spec.md

message/spec.md
duglin: I think you're right. Done

core/spec.md
```diff
@@ -615,6 +615,16 @@ to be met while maintaining the interoperability goals of the specification.
...
 
+Implementations MAY (but are NOT REQUIRED) to validate cross-entity
+constraints that might be violated due to changes in a referenced entity.
+For example, [Endpoint's `envelope`](../endpoint/spec.md#envelope) attribute
+mandates that all Messages in that Endpoint use that same `envelope` value.
+One of those Messages might have a `basemessage` value that points to a
+ Message that breaks that rule. For a variety of reasons (e.g. authorization
+ constraints), server implementations might not be able to verify this
+ constraint. Likewise, the same situation might occur via the use of `xref`.
+Clients need to be aware of these possibilities.
```
lailabougria: there's some weird spacing going on here

message/spec.md
```diff
@@ -309,16 +308,33 @@ the core xRegistry Resource
...
+  Referenced messages MAY themselves have a `basemessage` value, however,
+  the chain of messages MUST NOT be recursive.
+
+  The processing that a client SHOULD take to materialize the message is
+  to follow the `basemessage` attributes to the end of the chain of messages,
+  get that message's attributes, and then walk back up the chain performing a
+  "merge" operation of the next message's attributes. Note in the case of an
+  inherited attribute being a complex type (e.g. map, object), and the
+  overlaying attribute is scalar, then the entire inherited attribute (and
+  nested values) are replace by that scalar value.
```
lailabougria:
```suggestion
  nested values) are replaced by that scalar value.
```

core/spec.md

message/spec.md
duglin: done

jskeet:
> to prevent too many merge conflicts we're merging this one w/o a review from the full CE team. @jskeet if you happen to see anything odd please let us know.

I'll have another look when I get a chance, but I'm sure it's fine.

PR #400, opened 2025-08-05
Schema spec revision

clemensv:

New Intro

### DIFF
--- a/schema/spec.md
+++ b/schema/spec.md
@@ -36,7 +36,7 @@ allows for the storage, management and discovery of schema documents.
 ## 1. Overview
 
 A schema registry provides a repository for managing serialization, validation,
-and data type definitions schemas as they are commonly used in distributed
+and data type definition schemas as they are commonly used in distributed
 systems. Common schema formats include JSON Schema, JSON Structure, Apache Avro
 Schema, Google Protobuf Schema, and XML Schema. However, this specification does
 not mandate, or limit, which schema formats are used.
@@ -322,7 +322,7 @@ Every schema (i.e. the schema Resource) MUST reside inside a Schema Group.
 
 Example:
 
-The follow abbreviated Schema Registry's contents shows a single Schema Group
+The follow abbreviated Schema Registry's content shows a single Schema Group
 containing 5 schemas.
 
 ```yaml
@@ -602,7 +602,7 @@ Examples:
 `#TelemetryEvent` references the record declaration of the `TelemetryEvent`
 record.
 - If the Avro schema document is a local Schema Registry reference like
-`#/schemagroups/com.example.telemetry/schemas/com.example.telemetrydata`, in the
+`#/schemagroups/com.example.telemetry/schemas/com.example.telemetrydata`, in
 which the reference is already in the form of a URI fragment, the suffix is
 appended separated with a colon, for instance
 `.../com.example.telemetrydata:TelemetryEvent`.


schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+data.
+
+Formats like XML and JSON do not require a schema to decode the data, but
+schemas are still very useful to establish a common understanding of the data
+structures that are exchanged, provide a foundation for code generation, allow
+for validation of the data, and provide an anchor for documentation and semantic
+information, like scientific units for numeric values, that goes beyond simple
+labels and data types. Generally, it is a best practice for data structures that
+are exchanged in a distributed systems ought to be described by a schema, even
+if the data serialization model does not require one.
```
lailabougria:
```suggestion
labels and data types. Generally, it is a best practice for data structures that
are exchanged in a distributed system to be described by a schema, even
if the data serialization model does not require one.
```

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+
+### 1.2. Schema References
+
+In the [CloudEvents][CloudEvents dataschema] specification, the `dataschema`
+attribute holds a URI and is specifically thought to reference a schema document
```
lailabougria:
```suggestion
attribute holds a URI and is specifically meant to reference a schema document
```

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+    "data_base64": "...base64-encoded-data..."
+}
+```
 
-## Overview
+Since this URL might be a bit long, the xRegistry core specification allows for
+an implementation to provide an alternative, shorter, self-referencing URL that
+points to the same schema version, via the [`shortself`][xRegistry shortself]
+attribute. The specification is not prescriptive about the format of the shorter
+URL, but it might follow common URL-shorneter practices. With that, the above
```
lailabougria:
```suggestion
URL, but it might follow common URL-shortening practices. With that, the above
```

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+perform live updates. Once data has been published, data serialized based on
+several different versions may exist in a system, in queues, in databases, or in
+files.
+
+The schema registry therefore allows managing multiple versions of schemas,
+declare their lineage, and state the compatibility policy. The compatibility
+policy is used to determine whether a schema change is compatible with existing
+data, and MAY be enforced by implementations of the schema registry. For this,
+this specification leans on the [xRegistry Core][xRegistry Core] specification
+that already defines these versioning mechanisms for any kind of resource.
```
lailabougria:
```suggestion
that already defines these versioning and compatibility mechanisms for any kind of resource.
```

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+
+### 1.4. Document Store
+
+The schema registry is a document store and therefore also has the
+[`hasdocument`][xRegistry hasdocument] attribute defined in the xRegistry Core
+attribute (implicitly) defined as `true` for the `schema` Resource.
+
+What this means is that the schema registry immediately yields a document with
```
lailabougria:
```suggestion
This means that the schema registry immediately yields a document with
```

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+the stored content-type when the a client issues a GET request to the
```
lailabougria:
```suggestion
the stored content-type when a client issues a GET request to the
```

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+
+To access the metadata of the `schema` or the schema version as a JSON document,
+the client can append a `$details`suffix to the URL, like
+`https://example.com/schemagroups/com.example.schemas/schemas/com.example.event/versions/1.0$details`
+or
+`https://example.com/schemagroups/com.example.schemas/schemas/com.example.event$details`.
+
+Beyopnd this, the [xRegistry Core][xRegistry Core] specification provides rich
+filtering and export/import capabilities, which can be used to retrieve schema
+documents in bulk, or to export/import schemas and schema versions in a
```
lailabougria:
```suggestion
documents in bulk, or to export/import schemas and schema Versions in a
```

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+### 2.2. Terminology
+
+This specification defines the following terms:
+
+#### 2.2.1. Schema
+
+We use the term **schema** (or schema Resource) in this specification as a
+logical grouping of **schema Versions**. A **schema Version** is a concrete
+document. The **schema** Resource is a semantic umbrella formed around one or
+more concrete schema Version documents. Per the definition of the
```
lailabougria:
```suggestion
document. The **schema** Resource is a semantic umbrella formed around one or
more concrete schema Version documents that represent iterations of the same logical schema. 
Per the definition of the
```

I meant to clarify that they are logically the same schema although they can look quite different across versions

schema/spec.md
```diff
@@ -3,23 +3,212 @@
...
+[`compatibility`][xRegistry compatibility] attribute, all Versions of a single
+**schema** MUST adhere to the rules defined by the `compatibility` attribute.
+Any breaking change MUST result in a new **schema** Resource being created.
+
+In terms of versioning, you can think of a **schema** as a collection of
+versions that are compatible according to the selected `compatibility` mode.
+When that compatibility is broken across versions, a completely new **schema**
+MUST be created, to indicate the breaking change.
```
lailabougria:
```suggestion
MUST be created, to indicate the breaking change. The [`deprecated`](xRegistry deprecated)
attribute may be used to indicate the appropriate new schema to use following a breaking change.
```

schema/spec.md
```diff
@@ -545,4 +645,35 @@ Examples:
...
+[xRegistry Core]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html
+[xRegistry self]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#self-attribute
+[xRegistry shortself]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#shortself-attribute
+[xRegistry compatibility]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#compatibility-attribute
+[xRegistry version-ids]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#version-ids
+[xRegistry attributes-and-extensions]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#attributes-and-extensions
+[xRegistry ancestor]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#ancestor-attribute
+[xRegistry hasdocument]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#--model-groupsstringresourcesstringhasdocument
+[xRegistry default-version]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#registry-design
+[xRegistry pagination]: https://xregistry.io/xreg/xregistryspecs/pagination-v1/docs/spec.html
```
lailabougria, line 668:
```suggestion
[xRegistry pagination]: https://xregistry.io/xreg/xregistryspecs/pagination-v1/docs/spec.html
[xRegistry deprecated]: https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#deprecated
```

schema/spec.md
```diff
@@ -246,31 +336,35 @@ containing 5 schemas.
...
+
+This specification defines "compatibility" for schemas as follows; version B of
+a schema is said to be compatible with version A of a schema if all of the
+following are true:
+
 - Any document that adheres to the rules specified by schema A also adheres to
   rules specified by schema B.
 - Any processing rules defined for schema A also apply for schema B.
 - Any processing rules defined for schema B, that are not defined for schema
   A, do not conflict with the processing rules for schema A.
```
lailabougria: I know this isn't part of the change in this PR, but catching it now. This seems to be at least somewhat conflicting with what the [compat attribute docs state](https://xregistry.io/xreg/xregistryspecs/core-v1/docs/spec.html#compatibility-attribute):

> This specification makes no statement as to which parts of the data are examined for compatibility (e.g. xRegistry metadata, domain-specific document, etc.). This SHOULD be defined by the compatibility values. The exact meaning of what each compatibility value means might vary based on the data model of the Resource, therefore this specification only defines a very high-level abstract meaning for each to ensure some degree of consistency.

I think we need to remove this paragraph, because we've made no statement in the Core about what the individual compatibility modes mean, and therefore, summarizing the rules as such might not be accurate.

schema/spec.md
```diff
@@ -282,11 +376,17 @@ major version identifier in the `schemaid`, like `"com.example.event.v1"` or
 schemas can be more easily identified by users and developers. The schema
 `versionid` then functions as the semantic minor version identifier.
 
+Version lineage is defined by the [`ancestor`][xRegistry ancestor] attribute,
+which is a `versionid` of the Version that this Version is based on. The `ancestor`
+attribute permits multiple version branches to exist, and allows for
+implementations to determine the Version lineage. See the
```
lailabougria:
```suggestion
which is a `versionid` of the Version that this Version is based on. The `ancestor`
attribute permits multiple Versions to reference the same ancestor, and allows for
implementations to determine the Version's ancestor. See the
```

schema/spec.md
```diff
@@ -545,4 +645,35 @@ Examples:
   is appended separated with a colon, for instance
   `.../com.example.telemetrydata:TelemetryEvent`.
 
+### 5. Security Considerations
+
+Like [xRegistry Core][xRegistry Core] specification, this specification does not
```
lailabougria:
```suggestion
Like the [xRegistry Core][xRegistry Core] specification, this specification does not
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+    - [4.3. Schema Formats](#43-schema-formats)
+      - [4.3.1. JSON Schema](#431-json-schema)
+      - [4.3.2. XML Schema](#432-xml-schema)
+      - [4.3.3. Apache Avro Schema](#433-apache-avro-schema)
+      - [4.3.4. Protobuf Schema](#434-protobuf-schema)
+    - [5. Security Considerations](#5-security-considerations)
+
+## 1. Overview
+
+A schema registry provides a respository for managing serialization and
```
duglin:
```suggestion
A schema registry provides a respository for managing serialization,
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+validation and data type definitions schemas as they are commonly used in
+distributed systems. Common schema formats include JSON Schema, JSON
+Structure, Apache Avro Schema, Google Protobuf Schema, and XML Schema.
```
duglin: `However, this specification does not mandate, or limit, which schema formats are used.`

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+
+### 1.3. Versioning
+
+When schemas are used in a system, they typically evolve over time. Data
+structures are extended or modified, with parts added or deprecated or even
+removed. Some of these changes are compatible with existing data, while others
+are not.
+
+Serialization generally occurs based on a specific schema version that the data
+publisher uses. Multiple versions of publishers may exist in the same system,
```
duglin:
```suggestion
publisher uses. Multiple versions of publishers might exist in the same system,
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+using different schema versions, which is a common occurrence in systems that
+perform live updates. Once data has been published, data serialized based on
+several different versions may exist in a system, in queues, in databases, or in
```
duglin:
```suggestion
several different versions might exist in a system, in queues, in databases, or in
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+files.
+
+The schema registry therefore allows managing multiple versions of schemas,
+declare their lineage, and state the compatibility policy. The compatibility
```
duglin:
```suggestion
declares their lineage, and state their compatibility policy. The compatibility
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+policy is used to determine whether a schema change is compatible with existing
+data, and MAY be enforced by implementations of the schema registry. For this,
```
duglin: "data" -while true, if feels a bit indirect. Would it be better to talk about how it enforces compat between versions of the schema instead? "existing data" might not exist, so it's not really the thing we're focused on.

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+this specification leans on the [xRegistry Core][xRegistry Core] specification
+that already defines these versioning and compatibility mechanisms for any kind of resource.
```
duglin: the core spec doesn't actually fully define the compat stuff, just what the terms mean at a high level. It feels like we should say something about how impls will need to be more explicit for each schema format and compat mode, no?

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+
+### 1.4. Document Store
+
+The schema registry is a document store and therefore also has the
```
duglin:
```suggestion
The schema registry is a document store and therefore has the
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+[`hasdocument`][xRegistry hasdocument] attribute defined in the xRegistry Core
+attribute (implicitly) defined as `true` for the `schema` Resource.
+
+This means that the schema registry immediately yields a document with
+the stored content-type when a client issues a GET request to the
+[`self`][xRegistry self] URL of a schema Version. The associated metadata is
+returned in the HTTP headers. The [default version][xRegistry default-version]
+of the schema Version is returned when the client issues a GET request to the
```
duglin: I think this would read better if we removed the extra "Version" from "schema Version"? I don't think we've introduced that term yet and it sounds a bit off to talk about a "version of a version". W/o it we just talk about the "default version of a Resource", which is easier to grok.

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+[`self`][xRegistry self] URL of the `schema` Resource.
+
+This allows to provide external parties with a link that they can use without
```
duglin:
```suggestion
This enables the ability to provide external parties with a link that they can use without
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+needing to know any details about xRegistry.
+
+Storing a new schema is similarly straightforward for clients that do not know
+xRegistry specifics, by simply using a POST against [`self`][xRegistry self] URL
+of the `schema` Resource in the simplest case.
+
+To access the metadata of the `schema` or the schema version as a JSON document,
```
duglin:
```suggestion
To access the metadata of the `schema`, or the schema version as a JSON document,
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+the client can append a `$details`suffix to the URL, like
```
duglin:
```suggestion
the client can append a `$details` suffix to the URL, like
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+document. The **schema** Resource is a semantic umbrella formed around one or
+more concrete schema Version documents that represent iterations of the same logical schema.
+Per the definition of the
+[`compatibility`][xRegistry compatibility] attribute, all Versions of a single
+**schema** MUST adhere to the rules defined by the `compatibility` attribute.
+Any breaking change MUST result in a new **schema** Resource being created.
+
+In terms of versioning, you can think of a **schema** as a collection of
+versions that are compatible according to the selected `compatibility` mode.
+MUST be created, to indicate the breaking change. The [`deprecated`](xRegistry deprecated)
```
duglin: I think the first part of this line ("MUST be created, to indicate the breaking change") is old and not needed given the previous paragraph.

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+attribute may be used to indicate the appropriate new schema to use following a breaking change.
```
duglin:
```suggestion
attribute MAY be used to indicate the appropriate new schema to use following a breaking change.
```

schema/spec.md
```diff
@@ -3,23 +3,214 @@
...
+MUST be created, to indicate the breaking change.
```
duglin: Left over from a previous edit/version, I think.

schema/spec.md
```diff
@@ -99,132 +290,33 @@ this form:
...
-      ]
-    }
-  ]
-}
-```
-
 Implementations of this specification MAY include additional extension
 attributes, including the `*` attribute of type `any`.
 
 Since the Schema Registry is an application of the xRegistry specification, all
```
duglin: let's link to the core spec here - from the "xRegistry specification" text

schema/spec.md
```diff
@@ -99,132 +290,33 @@ this form:
...
+A schema group is a collection of schemas that are related to each other in
+some application-defined way. A Schema Group does not impose any restrictions
+on the contained schemas, meaning that a Schema Group MAY contain schemas of
+different formats.
 
-A schema Group is a collection of schemas that are related to each other in
-some application-defined way. A schema Group does not impose any restrictions
-on the contained schemas, meaning that a schema Group MAY contain schemas of
-different formats. Every schema MUST reside inside a schema Group.
+Every schema MUST reside inside a Schema Group.
```
duglin: It might be me, but trying to read this as a newbie ,someone may think that a schema can be part of a Group directly. So, perhaps a tiny bit of clarity:
`Every schema (i.e. schema Resource) MUST reside inside a Schema Group`

schema/spec.md
duglin: Just to make sure I understand what you're saying.... you think the compat definitions needs to be related to each schema format, right?

schema/spec.md
```diff
@@ -281,12 +377,18 @@ major version identifier in the `schemaid`, like `"com.example.event.v1"` or
 `"com.example.event.2024-02"`, so that incompatible, but historically related
 schemas can be more easily identified by users and developers. The schema
 `versionid` then functions as the semantic minor version identifier.
+which is a `versionid` of the Version that this Version is based on. The `ancestor`
```
duglin: something is missing before this line.

schema/spec.md
```diff
@@ -281,12 +377,18 @@ major version identifier in the `schemaid`, like `"com.example.event.v1"` or
...
+attribute permits multiple Versions to reference the same ancestor, and allows for
+implementations to determine the Version's ancestor. See the
```
duglin, line 382: s/See/Since/ ??

duglin: s/See/Since/ ?
But even then the sentence doesn't parse correctly. I think some tweaking is needed.

schema/spec.md
```diff
@@ -307,13 +409,13 @@ the core xRegistry Resource
...
 
-#### `format`
+#### 4.2.2. `format`
 
 - Type: String
 - Description: Identifies the schema format. In absence of formal media-type
   definitions for several important schema formats, we define a convention here
   to reference schema formats by name and version as `<NAME>/<VERSION>`. This
-  specification defines a set of common [schema format names](#schema-formats)
+  specification defines a set of common [schema format names](#43-schema-formats)
```
duglin, line 418: While being able to point users to a section # is handy, one downside to having numbers in section titles is that if we ever renumber things ALL docs (especially ones we don't control) will now have bad links - even if the section in question was just moved. Something to think about.

schema/spec.md
```diff
@@ -545,4 +647,36 @@ Examples:
   is appended separated with a colon, for instance
   `.../com.example.telemetrydata:TelemetryEvent`.
 
+Like the [xRegistry Core][xRegistry Core] specification, this specification does not
```
duglin: remove? Dup of next para

schema/spec.md
```diff
@@ -545,4 +647,36 @@ Examples:
...
+
+Like [xRegistry Core][xRegistry Core] specification, this specification does not
+explicitly address authentication or authorization levels of users, nor how to
+securely protect the APIs.
+
+It is expected that any implementation of this specification will use
+authentication and authorization mechanisms that are appropriate for the
+application domain and the deployment environment. This may include, but is not
```
duglin:
```suggestion
application domain and the deployment environment. This MAY include, but is not
```

duglin:
rebase is needed

schema/spec.md
lailabougria: To your question, yes, but also, we have not defined in this spec what each compatibility mode means, so either way it should be removed.

schema/spec.md
duglin: I think we need to make it clear we support more than just the "common" ones

schema/spec.md
```diff
@@ -3,23 +3,213 @@
...
+are exchanged in a distributed system to be described by a schema, even
+if the data serialization model does not require one.
+
+### 1.2. Schema References
+
+In the [CloudEvents][CloudEvents dataschema] specification, the `dataschema`
+attribute holds a URI and is specifically meant to reference a schema document
+residing in a registry. For example, a CloudEvent with a `dataschema` attribute
+pointing to a schema version in a schema registry might look like this, using
+the schema version's [`self`][xRegistry self] URL as the value of the `dataschema` attribute:
```
duglin: can you have your editor wrap lines like this at 80?

schema/spec.md
```diff
@@ -3,23 +3,213 @@
...
+When schemas are used in a system, they typically evolve over time. Data
+structures are extended or modified, with parts added or deprecated or even
+removed. Some of these changes are compatible with existing data, while others
+are not.
+
+Serialization generally occurs based on a specific schema version that the data
+publisher uses. Multiple versions of publishers MAY exist in the same system,
+using different schema versions, which is a common occurrence in systems that
+perform live updates. Once data has been published, data serialized based on
+several different versions MAY exist in a system, in queues, in databases, or in
```
duglin:
```suggestion
several different versions might exist in a system, in queues, in databases, or in
```

schema/spec.md
```diff
@@ -3,23 +3,213 @@
...
+the stored content-type when a client issues a GET request to the
+[`self`][xRegistry self] URL of a schema Version. The associated metadata is
+returned in the HTTP headers. The [default version][xRegistry default-version]
+of the schema Version is returned when the client issues a GET request to the
+[`self`][xRegistry self] URL of the `schema` Resource.
+
+This allows to provide external parties with a link that they can use without
+needing to know any details about xRegistry.
+
+Storing a new schema is similarly straightforward for clients that do not know
```
duglin:
```suggestion
Storing a new Version of a schema is similarly straightforward for clients that do not know
```

schema/spec.md
```diff
@@ -100,83 +290,33 @@ this form:
...
-In the pseudo JSON format snippets `?` means the preceding attribute is
-OPTIONAL, `*` means the preceding attribute MAY appear zero or more times,
-and `+` means the preceding attribute MUST appear at least once. The presence
-of the `#` character means the remaining portion of the line is a comment.
-Whitespace characters in the JSON snippets are used for readability and are
-not normative.
+### 4.1. Schema Groups
 
-### Terminology
+The Group (`<GROUP>`) name for the Schema Registry is `schemagroups`. The
```
duglin: this is switched up: GROUP->schemagroup, GROUPS->schemagroups.
You may want to explicitly mention one "singular" vs "plural", not just the plural name itself

schema/spec.md
```diff
@@ -198,32 +338,35 @@ containing 5 schemas.
 }
 ```
 
-### Schema Resources
+### 4.2. Schema Resources
+
+The Resources (`<RESOURCE>`) collection inside of Schema Groups is named
```
duglin: Sam here - talk about singular vs plural - the current text is a bit mixed up

schema/spec.md
```diff
@@ -198,32 +338,35 @@ containing 5 schemas.
...
+`schemas`. The type of the Resource is `schema`. Any single `schema` is a
+container for one or more `versions`, which hold the concrete schema
+documents or schema document references.
 
-The Resource plural name (`<RESOURCES>`) is `schemas`, and the Resource
-singular name (`<RESOURCE>`) is `schema`.
+All Versions of a Schema MUST adhere to the semantic rules of the schema's
```
duglin:
```suggestion
All Versions of a single Schema Resource MUST adhere to the semantic rules of the schema's
```

schema/spec.md
```diff
@@ -3,23 +3,218 @@
...
+    - [4.3. Schema Formats](#43-schema-formats)
+      - [4.3.1. JSON Schema](#431-json-schema)
+      - [4.3.2. XML Schema](#432-xml-schema)
+      - [4.3.3. Apache Avro Schema](#433-apache-avro-schema)
+      - [4.3.4. Protobuf Schema](#434-protobuf-schema)
+
+## 1. Overview
+
+A schema registry provides a repository for managing serialization, validation,
+and data type definitions schemas as they are commonly used in distributed
```
deissnerk:
```suggestion
and data type definition schemas as they are commonly used in distributed
```

schema/spec.md
```diff
@@ -100,83 +295,34 @@ this form:
...
-A schema Group is a collection of schemas that are related to each other in
-some application-defined way. A schema Group does not impose any restrictions
-on the contained schemas, meaning that a schema Group MAY contain schemas of
-different formats. Every schema Resource MUST reside inside a schema Group.
+Every schema (i.e. the schema Resource) MUST reside inside a Schema Group.
 
 Example:
 
-The follow abbreviated Schema Registry's contents shows a single schema Group
+The follow abbreviated Schema Registry's contents shows a single Schema Group
```
deissnerk:
```suggestion
The follow abbreviated Schema Registry's content shows a single Schema Group
```

schema/spec.md
```diff
@@ -462,16 +602,16 @@ Examples:
 `#TelemetryEvent` references the record declaration of the `TelemetryEvent`
 record.
 - If the Avro schema document is a local Schema Registry reference like
-`#/schemagroups/com.example.telemetry/schemas/com.example.telemetrydata`, in
-the which the reference is already in the form of a URI fragment, the suffix is
+`#/schemagroups/com.example.telemetry/schemas/com.example.telemetrydata`, in the
+which the reference is already in the form of a URI fragment, the suffix is
```
deissnerk:
```suggestion
`#/schemagroups/com.example.telemetry/schemas/com.example.telemetrydata`, in
which the reference is already in the form of a URI fragment, the suffix is
```

PR #410, opened 2025-09-19
split http

### LINKED ISSUE #409: modelsource=null -> err or erase everything?
`modelsource` include this:

```
During the processing of an update operation, if this attribute is present,
  then the Registry's model MUST be updated prior to any entities being 
  updated. A value of `null` MUST generate an error 
  ([invalid_data](#invalid_data)).
```
Should "null" just erase the model instead of generating an error?

Note, that `{}` will erase the model

duglin:
- Split the http stuff from spec.md. Now we have spec.md, http.md and model.md.
- includes a lot of rewording, removing of duplicate statements and a bit of clarifications.
- spec.md is now the core xReg stuff w/o any protocol specific APIs or model management discussions
- As of today model.md still has http stuff in it but that'll be merged into http.md next week.
- a few "TODOs" still in http.md for discussion for next week's PR
- #409 -  allow for setting modelsource=null to erase entire model

fixes #409

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -26,6 +26,7 @@ or automation and tooling usage.
   - [Importing Data](#design-importing-data)
   - [JSON `$schema` keyword](#design-json-schema-keyword)
   - [Implicit Creation of Parent Entities](#design-implicit-creation-of-parent-entities)
+  - [Events](#design-events)
 - [Registry Model](#registry-model)
   - [Implementation Customizations](#implementation-customizations)
   - [Attributes and Extensions](#attributes-and-extensions)
@@ -51,7 +52,6 @@ or automation and tooling usage.
   - [Sort Flag](#sort-flag)
   - [SpecVersion Flag](#specversion-flag)
 - [Error Processing](#error-processing)
-- [Events](#events)
 
 ## Overview
 
@@ -675,6 +675,12 @@ would need to be created directly. This also means that the creation of the
 original entity would fail and generate an error
 ([required_attribute_missing](./spec.md#required_attribute_missing)) for the
 appropriate parent entity.
+
+### Design: Events
+
+xRegistry defines a set of events that SHOULD be generated when changes are
+made to the entities within a Registry. See the [xRegistry Events](./events.md)
+specification for more information.
 
 ---
 
@@ -4121,14 +4127,6 @@ something unexpected happened in the server that caused an error condition.
 
 
 <!-- end-err-def -->
-
----
-
-## Events
-
-xRegistry defines a set of events that SHOULD be generated when changes are
-made to the entities within a Registry. See the [xRegistry Events](./events.md)
-specification for more information.
 
 [rfc3986-section-2-1]: https://tools.ietf.org/html/rfc3986#section-2.1
 [rfc3339]: https://tools.ietf.org/html/rfc3339


core/spec.md
```diff
@@ -86,18 +82,23 @@ The following 3 diagrams show (from left to right):<br>
...
-specifications as examples.
+This specification is meant to be a framework from which additional
+specifications can be defined that expose model-specific Resources and
+metadata. See the [Endpoint](../endpoint/spec.md) and
+[Schema](../schema/spec.md) extension specifications as examples.
+
+Additionally, this specification defines the core model and semantics of
+xRegistry, while secondary specifications (such as
+[xRegistry HTTP Binding](./http.md)) will discuss how to expose them in a
+protocol specific way.
```
lailabougria:
```suggestion
protocol-specific way.
```

core/spec.md
```diff
@@ -857,6 +897,9 @@ attributes. However, they MUST adhere to the following rules:
 
 #### Common Attributes
 
+TODO is it better to have this section, or to just define them fully
+upon first usage and then have all other point to that first usage?
```
duglin: Keep it

core/spec.md
```diff
@@ -3065,107 +2084,76 @@ The Resource MAY choose to simply store a URL reference to the externally
...
 
 - `<RESOURCE>base64`: this attribute MUST be used when the contents of the
-  Resource are stored within the Registry but `<RESOURCE>` cannot be used.
+  document are stored within the Registry but `<RESOURCE>` cannot be used.
   The Resource's document is base64 encoded and serialized as a string.
 
-- `<RESOURCE>url`: this attribute MUST be used when the Resource is stored
+- `<RESOURCE>url`: this attribute MUST be used when the document is stored
   external to the Registry. Its value MUST be a URL that can be used to
   retrieve its contents via an HTTP(s) `GET`. There is no requirement for the
```
duglin: make it generic

core/spec.md
```diff
@@ -1115,6 +1170,9 @@ of the existing entity. Then the existing entity would be deleted.
   This specification does not place any constraints on the data returned from
   an HTTP `GET` to this URL.
 
+TODO do we want to keep the HTTP-ness of this URL? And 'icon' attr too.
+deprecated.docs too
```
duglin: make it generic

core/spec.md
```diff
@@ -1957,18 +1601,24 @@ and the following Registry-level attributes:
...
   aspects of the model without the "noise" of the specification-defined parts.
 
-  If the implementation supports modifying the model, then this attribute,
-  or the `/modelsource` API, are the mechanisms by which it MAY be done.
+  During the processing of an update operation, if this attribute is present,
+  then the Registry's model MUST be updated prior to any entities being
+  updated. A value of `null` MUST generate an error
+  ([invalid_data](#invalid_data)).
+
+TODO - null->err or erase?
```
duglin: null -> erase

core/spec.md
```diff
@@ -2363,18 +1722,21 @@ The following defines the specification-defined capabilities:
 - It is STRONGLY RECOMMENDED that implementations support at least
   `/capabilities` and `/model`.
 
+TODO are these values HTTP specific? Perhaps not call this
+"apis", but instead be called "readable" and remove the leading "/" ?
+And then include "entities" in the list to align with "mutable" below?
```
duglin: leave as is

core/spec.md
```diff
@@ -2400,6 +1762,8 @@ The following defines the specification-defined capabilities:
   [pagination](../pagination/spec.md) specification (value of `true`).
 - When not specified, the default value MUST be `false`.
 
+TODO is this http specific? Maybe the flag isn't but the pagination spec is.
```
duglin: keep capability but move ref to spec to http spec

duglin: mention that some protocols might not support it at all

core/spec.md
```diff
@@ -2411,97 +1775,79 @@ The following defines the specification-defined capabilities:
 #### `specversions`
 - Name: `specversions`
 - Type: Array of strings
-- Description: List of xRegistry specification versions supported.
-- Supported values include:
+- Description: List of xRegistry specification versions supported by the
+  `specversion` flag.
+- Known values include:
```
duglin: "Non-exhaustive" list

core/spec.md
```diff
@@ -2411,97 +1775,79 @@ The following defines the specification-defined capabilities:
...
 change MUST NOT take effect until after the processing of the current
 request. Note that if the response includes the serialization of the
 Registry's capabilities, then the changes MUST appear in that serialization.
 
-For any capability that is an array of strings, a value of `"*"` MAY be used to
-indicate that the server MUST replace `"*"` with the full set of items that
-are available. An error ([capability_error](#capability_error)) MUST be
-generated if `"*"` appears with any other value in the list. `"*"` MUST NOT
-appear in the serialization in any server's response.
+TODO double check that "going into effect after" is ok
```
duglin: yes

duglin: add an example because "the semantic
change MUST NOT take effect" is too abstract

core/spec.md
```diff
@@ -2694,369 +2008,74 @@ and the following Group-level attributes:
...
+- A few extra attributes will appear to help with the navigation from the
+  Resource. For example, a URL to the `meta` sub-object and the `versions`
+  Collection attributes.
 
 The remainder of this section discusses the processing rules for Resources
 and Versions. While it mainly uses the term "Resource" for ease of reading, in
 most cases it can be assumed that the same processing applies for "Versions".
 When this is not the case, it will be explicitly called out.
 
+TODO perhaps we ought to use Versions instead of Resources for the generic term
```
duglin: Keep it

core/spec.md
```diff
@@ -3250,24 +2250,30 @@ and the following Resource-level attributes:
...
-  attribute is a server-controlled attribute and therefore cannot be modified
-  by clients. This specification makes no statement as to when Resources are to
-  be read-only.
+  attribute is a server-controlled attribute and therefore SHOULD NOT be
+  modifiable by clients. This specification makes no statement as to when
+  Resources are to be read-only.
+
+  It is expected that only certain clients (e.g. "admin" users) would be
+  allowed to edit this attribute. So, for "normal" users the "SHOULD"
+  statements are expected to be "MUST" statements.
```
duglin: s/statements/statement/
SHOULD NOT

core/spec.md
```diff
@@ -4994,340 +3075,62 @@ do so, MUST adhere to the following rules:
...
 
 This flag is intended for cases where the client wishes to avoid the server
 modifying the bytes of the document in any way - such as "pretty printing" it.
 
+This flag MAY be used on any request but will only influence the serialization
+of Resource and Version entities that have a domain-specific document.
+
+See: [`?binary` Flag](./http.md#binary-flag) for the HTTP usage.
+
+TODO do we want references like this?
```
duglin: do it for all

core/spec.md
```diff
@@ -5423,60 +3243,51 @@ the following:
...
+means that any Registry Collection's attributes MUST be modified to match the
+resulting subset. In particular:
 - If the collection is inlined, it MUST only include entities that match the
   filter expression(s).
 - The collection `url` attribute MUST include the appropriate filter
-  expression(s) in its query parameters such that an HTTP `GET` to that URL
-  would return the same subset of entities.
+  expression(s) such that a query to that URL would return the same subset of
+  entities.
+TODO ^^^ what if the protocol isn't http?
```
duglin: check the protocol to see if the filter flag appears in the URL or not

duglin: add link to http spec

core/spec.md
```diff
@@ -5765,19 +3616,62 @@ those cases, the client will need to query the individual inlineable attributes
 in isolation so the Registry can leverage
 [pagination](../pagination/spec.md) of the response.
 
+TODO ^^^ ref to pagination
```
duglin: tweak this

core/spec.md
```diff
@@ -2411,97 +1775,79 @@ The following defines the specification-defined capabilities:
...
-For any capability that is an array of strings, a value of `"*"` MAY be used to
-indicate that the server MUST replace `"*"` with the full set of items that
-are available. An error ([capability_error](#capability_error)) MUST be
-generated if `"*"` appears with any other value in the list. `"*"` MUST NOT
-appear in the serialization in any server's response.
+TODO double check that "going into effect after" is ok
 
-Regardless of the mechanism used to update the capabilities, the Registry's
-`epoch` value MUST be incremented.
+#### Offered Capabilities
```
duglin, line 1900: make this `/capabilitiesoffered`

duglin (commented):
reviewed with group

core/spec.md
```diff
@@ -60,7 +55,7 @@ A Registry consists of two main types of entities: Resources and Groups of
 such Resources.
 
 Resources typically represent the main data of interest in the Registry, while
-Groups, as the name implies, allow related Resources to be arranged together
+Groups, as the name implies, allows related Resources to be arranged together
```
lailabougria: I think it was right as it was, because if you drop what's in between the comma, it would be "allow"

core/spec.md
```diff
@@ -269,22 +265,24 @@ user-defined extensions attributes. In practice, Group often act as lightweight
...
 
-A Resource entity in the Registry holds one of more Versions of metadata or a
-domain-specific document. If a Resource holds multiple Versions, those can be
-organized with compatibility policies and lineage. Each Resource always has a
-default Version corresponding to one of the available Versions that is
-returned when requesting the Resource content. All held Versions can be
-accessed through the Versions collection.
+A Resource entity in the Registry holds one or more Versions of metadata, and
+optionally a domain-specific document. If a Resource holds multiple Versions,
+those can be organized with compatibility policies and lineage. Each Resource
```
lailabougria: Let's link to compatibility and ancestor here?

core/spec.md
```diff
@@ -269,22 +265,24 @@ user-defined extensions attributes. In practice, Group often act as lightweight
...
 
 #### Document Resources vs Metadata-Only Resources
 
 Each Version of a Resource MAY be defined to have a "domain-specific" document
 associated with it. These documents MAY be stored within the Version as an
 attribute, or MAY be stored external to the Version and a URL to its location
 will be stored within the Version instead. This model design choice is
-specified via the `hasdocument` aspect of the Resource's model definition.
+specified via the
+[`hasdocument` aspect](./model.md#groupsstringresourcesstringhasdocument)
```
lailabougria, line 298: You can ignore this to a later point, but do we differentiate between attribute and aspect, and why and when? I think hasdocument is consistently referred to as an aspect, but not sure why the differentiation is made.

core/spec.md
```diff
@@ -296,34 +294,103 @@ interest will be in the Schema documents associated with the Resources.
...
-interoperability.
-
-Note that simple file servers SHOULD support exposing Resources where the HTTP
-body response contains the Resource's associated "document" as well as the
-case where the HTTP response body contains a JSON serialization of the
-Resource via the `$details` suffix on the URL path. This can be achieved by
-creating a secondary sibling file on disk with `$details` at the end of its
-filename.
+This specification is designed such that clients can choose how they want the
+data from a server to be returned. There are three main "views" that clients
```
lailabougria, line 314:
```suggestion
This specification is designed such that clients can choose how they want to interact with the Registry's data. There are three main interaction models that clients
```

Just a suggestion

core/spec.md
```diff
@@ -296,34 +294,103 @@ interest will be in the Schema documents associated with the Resources.
...
+can choose from:
 
----
+- Single Document View
+
+  In this view, clients are retrieving all (or part) of the Registry hierarchy
```
lailabougria:
```suggestion
  In this view, clients retrieve all (or part) of the Registry hierarchy
```

core/spec.md
```diff
@@ -296,34 +294,103 @@ interest will be in the Schema documents associated with the Resources.
...
+  as a single document. In this case, nested (or child) entities MAY be
+  "inlined" into the retrieved document such that the need for secondary
+  interactions with the server is reduced.
+
+  This is often used for cases where the documents are to be stored in some
```
lailabougria:
```suggestion
  This is often used for cases where the documents are stored in some
```

core/spec.md
```diff
@@ -296,34 +294,103 @@ interest will be in the Schema documents associated with the Resources.
...
+  via independent "read" operations.
+
+  The [Registry HTTP `GET` APIs](./http.md#registry-http-apis), without the
+  use of the [`?doc` flag/query parameter](./http.md#doc-flag), is an example
+  of how to generate this view.
+
+- Multiple Document View
+
+  This is a variant of the "API view". In situations where the data within the
+  Registry is to be stored as independent files either on disk, or in some
```
lailabougria:
```suggestion
  This is a variant of the "API view". In situations where the Registry data is stored as independent files either on disk, or in some
```

core/spec.md
```diff
@@ -296,34 +294,103 @@ interest will be in the Schema documents associated with the Resources.
...
+  be managed independently for use in tooling or stored in some source-code
+  control system (e.g. Github) to minimize the number of conflicting edits
+  between users in a very fluid environment.
+
+  The [Registry HTTP `GET` APIs](./http.md#registry-http-apis), with the use
+  of the [`?doc` flag/query parameter](./http.md#doc-flag), is an example of
+  how to generate this view.
+
+This specification provides the mechanisms to allow for users to choose the
+best "view" for their needs. Regardless of the view, the design allows for
```
lailabougria, line 363: If the previous suggestion is accepted, this will need to be updated to interaction model as well.

core/spec.md
```diff
@@ -296,34 +294,103 @@ interest will be in the Schema documents associated with the Resources.
...
+this might manifest itself for HTTP servers.
+
+### Design: Protocol Bindings
+
+This specification defines the core model and semantics of an xRegistry
+server implementation without regard to what protocol might be used to
+interact with it.
+
+In general, all interactions with a server SHOULD be OPTIONAL and dictated by
+the specific needs of the environment which it is being used. However, it is
```
lailabougria:
```suggestion
the specific needs of the environment in which it is being used. However, it is
```

core/spec.md
```diff
@@ -992,7 +1037,7 @@ of the existing entity. Then the existing entity would be deleted.
   - MUST NOT appear in responses if the `shortself` capability is disabled.
   - MUST be unique across all entities in the Registry.
   - MUST be a non-empty absolute URL referencing the same entity as the `self`
-    URL, either directly or indirectly via an HTTP redirect.
+    URL, either directly or indirectly via a protocol specific redirect.
```
lailabougria:
```suggestion
    URL, either directly or indirectly via a protocol-specific redirect.
```

core/spec.md
```diff
@@ -1006,15 +1051,16 @@ of the existing entity. Then the existing entity would be deleted.
   Unlike `<SINGULAR>id`, which is unique within the scope of its parent, `xid`
   MUST be unique across the entire Registry, and as such is defined to be a
   relative URL from the root of the Registry. This value MUST be the same as
-  the `<PATH>` portion of its `self` URL, after the Registry's base URL, without
-  any `$` suffix (e.g. `$details`). Unlike some other relative URIs, `xid`
-  values MUST NOT be shortened based on the incoming request's URL; `xid`s
-  are always relative to the root path of the Registry.
+  the `<PATH>` portion of its `self` URL, after the Registry's base URL,
+  without any protocol specific xRegistry suffix (e.g. `$details` in the HTTP
```
lailabougria:
```suggestion
  without any protocol-specific xRegistry suffix (e.g. `$details` in the HTTP
```

core/spec.md
```diff
@@ -1006,15 +1051,16 @@ of the existing entity. Then the existing entity would be deleted.
...
+  case).  Unlike some other relative URIs, `xid` values MUST NOT be shortened
```
lailabougria:
```suggestion
  case). Unlike some other relative URIs, `xid` values MUST NOT be shortened
```

core/spec.md
```diff
@@ -2411,97 +1781,78 @@ The following defines the specification-defined capabilities:
...
 
-For any capability that is an array of strings, a value of `"*"` MAY be used to
-indicate that the server MUST replace `"*"` with the full set of items that
-are available. An error ([capability_error](#capability_error)) MUST be
-generated if `"*"` appears with any other value in the list. `"*"` MUST NOT
-appear in the serialization in any server's response.
+The above requirement is driven by the idea that modifying the capabilities
+of a server and modifying any entity data are typically two very distinct
+actions, and will not normally happen at the same time. However, if the
+situation does occur a consistent (interoperable) processing order needs to
```
lailabougria:
```suggestion
situation does occur, a consistent (interoperable) processing order needs to
```

core/spec.md
```diff
@@ -2694,550 +2016,492 @@ and the following Group-level attributes:
...
-  "createdat": "2024-04-30T12:00:00Z",
-  "modifiedat": "2024-04-30T12:00:01Z",
-
-  "messagesurl": "https://example.com/endpoints/ep1/messages",
-  "messagescount": 5
-}
-```
+- Document View Constraints:
+  - REQUIRED.
+  - If the `meta` entity is inlined in the document then this attribute
```
lailabougria:
```suggestion
  - If the `meta` entity is inlined in the document, then this attribute
```

core/spec.md
```diff
@@ -2694,550 +2016,492 @@ and the following Group-level attributes:
...
+  the passed-in `versionid` attribute value. This is done as a convenience
+  for clients to avoid them having to include a `versions` collection just
+  to set the initial default Version's `versionid`. In other words, when
+  the `versions` collection is absent on a create, but `versionid` is
+  present, there is an implied `"versions": { "<VID>": {} }` (where `<VID>`
+  is the `versionid` value).
 
-#### Resource Attributes
+When the xRegistry metadata is serialized as a JSON object, the processing
+of the 3 Version `<RESOURCE>` attributes MUST follow these rules:
```
lailabougria: Can we make this Version-level, and add a * after <RESOURCE to clarify exactly what is meant?

core/spec.md
```diff
@@ -2694,550 +2016,492 @@ and the following Group-level attributes:
...
+
+Both the source and target Resources MUST be of the same Resource model type,
+simply having similar Resource type definitions is not sufficient. This
+implies that the
+[`ximportresources`]./model.md#groupsstringximportresources) feature to
+reference a Resource type from another Group type definition MUST be
+used.
+
+An `xref` value that points to a non-existing Resource, either because
+it was deleted, never existed or the current client does not have permission
```
lailabougria, line 2485: Shouldn't the permission case be handled differently? I agree for the other situations, but accessing a resource that you're not authorized to see would give you a NotAuthorised in most other environments, which is helpful to understand why it can't be found, and now, there are multiple possible reasons it can't be found that may have to be validated somehow by the user. Why can't this be handled differently?

I know this was already here and you moved it around, but now I reread it and this thought came up.

core/spec.md
```diff
@@ -5694,39 +3544,34 @@ Some examples:
...
-```
-
-Where `<PATH>` is a string indicating which inlineable attributes to show in
-the response. References to nested attributes are represented using a
-dot (`.`) notation where the xRegistry collection names along the hierarchy
-are concatenated. For example: `endpoints.messages.versions` will inline all
-Versions of Messages. Non-leaf parts of the `<PATH>` MUST only reference
-xRegistry collection names and not any specific entity IDs since `<PATH>` is
-meant to be an abstract traversal of the model.
+The value of the `inline` flag is list of `<PATH>` values where is a string
```
lailabougria:
```suggestion
The value of the `inline` flag is a list of `<PATH>` values where is a string
```

lailabougria: Actually something is wrong at the end too, but not sure how to fix it 😅

core/spec.md

core/spec.md

core/spec.md
duglin: I've been trying to refer to "model attributes" as "aspects"... sort of like another word for "feature".
Then I use "attribute" for user-facing attributes in the data model (group, resources...).
But perhaps there's no point?

core/spec.md
duglin: Let's chat tomorrow because we do use the terms "document view" and "api view" - so I was trying to (sort of) relate to those too

core/spec.md

core/spec.md

core/spec.md

core/spec.md

core/spec.md

core/spec.md

core/spec.md
duglin: got them all

core/spec.md

core/spec.md
duglin: got them all

core/spec.md

core/spec.md
duglin: Let's discuss tomorrow, but they do have access to the Resource with the "xref" (the source), so they can always "GET" that one. If they don't have access to the target of the "xref" I'm not sure they should get a "401 Not Authorized" on a GET to the source.

It seems safest to treat a "missing target" the same as "you don't have permission to see the target". 
From a security perspective, even saying "the target exists but you can access it" is a security issue to some folks because it makes hackers aware that "something" is even there.

core/spec.md
duglin: See if this is better:
```
The value of the `inline` flag is a list of `<PATH>` values where each is a 
string indicating which inlineable attribute to show in the response. 
```

core/http.md
```diff
@@ -0,0 +1,3148 @@
...
+document that describes the message contents. These notations can be used or
+ignored by receivers of these messages. There is no requirement for
+implementations of this specification to persist these values, to include them
+in responses or to use this information.
+
+
+### No-Code Servers
+
+In the [core specification](./spec.md) there is a
+disussion about ["no-code servers"](./spec.md#design-no-code-servers). In the
```
clemensv: typo: discussion

core/spec.md
duglin: define "aspect"

core/spec.md
duglin: sticking with "view"

core/spec.md
duglin: stick with view

core/spec.md
duglin: leave as is

core/http.md
duglin: good catch! the spell checker wasn't turned on for this file!

core/spec.md
```diff
@@ -15,44 +15,43 @@ or automation and tooling usage.
...
-    - [Binary Flag](#binary-flag)
-    - [Collections Flag](#collections-flag)
-    - [Doc Flag](#doc-flag)
-    - [Filter Flag](#filter-flag)
-    - [Inline Flag](#inline-flag)
-  - [HTTP Header Values](#http-header-values)
-  - [Error Processing](#error-processing)
-  - [Events](#events)
+  - [Version Entity](#version-entity)
+- [Flags](#flags)
```
jskeet: This currently links to the flags bit of capabilities. Renaming the section to "Request flags" would resolve this issue *and* make it clearer. (Even before trying to click through, the title felt incongruous to me.)

core/spec.md
```diff
@@ -15,44 +15,43 @@ or automation and tooling usage.
...
+  - [IgnoreDefaultVersionID Flag](#ignoredefaultversionid-flag)
+  - [IgnoreDefaultVersionSticky Flag](#ignoredefaultversionsticky-flag)
+  - [IgnoreEpoch Flag](#ignoreepoch-flag)
+  - [IgnoreReadOnly Flag](#ignorereadonly-flag)
+  - [Inline Flag](#inline-flag)
+  - [SetDefaultVersionID Flag](#setdefaultversionid-flag)
+  - [Sort Flag](#sort-flag)
+  - [SpecVersion Flag](#specversion-flag)
+- [Error Processing](#error-processing)
+- [Events](#events)
```
jskeet: I'm not sure this deserves its own top-level section, given that it just links elsewhere. The text could potentially be inlined directly, or it could be a subsection - maybe of "Registry Design".

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+- [Registry HTTP APIs](#registry-http-apis)
+  - [Entity Processing Rules](#entity-processing-rules)
+    - [Creating or Updating Entities](#creating-or-updating-entities)
+  - [Registry Entity](#registry-entity)
+    - [`GET /`](#get-)
+    - [`PATCH /`](#patch-)
+    - [`PUT /`](#put-)
+    - [`POST /`](#post-)
+    - [`GET /export`](#get-export)
+  - [Registry Capability](#registry-capability)
```
jskeet: Possibly rename to Registry Capabilities, given that all requests operate on the collection?

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+    - [`DELETE /<GROUPS>/<GID>/<RESOURCES>/<RID>/versions/<VID>`](#delete-groupsgidresourcesridversionsvid)
+- [Flags / Query Parameters](#flags--query-parameters)
+  - [`?binary` Flag](#binary-flag)
+  - [`?collections` Flag](#collections-flag)
+  - [`?doc` Flag](#doc-flag)
+  - [`?epoch` Flag](#epoch-flag)
+  - [`?filter` Flag](#filter-flag)
+  - [`?ignoredefaultversionid` Flag](#ignoredefaultversionid-flag)
+  - [`?ignoredefaultversionsticky` Flag](#ignoredefaultversionsticky-flag)
+  - [`?ignoreeopch` Flag](#ignoreepoch-flag)
```
```suggestion
  - [`?ignoreepoch` Flag](#ignoreepoch-flag)
```

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+
+```yaml
+DELETE /endpoints/ep1/messages/msg1/versions?epoch=5
+```
+
+```
+HTTP/1.1 204 No Content
+```
+
+## Flags / Query Parameters
```
jskeet: Do we need all the detail here? I wonder whether just saying "Flags are represented by query parameters of the same name" would be enough (and keep a link to the flags in the spec, of course). Otherwise whenever a new flag is added or changed, we'll have to update two places.

If there are aspects to handling of some specific flags that are HTTP-specific, I think those could be called out here, only where necessary.

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+### `?binary` Flag
+
+A server MAY support the `?binary` flag/query parameter allowing clients to
+request that any Resources returned in the response always have its
+domain-specific documents be serialized under its `<RESOURCE>base64`
+ attribute.
+
+See [Binary Flag](./spec.md#binary-flag) for more information.
+
+This query parameter MUST be serialize as:
```
jskeet: If we keep this section, "serialize" => "serialized" in all cases.

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+is defined once in this section and any entity-, or collection-specific rules
+will be detailed in the appropriate section in the specification.
+
+#### Creating or Updating Entities
+
+This defines the general rules for how to update entities.
+
+Creating or updating entities MAY be done using HTTP `PUT`, `PATCH` or `POST`
+methods:
+- `PUT    <PATH-TO-ENTITY>                     # Process a single entity
```
jskeet: These are all missing closing backticks

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+
+  "endpointsurl": "https://example.com/endpoints",
+  "endpointscount": 42,
+
+  "schemagroupsurl": "https://example.com/schemagroups",
+  "schemagroupscount": 1
+}
+```
+
+#### `PATCH /`
```
jskeet, line 509: Is there a purpose to the empty patch sections? It's unclear whether that means "we haven't written the docs yet" or "it's not relevant to this collection".

jskeet: If the point is "the post section below applies" then I think that's confusing. (At first I thought it was just missing.)

A section of 

```md
#### `PATCH /` and `PUT /`
```

... would make more sense to me.

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+
+{ ... Contents of a schema doc excluded for brevity ...  }
+```
+
+#### `PATCH /<GROUPS>/<GID>/<RESOURCES>/<RID>/versions/<VID>`]
+#### `PUT /<GROUPS>/<GID>/<RESOURCES>/<RID>/versions/<VID>`]
+
+A server MAY support clients creating or updating a
+[Version](./spec.md#version-entity) of a
+[Resource](./spec.md#resource-entity) via an HTTP `PATCH` or `POST` directed
```
jskeet: Should POST be PUT here, as per the section title? POST is described later.

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+xRegistry-isdefault: true
+xRegistry-createdat: 2024-04-30T12:00:00Z
+xRegistry-modifiedat: 2024-04-30T12:00:01Z
+xRegistry-ancestor: 1.0
+Content-Disposition: myschema
+
+{ ... Contents of a schema doc excluded for brevity ...  }
+```
+
+#### `PATCH /<GROUPS>/<GID>/<RESOURCES>/<RID>/versions/<VID>`]
```
jskeet: Not sure what these `]` end-of-line characters are doing here

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+```
+
+references the (domain-specific) schema document associated with the
+`myschema` Resource.
+
+If a Resource does not support domain-specific documents (i.e.
+[`hasdocument`](./model.md#groupsstringresourcesstringhasdocument)
+is set to `false`), then use of the `$details` suffix in a request URL MUST
+be treated the same as if it were absent and any URLs in server response
+messages referencing that Resource (e.g. `self`) MUST NOT include it.
```
jskeet, line 1442: This feels odd to me. It means that if a resource changes from not having a domain-specific document to having one, a response can change from metadata to document... rather than changing from error to document.

This may all be ancient history and decided by now, but it feels to me like if I ask for something (the document) and that's not supported, that should give an error, rather than a different result.

To put it another way - suppose we'd made the metadata the default, and had `$document` as a suffix when requesting a document, would we want an explicit `$document` suffix request to return the metadata if there was no document?

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+
+##### Serializing Resource Domain-Specific Documents
+
+When a Resource is serialized as its underlying domain-specific document,
+in other words `$details` is not appended to its URL path, the HTTP body of
+requests and responses MUST be the exact bytes of that document. If the
+document is empty, or there is no document, then the HTTP body MUST be empty
```
jskeet, line 1449: I think we need to be clear about the difference between "there is no document" and "the resource type doesn't support documents". They can be easily confused, and we're saying they have very different meanings and results.

core/http.md
```diff
@@ -0,0 +1,3293 @@
...
+MUST be interpreted as a request to leave their values unchanged. Using a
+value of `null` (case-sensitive) indicates a request to delete that attribute.
+
+Any top-level map attributes that appear as HTTP headers MUST be included
+in their entirety and any missing keys MUST be interpreted as a request to
+delete those keys from the map.
+
+Since only some types of attributes can appear as HTTP headers, in order
+to manage the full set of attribute the xRegistry metadata view (via use
+of the `$details` URL suffix) MUST be used instead.
```
jskeet, line 1492: Unless the resource doesn't support a document, in which case a request without a `$details` suffix should apparently be treated the same way as if `$details` was specified? (This is an example of why I think that rule is unhelpful.)

core/spec.md

core/spec.md
duglin: I definitely see your point. I think the main reason I put it as a top-level section is because if someone was coming to the spec to see the list of events I didn't want the section to be hidden in the middle of the TOC. Let's see what other people think.

core/http.md
duglin: excellent catch!

core/http.md
duglin: got 'em all

core/http.md
duglin: Glad you focused on this - I was wonder about it.
I like your idea of just calling out the few flags that might need extra text. I'll hold off making the change until the other xReg folks chime in just in case there's a reason to keep the 1:1 mapping.

core/http.md
duglin: done

core/http.md
duglin: got 'em. And found another one elsewhere in the spec - thanks!

duglin: wow - found some in spec.md and model.md too - thanks!!
I added a checker tool for this!

core/http.md
duglin: Yup - that's the point.
For short URLs the "and" makes sense and still reads ok.

But for long URLs it awkward as a section header:
#### `PATCH /<GROUPS>/<GID>/<RESOURCES>/<RID>[$details]` and `PUT /<GROUPS>/<GID>/<RESOURCES>/<RID>[$details]`

Is this too terse/weird?
#### `PATCH ` and `PUT /<GROUPS>/<GID>/<RESOURCES>/<RID>[$details]`

core/http.md
duglin: yup! done

core/http.md
duglin: I think those used to have some flags listed at the end. Done

core/http.md
duglin: We may want to add some background on this to the primer since other people may wonder about it too, but....

switching between having a domain-specific doc and not is not something that will happen often. It's kind of a big decision point in how someone wants to store their data in the registry. Once that decision is made we had to consider each of those two types of users and what they would expect to be "default" for GET. For the non-domain-specific doc users it's easy since there's no choice... they'll get the metadata.

However, for a user with a domain-specific doc we decided that the information that's most important to them is that doc, not the xReg metadata. We did consider a `$document` type of flag (for consistency as you suggested) but then consider what a schema user would see/use:

```
GET https://example.com/schemagroups/group1/schemas/myschema.json$document
```
yuck! Kind of kills the usability factor.  So we ended up where we are.

This is related to why we return the scalar metadata as `xRegistry-` HTTP headers... just in case they wanted some of the metadata too, but really the doc is still the main reason for the GET. This allows xReg-aware clients and non-xReg-aware client to use the same URL.

I will point out though, we purposely support the idea of `$details` always being allowed on a Resource/Version URL so that clients shouldn't be required to check the `hasdocument` flag for each Resource prior to doing a "give me the metadata vie" request.

Does this make sense?

core/http.md
duglin: This paragraph isn't dealing with "the resource doesn't support docs" though. Do I need to make that clearer or were you just talking about the overall spec in general?

core/http.md
duglin: You have that correct. Let's discuss on the call given my explanation in a previous comment.

core/http.md
jskeet: I was saying that just a paragraph ago, we're talking about an attribute of `hasdocument` which *really* sounds like it's similar to "does this resource have a document" but isn't. Now we're talking about "does this resource have a document" which is different, and I think the two concepts can be confused really easily.

core/http.md
jskeet: I wasn't actually proposing using a `$document` suffix - I was just trying to introduce it as a thought experiment for how reasonable it is for `$details` to be effectively optional for any resource that doesn't support a document.

Fundamentally, it feels like this approach reduces predictability. If I see a URL which *doesn't* have `$details`, I don't know whether that will return the document or the metadata, unless I know whether the resource supports documents or not.

I suspect it's worth discussing in the meeting - or simply overruling the objection as "this has been thrashed out a lot in the past", which is totally valid.

duglin: > Fundamentally, it feels like this approach reduces predictability. If I see a URL which doesn't have $details, I don't know whether that will return the document or the metadata, unless I know whether the resource supports documents or not.

Yup that's true, and to some extent it's done on purpose. As i mentioned, we wanted the naked URL to always be "the data that's most important to THAT client". And the answer to that question changes based on whether the Resource type has a doc or not. So, here, "usability for client of THIS Resource type" takes precedence over "consistency between doc-Resources and non-doc-Resources".

Let's discuss on the call

core/http.md
duglin: ok - let me see if I can work on that...might not be until after the call though

duglin: If right before this paragraph (so at the start of this section), I added this:

```
This section goes into more details concerning situations where a Resource
type is defined to support domain-specific documents (i.e.
[`hasdocument`](./model.md#groupsstringresourcesstringhasdocument)
is set to `true`.
```

would that have helped?

duglin: You mentioned that `hasdocument` might not be as clear as it could be. See if I used this as the first paragraph of the previous section would help:

```
In the core specification's
[Resource Metadata vs Resource Document](./spec.md#resource-metadata-vs-resource-document)
section, it explains how Resources types might be defined to optionally have a 
domain-specific document associated with them via their
[`hasdocument` model aspect](./model.md#groupsstringresourcesstringhasdocument`)
being set to `true`. For HTTP, clients indicate whether they want to interact
with the Resource's xRegistry metadata or the Resource's domain-specific 
document by the use of a `$details` suffix on the `<RESOURCE>id` in its URL. 
Presence of the `$details` suffix MUST be interpreted as a request to interact 
with the xRegistry metadata of the Resource, while its absence MUST be
interpreted as a request to interact with the Resource's domain-specific
document.
```

jskeet: Quite possibly. I think naming the attribute "supportsdocuments" instead of "hasdocument" would probably be even clearer, but that ship's sailed :)

Possibly where we mention empty serialization of a resource where the type supports documents but there isn't one for the specific resource, it's worth adding a sentence to highlight that this isn't the same as the resource type not supporting documents.

duglin: hmm `supportsdocuments` is clearer. We'd need to discuss with the rest of the team but I don't think it's too late to rename it. "hasdoc" is so much easier to say though :-)

jskeet: > You mentioned that hasdocument might not be as clear as it could be. See if I used this as the first paragraph of the previous section would help:

I *believe* I understand that. Basically there are three possibilities:

- The model's `hasdocument` is true, and the resource itself has a document attached
- The model's `hasdocument` is true, and the resource itself doesn't have a document attached
- The model's `hasdocument` is false, so the resource itself *can't* have a document attached

What I'm saying is that those last two are pretty easily confused. "There is no document" sounds - at first glance - like `hasdocument=false`, right? But one is on the model, one is on the resource. I think it's okay as a design, but we need to be aware of an audience who isn't completely tuned in to that.

core/spec.md
```diff
@@ -3370,12 +2688,14 @@ and the following Resource-level attributes:
...
 
-##### `defaultversionurl` Attribute
+#### `defaultversionurl` Attribute
 - Type: URL
 - Description: a URL to the default Version of the Resource.
 
-  This URL MUST NOT include the `$detail` suffix even if the Resource's
-  `hasdocument` aspect is set to `true`.
+  Note that if the Resource's
+  [`hasdocument` aspect][./model.md#groupsstringresourcesstringhasdocument)
```
```suggestion
  [`hasdocument` aspect](./model.md#groupsstringresourcesstringhasdocument)
```

(The link is broken at the moment)

core/spec.md
duglin: AI: move into "design" section

core/spec.md
duglin: thanks! done

core/http.md
duglin: @jskeet any thoughts on this comment? I missed it during last week's Thursday call.

jskeet: Sorry, thought I'd replied to that - yes, I think the last option looks good.

PR #412, opened 2025-10-03
Week of Oct 6

duglin:
- typos
- Added support for OPTIONS HTTP method
- Make "Method not supported" error less http specific

### DIFF
--- a/core/http.md
+++ b/core/http.md
@@ -125,7 +125,7 @@ use of that API MUST generate an error
 ([api_not_found](./http.md#api_not_found)).
 
 If an HTTP method is not supported for a supported HTTP path, then an error
-([action_not_allowed](./spec.md#action_not_allowed)) MUST be generated.
+([action_not_supported](./spec.md#action_not_supported)) MUST be generated.
 
 Implementations MAY support extension APIs, however, the following rules apply:
 - New HTTP paths that extend non-root paths MUST NOT be defined.
@@ -2415,7 +2415,7 @@ Content-Type: application/json; charset=utf-8
 #### `DELETE /<GROUPS>/<GID>/<RESOURCES>/<RID>/meta`
 
 A `DELETE` directed to the Meta entity is not supported and MUST generate an
-error ([action_not_allowed](./spec.md#action_not_allowed)).
+error ([action_not_supported](./spec.md#action_not_supported)).
 
 ### Version Entity
 
--- a/core/spec.md
+++ b/core/spec.md
@@ -2497,7 +2497,7 @@ operation targeted directly to it.
 Each Resource MUST have a Meta entity, and when the Resource is deleted then
 the Meta entity MUST also be deleted. Delete requests directed to the Meta
 entity MUST generate an error
-([action_not_allowed](./spec.md#action_not_allowed)).
+([action_not_supported](./spec.md#action_not_supported)).
 
 A non-patch type of write operation that includes a Resource, as a mechanism
 to update its `meta` entity would need to include the Resource's default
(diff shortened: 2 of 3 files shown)


core/http.md
```diff
@@ -423,6 +424,46 @@ In the remainder of this specification, the presence of the `Link` HTTP header
 indicates the use of the [pagination specification](../pagination/spec.md)
 MAY be used for that API.
 
+#### HTTP OPTIONS Method
+
+A server MAY support clients querying the list of supported HTTP methods
```
jskeet, line 429: Is this usually supported out of the box by simple "just dump a load of files" hosting systems? (That being one of the design goals...)

core/spec.md
```diff
@@ -3965,12 +3965,16 @@ SHOULD attempt to use a more specific error when possible.
...
-* Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#method_not_allowed`
+In the case of using the HTTP protocol binding, this error MUST include
+the HTTP `Allow` header in the response indicating which HTTP methods
+the API supports, if any.
+
+* Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#action_not_allowed`
 * Code: `405 Method Not Allowed`
 * Instance: `<REQUEST URL>`
-* Title: `The specified HTTP method (<INVALID METHOD>) is not supported for: <REQUEST URL>`
+* Title: `The specified action (<INVALID ACTiON>) is not supported for: <REQUEST URL>`
```
```suggestion
* Title: `The specified action (<INVALID ACTION>) is not supported for: <REQUEST URL>`
```

core/spec.md
```diff
@@ -3965,12 +3965,16 @@ SHOULD attempt to use a more specific error when possible.
 * Title: `The data provided for "<ATTRIBUTE/PARAMETER NAME>" is invalid`
 * Detail: `<THE INVALID DATA>`
 
-### method_not_allowed
+### action_not_allowed
```
jskeet: Hmmm.... this name feels like it invites ambiguity. If I'm "not allowed" to do something, is it because I don't have *permission* to do it, or because what I'm asking is inherently invalid? It sounds like in this case, it's really about the latter - but I suspect it could easily be interpreted as the former.

For HTTP, code 405 is (I believe, anyway) specific about what's supported by the resource, as opposed to permissions. But if we're trying to describe it generically, perhaps we should call this "action_not_suppported"? After all, that's what the title description mentions...

core/http.md
duglin: All APIs are optional based on need and ability to support them in the hosting environment. I think OPTIONS is no different than POST - neither of which would probably be supported if the registry is hosted on S3.

core/spec.md
duglin: done - nice catch

core/spec.md
duglin: Couple of things...
- "not allowed" is taken directly from the http spec so I'm leaning on the fact folks seem to be ok with them using that phrase
- I think the HTTP is purposely vague as to whether it's "no permission" or "inherently invalid" (no support at all) because to the end user it's the same net result - they can't do it. And I like that because from a security perspective I don't think we want hackers to know the difference. Telling a hacker they need "more permission" is an invite to keep trying, or provides insight into something they're not allowed to access.

core/http.md
jskeet: I think I'd misread this as MUST or SHOULD for some reason...

core/spec.md
jskeet: Hmm... if an HTTP server returned 405 to me and I could fix it by authenticating, I'd be very surprised that it hadn't returned a 403 instead. Maybe that's just me - might be worth discussing.

core/http.md
```diff
@@ -124,7 +125,7 @@ use of that API MUST generate an error
 ([api_not_found](./http.md#api_not_found)).
 
 If an HTTP method is not supported for a supported HTTP path, then an error
-([method_not_allowed](./spec.md#method_not_allowed)) MUST be generated.
+([action](./spec.md#action_not_allowed)) MUST be generated.
```
deissnerk:
```suggestion
([action_not_allowed](./spec.md#action_not_allowed)) MUST be generated.
```

duglin: fixed!

core/spec.md
duglin: Perhaps, but I think that's an impl choice. And our spec even talks about how people can choose a more appropriate error message for their needs.

In general I do agree that 405 is probably more likely to mean the method isn't supported at all regardless of auth. However, as with most HTTP error code, there could be any number of reasons for why the error was generated and I think "not allowed" is vague/wide enough to cover lots of them.

Who would have thought that s/method/action/ would be the focus of the review! :-)  I just wanted to remove the HTTP specific aspect ("method") from the error. If my use of the word "supported" in the text causes you concern I can twiddle that. But I do like the synergy with HTTP's 405 status code's phrase.

duglin: Just did: s/supported/allowed/ in the text for consistency. :-)

duglin: Use: action_not_supported

PR #420, opened 2025-11-11
Week of Nov 3

duglin:
- tweaks to error messages
- change our error structure:
```
{
  "type": "https://github.com/xregistry/spec/blob/main/core/spec.md#bad_flag",
  "title": "The specified flag (collections) is not allowed in this context: /schemagroups.",
  "detail": "?collections is only allow on the Registry or Group instance level.",
  "subject": "/schemagroups",
  "args": {
    "flag": "collections"
  },
  "instance": "123e4567-e89b-12d3-a456-426614174000",
  "source": "parser:4123#d3beeeb5b0f"
}
```
- add new error messages
- add new HTTP header to all responses:
  `Link: <URL-TO-XREGISTRY-ROOT>;rel=xregistry-root`

Fixes: https://github.com/xregistry/spec/issues/421

### DIFF
--- a/core/http.md
+++ b/core/http.md
@@ -16,6 +16,8 @@ model and semantics that apply to all protocols.
 - [Registry HTTP APIs](#registry-http-apis)
   - [Entity Processing Rules](#entity-processing-rules)
     - [Creating or Updating Entities](#creating-or-updating-entities)
+    - [HTTP-Specific Attribute Processing Rules](#http-specific-attribute-processing-rules)
+    - [Pagination](#pagination)
     - [HTTP OPTIONS Method](#http-options-method)
   - [Registry Entity](#registry-entity)
     - [`GET /`](#get-)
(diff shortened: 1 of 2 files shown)


core/http.md
```diff
@@ -350,6 +352,28 @@ Note that the response MUST be generated applying the semantics of any
 [request flags](#request-flags--query-parameters) specified in the request
 URL (e.g. `?inline`).
 
+##### xRegistry Root HTTP Header
+
+All API responses SHOULD include a `Link` HTTP header in all responses of the
```
jskeet: Possibly clarify whether this is true for absolutely all responses including failures?

(If the response indicates a lack of authentication/authorisation, is this potentially leaking info? Pretty minor, admittedly...)

duglin: I tweaked it a bit to make it clear that errors should include it too.

I think the "SHOULD" and other text in the spec(s) where we say you can NOT do something for good reasons (like security) is sufficient. I don't think we need to repeat it here. Plus, I'm not sure what the risk is since that client already has the URL since they sent it the request to begin with.

jskeet: Yup, that sounds fine to me.

core/spec.md
```diff
@@ -2487,6 +2499,11 @@ detect. In these "dangling xref" situations, the serialization of the source
 Resource will not include any target Resource attributes or nested collections.
 Rather, it will only show the `<RESOURCE>id` and `xref` attributes.
 
+However, a non-existing Resource is not the same as a poorly formed XID
+value. An `xref` that isn't syntactically correct, or references a
+non-existing Group or Resource, MUST generate an error
+([malformed_xref](#malformed_xref))
```
jskeet: Nit: period at the end of the sentence?

core/spec.md
```diff
@@ -3902,157 +3962,304 @@ the API supports, if any.
...
+* Title: `Processing Resource "<subject>", the "defaultversionid" attribute is not allowed to be specified for Resources of type "<resource_type>".`
+* Args:
+  - `resource_type`: The Resource type name of the offending Resource.
+
+### defaultversionid_request
+
+* Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#defaultversionid_request`
+* Code: `400 Bad Request`
+* Subject: `<resource_xid>`
+* Title: `Processing Resource "<subject>", the "defaultversionid" attribute is not allowed to "request" is a Version wasn't processed.`
```
jskeet: The end of this sentence doesn't quite work at the moment.

core/spec.md

core/spec.md
duglin: fixed!
`Processing Resource "<subject>", the "defaultversionid" attribute is not allowed to be "request" since a Version wasn't processed.`

PR #425, opened 2025-12-02
Week of Dec 1

### LINKED ISSUE #426: POST .../rID isn't an alias for POST .../rID/versions - remove the lie

### LINKED ISSUE #427: Do we ever actually say to ingore read-only attribute on write ops?
I can't seem to find it. Closest I've found is in model.md where it talks about marking an extension as R/O means ignore updates. But we need to say that in general too.

### LINKED ISSUE #428: convert ?ignore* flags into just ?ignore=xxx,yyy,zzz  and support "modelsource", "capabilities"
Then people can use the same JSON file to preload a local xReg (with model+caps) and use the same json on a production server by specifying `?ignore=modelsource,capabilities`

`?ignore=defaultversionid,defaultversionsticky,epoch,readonly,modelsource,capabilities`

And a server can advertise which it supports in its "capabilities.ignore" field

Make it extensible so impls can add new values, w/o having to add an entirely new flag

duglin:
- More and better error messages
- make it clear that read-only attributes should be ignored on write ops
- fix the "lie" for POST .../rID - it's not an alias for "POST ../rID/versions".
  Also include text about ignoring the Resource-level attributes that might appear.
- define ?ignore=XXX* and remove the individual ?ignoreXXX flags

Fixes #427 
Fixes #426 
FIxes #428

### DIFF
--- a/core/http.md
+++ b/core/http.md
@@ -3227,7 +3227,7 @@ starting with `/`. E.g. `/export` if the "export" feature is not supported.
 
 * Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#details_required`
 * Code: `405 Method Not Allowed`
-* Title: `$details suffix is needed when using PATCH for an entity: <subject>.`
+* Title: `$details suffix is needed when using PATCH for the entity: <subject>.`
 * Subject: `<resource_xid>`
 
 #### extra_xregistry_header


core/http.md
```diff
@@ -248,8 +248,8 @@ semantics defined above with the following exceptions:
     operations, any missing REQUIRED attributes MUST generate an error
     ([required_attribute_missing](./spec.md#required_attribute_missing)).
 
-The `POST` variant when directed at a single entity, MUST adhere to the
-following:
+The `POST` variant when directed at a single entity (not a Resource), MUST
```
```suggestion
The `POST` variant when directed at a single entity other than a Resource, MUST
```

(Just to clarify that this isn't claiming that resources aren't entities.)

core/http.md
```diff
@@ -320,12 +332,15 @@ The processing of each individual entity follows the same set of rules:
...
 - When a write operation request includes the Resource/Version's metadata in
   the HTTP body, then the inclusion of any xRegistry `xRegistry-` HTTP headers
-  MUST generate an error ([extra_xregistry_headers](#extra_xregistry_headers)).
+  MUST generate an error ([extra_xregistry_header](#extra_xregistry_header)).
 - When a write operation request includes a domain-specific document in the
   HTTP body, and the `<RESOURCE>url` xRegistry HTTP header is present with a
   non-null value, the HTTP body MUST be empty. If the `<RESOURCE>url`
   attribute is absent, then the contents of the HTTP body (even if empty) are
   to be used as the entity's document.
+- The `<RESOURCE>` and `<RESOURCEbase64>` attribute MUST never appear as
```
```suggestion
- The `<RESOURCE>` and `<RESOURCEbase64>` attributes MUST never appear as
```

core/http.md
```diff
@@ -1524,6 +1547,10 @@ is not to allow for servers to decide whether or not to do so, rather it is to
 allow for [No-Code Servers](#no-code-servers) servers that might not be
 able to control the HTTP response headers.
 
+The `<RESOURCE>` and `<RESOURCEbase64>` attribute MUST NOT be serialized as
```
```suggestion
The `<RESOURCE>` and `<RESOURCEbase64>` attributes MUST NOT be serialized as
```

core/http.md
```diff
@@ -3027,6 +3059,16 @@ The abstract processing logic would be:
   sub-trees into one result set and remove any duplicates - adjusting any
   collection `url` and `count` values as needed.
 
+### `?ignore` Flag
```
jskeet, line 3062: This doesn't describe:

- What the flag is for or what it does
- What it means to have just `?ignore` rather than `?ignore=abc,xyz`

(Linking to the core spec would cover these, but at least a single sentence would help to give an idea of what it's about, like we have for `inline` and `filter`.)

core/http.md
```diff
@@ -3177,26 +3218,26 @@ starting with `/`. E.g. `/export` if the "export" feature is not supported.
 
 * Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#details_required`
 * Code: `405 Method Not Allowed`
-* Title: `$details suffix is needed when using PATCH for Resource: <subject>.`
+* Title: `$details suffix is needed when using PATCH for entity: <subject>.`
```
```suggestion
* Title: `$details suffix is needed when using PATCH for an entity: <subject>.`
```

core/spec.md
```diff
@@ -4271,6 +4351,30 @@ something unexpected happened in the server that caused an error condition.
...
+* Subject: `<resource_xid>`
+* Title: `The model attribute "setdefaultversionsticky" needs to be "false" since "maxversions" is "1".`
+
+### sort_noncollection
+
+* Type: `https://github.com/xregistry/spec/blob/main/core/spec.md#sort_noncollection`
+* Code: `400 Bad Request`
+* Subject: `<request_path>`
+* Title: `Can't sort on a non-collection result set. Query path: <subject>.`
+* Detail: Suggestion: list of attributes that are too large.
```
jskeet: I don't understand how this detail is relevant to this error. Not sure if this line is just in the wrong place?

core/http.md
duglin: done

core/http.md
duglin: done

core/http.md
duglin: done

core/http.md
duglin: Doi - sorry - fixed

core/http.md
duglin: done

core/spec.md
duglin: LOL - copy-n-paste error - good eye

PR #436, opened 2026-02-17
Week of Feb 16

duglin:
- A bit of clarity around xref <-> normal resource processing
- modify resource processing logic to handle the case where "versions"
  is present w/o any hint as to whether a separate default Version should
  be created. Adjust resource.md with more samples

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -2374,26 +2374,37 @@ The following provides additional details:
       might allow for skipping unnecessary work.
 
 - Step 2 - Processing the Resource's default Version attributes:
-  - This step MUST use the Version referenced by `meta.defaultversionid` as
-    defined prior to the operation being processed.
-  - If the Resource is created as part of this operation then determining the
-    `versionid` to be used as "default" MUST adhere to the following:
-    - If a Resource-level `versionid` attribute is present in the request,
-      then its value MUST be used.
-    - Otherwise, if a `meta.defaultversionid` attribute is present in the
-      request, then its value MUST be used. Also see
+  - In this step the Resource's default Version (as defined prior to the
+    start of the operation) is to be updated with the Resource-level attributes
+    specified in the request. Note that an empty list of attributes, when not
+    doing a patch type of operation, will reset all attributes to their default
+    values.
+  - However, when the Resource is created as part of the current operation
+    there is no "prior" default Version, so knowing how to apply those
+    attributes requires some additional clarity:
+    - If a Resource-level `versionid` attribute is present, and that Version
+      is not in the request's `versions` collection, then a new Version with
+      that `versionid` MUST be created with the Resource-level attributes. If
+      it is in the `versions` collection then the Resource-level attributes
+      MUST be ignored.
+    - If `versionid` is not present, but `meta.defaultversionid` is present,
+      and that Version is not in the request's `versions` collection, then a
+      new Version with that `versionid` MUST be created, with the
+      Resource-level attributes. If it is in the `versions` collection then
+      the Resource-level attributes MUST be ignored. Also see
       [SetDefaultVersionID Flag](#setdefaultversionid-flag) for an additional
       mechanism by which it might be set.
-    - Otherwise, if the `versions` collection contains at least one Version,
-      then all of the Resource's default Version attributes MUST be ignored
-      and the remainder of the steps below MUST be skipped. This is because
-      without any indication from the client as to which Version those
-      attributes need to be applied to, rather than creating one, the assumption
-      is that one of the Versions in the `versions` collection is that Version.
-    - If neither the `versionid` nor `defaultversionid` are present, then the
-      server MUST generate a new unique [`versionid` value](#version-ids).
-    - If a Version with the target default `versionid` does not exist then a
-      new Version with that `versionid` MUST be created.
+    - If neither of those attributes are present, then the `versions`
+      collection will dictate the result:
+      - If `versions` is missing, or empty, then the server MUST generate a
+        new unique [`versionid` value](#version-ids) and create a new Version
+        using that value, with the Resource-level attributes.
+      - If `versions` is present with at least one Version, then the
+        Resource-level attributes MUST be ignored. This is because without
+        any indication from the client as to which Version the Resource-level
+        attributes need to be applied to, rather than creating a new Version,
+        the assumption is that `versions` contains all of the Versions that are
+        meant to be created.
   - If the operation is converting a cross-reference Resource back into a
     "normal" Resource then the new Version created MUST adhere the rules
     specified in the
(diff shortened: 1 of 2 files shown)


core/spec.md
```diff
@@ -2378,16 +2378,26 @@ The following provides additional details:
...
       mechanism by which it might be set.
-    - If neither are present, then the server MUST generate, and use, a new
-      [`versionid` value](#version-ids).
+    - Otherwise, if the `versions` collection contains at least one Version,
+      then all of the Resource's default Version attributes MUST be ignored
+      and the remainder of the steps below MUST be skipped. This is because
+      without any indication from the client as to which Version those
+      attributes need to be applied to, rather than creating one, the assumption
+      is that one of the Versions in the `versions` collection is that Version.
+    - If neither the `versionid` nor `defaultversionid` are present, then the
```
jskeet: I think this bullet needs to be updated to include the `versions` collection. (If there's *just* a versions collection, it doesn't need to generate a new versionid.)

core/spec.md
```diff
@@ -2378,16 +2378,26 @@ The following provides additional details:
...
+    - If a Resource-level `versionid` attribute is present in the request,
+      then its value MUST be used.
+    - Otherwise, if a `meta.defaultversionid` attribute is present in the
+      request, then its value MUST be used. Also see
       [SetDefaultVersionID Flag](#setdefaultversionid-flag) for an additional
       mechanism by which it might be set.
-    - If neither are present, then the server MUST generate, and use, a new
-      [`versionid` value](#version-ids).
+    - Otherwise, if the `versions` collection contains at least one Version,
+      then all of the Resource's default Version attributes MUST be ignored
```
jskeet: I don't think this is right - or resource.md is wrong. See the "Create Resource with versionid and Versions" example, where v0 is only mentioned in the resource's default version attributes, but it *is* in the result.

I think we *do* want to ignore all the resource's default version attributes if there's a version in the collection *and* meta.defaultversionid refers to a version within the collection.

core/spec.md
```diff
@@ -2378,16 +2378,26 @@ The following provides additional details:
     defined prior to the operation being processed.
   - If the Resource is created as part of this operation then determining the
     `versionid` to be used as "default" MUST adhere to the following:
-    - If a Resource `versionid` attribute is present in the request, then it
-      MUST be used.
-    - Otherwise, if `meta.defaultversionid` attribute is present in the
-      request, then it MUST be used. Also see
+    - If a Resource-level `versionid` attribute is present in the request,
+      then its value MUST be used.
+    - Otherwise, if a `meta.defaultversionid` attribute is present in the
```
jskeet: I don't think I'd make this an "otherwise" - i.e. we shouldn't just entirely ignore `meta.defaultversionid` because there's a `versionid`. I *think* yesterday I suggested we should check them for consistency - but I may be wrong. (The fact that I'm finding it hard to remember what I said yesterday is somewhat worrying in itself, in terms of the complexity of these rules...)

lailabougria: If we check for consistency, we should throw an error if they're inconsistent, right?
I'm generally in favor of that, but I believe there was a preference towards throwing less and ignoring values in the order of relevance, as the text currently suggests.

duglin: The "otherwise" is actually really important because it's an if-then-else thing. If Resource.versionid is present then meta.defaultversionid IS ignored for the purpose of trying to figure out which versionid should be used as the "default version as seen before the op started". Yes meta.defaultversionid might be used for other purposes later, but not for this specific task/step of steps.

core/spec.md
duglin: Again, the "otherwise" is critical. If Resource.versionid is present then this bullet will never be processed.

core/spec.md
duglin: This bullet will never be processed if there's a "versions" collection w/o versionid & defaultversionid though - that's why I purposely didn't include it. See the previous bullet about stopping processing.

Maybe I need to clean this up more...

core/spec.md
jskeet: Okay, just for processing the default resource attributes, I agree.

core/spec.md
jskeet: I think it's more that I hadn't clocked the scope of this change is only within "determine the default version".

core/spec.md
jskeet: It makes it hard to read when sometimes the "stopping" is performed by "otherwise", and sometimes by an instruction within the previous bullet.

If you just change this to "Otherwise, the server must" then I think it's clearer that this is picking the default value from multiple steps and each one stops the following ones.

core/spec.md
```diff
@@ -2378,16 +2378,26 @@ The following provides additional details:
...
-      [`versionid` value](#version-ids).
+    - Otherwise, if the `versions` collection contains at least one Version,
+      then all of the Resource's default Version attributes MUST be ignored
+      and the remainder of the steps below MUST be skipped. This is because
+      without any indication from the client as to which Version those
+      attributes need to be applied to, rather than creating one, the assumption
+      is that one of the Versions in the `versions` collection is that Version.
+    - If neither the `versionid` nor `defaultversionid` are present, then the
+      server MUST generate a new unique [`versionid` value](#version-ids).
     - If a Version with the target default `versionid` does not exist then a
```
jskeet: I think this should outdented, because it's not about *determining* the version to be used as the default version ID.

core/spec.md
```diff
@@ -2374,20 +2374,41 @@ The following provides additional details:
...
-    - If neither are present, then the server MUST generate, and use, a new
-      [`versionid` value](#version-ids).
-    - If a Version with the target default `versionid` does not exist then a
-      new Version with that `versionid` MUST be created.
+    - If neither of those attributes are present, then the `versions`
+      collection will dictate the result:
+      - If `versions` is missing, or empty, then the server MUST generate a
+        new unique [`versionid` value](#version-ids) and create a new Version
+        using that value, with the Resource-level attributes.
+      - If `versions` is present with at least one Version, then the
```
lailabougria, line 2402: Would it be worth repeating here that the default version will be the last one based on alphabetical order?

lailabougria (approved):

core/spec.md
duglin: Perhaps, *IF* this bullet mentioned "default version", but notice that it doesn't. I didn't think mentioning "default version" here helped (in fact I think it might confuse things) because regardless of which Version in "versions" ends up becoming the default one doesn't change the fact that we're ignoring Resource.* attributes. And calculating which Version in "version" ends up being the final default Version isn't really part of this discussion.  Make sense?

lailabougria: Fair enough, yes.

duglin:
Approved on the 2/19 call

PR #438, opened 2026-03-06
Week of Mar2

duglin:
- minor tweaks and typos
- add 'formatauth' and 'format' to core spec/model, remove 'format' from schema
- compatiblewith -> modelcompatiblewith
- s/formatauth/validateformat/ and moved it into the Resource model
- s/compatibilityauth/validatecompatibility/ and moved it into the Reource model


jskeet (approved):
I've only done a skim of this, and it's before the first coffee of the day has had time to kick in, but it seems reasonable.

duglin:
Approved on the 3/12 call

PR #440, opened 2026-03-12
Adding `dataschemaxid` to the `model.json`

deissnerk:
The `dataschemaxid` as defined [here](https://xregistry.io/xreg/xregistryspecs/message-v1/docs/spec.html#dataschemaxid), was missing in `model.json`.

### DIFF
--- a/message/model.json
+++ b/message/model.json
@@ -1360,7 +1360,7 @@
             "dataschemaxid": {
               "name": "dataschemaxid",
               "description": "Contains the xid of the xRegistry schema Resource entity associated with the dataschemauri referenced schema document. Note that this means the entity MUST be located within the same Registry.",
-              "type": "string"
+              "type": "xid"
             },
             "datacontenttype": {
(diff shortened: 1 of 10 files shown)


message/model.json
```diff
@@ -1357,6 +1357,11 @@
               "description": "The URI of the schema for the message payload, equivalent to the 'schemauri' attribute of the schema registry",
               "type": "uri"
             },
+            "dataschemaxid": {
+              "name": "dataschemaxid",
+              "description": "Contains the xid of the xRegistry schema Resource entity associated with the dataschemauri referenced schema document. Note that this means the entity MUST be located within the same Registry.",
+              "type": "string"
```
duglin: I think this should be "xid" not "string", right?

duglin:
@deissnerk you'll need to regen the schema files.  `make -f tools/Makefile` should do it

duglin:
more like a typo - so merging

PR #454, opened 2026-04-08
Week of Apr 6

### LINKED ISSUE #453: Questons about compatibility and format
### `epoch` & `modifiedat`:
- When `Version.compatibilityvalidated` changes, should the Version's `epoch`/`modifiedat` change? Today, spec says **'no'**
- When `Version.formatvalidated` changes, should the Version's `epoch`/`modifiedat` change? Today, spec says **'no'**
- Should there be a mechanism by which a client can know when a Version is no longer compatible? If `epoch`/`modifiedat` do not change then it's likely no notification would be sent either. Is this a problem? Similar question for `formatvalidated` and even `isdefault`.  **We'll deal with this when/if we define notifications.**

### Compatibility:
Format validation is a Version localized check.

Compatibility validation is always influenced by at least 2 Versions, could be all Versions.
- Determining, after the update, which Version is the problem child could be challenging.
- Is this something we want to address? **No**

If v1 changes such that v2 is no longer "backward" compatibility-validated:
  - should v2 be marked as `compatibilityvalidated=false` ? Today the spec says 'yes'
  - should v1 be marked as `compatibilityvalidated=false` ? Today the spec says 'no'
  - Is this ok? **yes**

Scenario:
- `meta.compatility=fulltransitive`
- `versionsCount=5`
- `strictvalidation=false`
- `v*.format=avro`
- `v*.formatvalidated=true`
- `v*compatibilityvalidated=true`

Then:
- set `v3.format="unknown"`   ->  `v*.compatibilityvalidated=false`
- Yes / no ? Is this ok?  **yes**

duglin:
- add 'consistentformat' to Resource model
- make the default value for the validation flags 'false'
- make schema registry be true/strict wrt 3 format/compat flags (not strict)

Fixes #453

### DIFF
--- a/core/model.md
+++ b/core/model.md
@@ -2,6 +2,7 @@
 
 <!-- words: compat validatecompatibility validateformat strictvalidation -->
 <!-- words: matchcase compatibilityvalidated formatvalidated -->
+<!-- words: consistentformat validators -->
 
 ## Abstract
 
@@ -134,9 +135,10 @@ The overall format of a model definition is as follows:
           "hasdocument": <BOOLEAN>, ?       # Has separate document. Default=true
           "versionmode": "<STRING>", ?      # 'ancestor' processing algorithm
           "singleversionroot": <BOOLEAN>, ? # Enforce single root. Default=false
-          "validatecompatibility": <BOOLEAN>, ? # Do Version compat checks. Default=true
-          "validateformat": <BOOLEAN>, ?    # Do Version format checks. Default=true
+          "validateformat": <BOOLEAN>, ?    # Do Version format checks. Default=false
+          "validatecompatibility": <BOOLEAN>, ? # Do Version compat checks. Default=false
           "strictvalidation": <BOOLEAN>, ?  # Block unknown format/compat. Default=false
+          "consistentformat": <BOOLEAN>, ?  # Same format for all Vers. Default=false
           "typemap": <MAP>, ?               # ContentType mappings
           "attributes": { ... }, ?          # Version attributes/extensions
           "resourceattributes": { ... }, ?  # Resource attributes/extensions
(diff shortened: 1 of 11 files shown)


core/model.md
```diff
@@ -777,12 +779,29 @@ The following describes the attributes of the Registry model:
   Enforcement](./primer.md#singleversionroot-policy-enforcement) section of
   the Primer for more information.
 
+### `groups.<STRING>.resources.<STRING>.validateformat`
```
duglin, line 782: Just moved up - no changes

core/model.md
```diff
@@ -777,18 +779,49 @@ The following describes the attributes of the Registry model:
...
   Resource type adhere to its owning Resource's `meta.compatibility` value.
-- When not specified, the default value MUST be `true`.
+- When not specified, the default value MUST be `false`.
 - A value of `true` indicates that the server MUST generate an error
   [compatibility_violation](spec.md#compatibility_violation)) if any Version
   of a Resource instance of this Resource type does not adhere to the rules of
   the `meta.compatibility` value for that Resource's `format` value.
   See [`strictvalidation`](#groupsstringresourcesstringstrictvalidation) for
   exceptions to this requirement.
+- When a Version does not have a `format` value then compatibility verification
```
jskeet, line 823: So if `validatecompat` is true, but `format` is missing, should that raise an error?

core/spec.md
```diff
@@ -3218,24 +3220,50 @@ the [`compatibility`](#compatibility-attribute) conformance checks, if
...
+  - `"true"`
+  - `"false, <REASON>"`
 
-  A value of `false` indicates that the server has not validated the Version
-  due to the `format` attribute value not being supported. Note that this
-  attribute MUST only ever be `false` when the Resource's
+  A value of `"true"` indicates that the server has validated the Version and
+  it adheres to the `format` attribute's rules.
+
+  A value that starts with `false` indicates that the server did not attempt
```
jskeet, line 3237: So is the format just `false because I felt like it` or is there more structure, e.g. `false (reason here)`?

core/spec.md
```diff
@@ -3244,33 +3272,62 @@ the [`compatibility`](#compatibility-attribute) conformance checks, if
...
+  - If present, MUST be non-empty.
   - MUST be a read-only attribute.
   - MUST be present if the Resource model's `validateformat` attribute is
     `true` and the Version's `format` attribute is present.
   - MUST NOT be present if the Resource model's `validateformat` attribute is
     `false` or the Version's `format` attribute is absent.
 
+- Examples:
+  - `"true"`
+  - `"false, unknown format"`
```
jskeet, line 3284: This answers my earlier question - I'm not particularly comfortable with it though. I'd personally prefer a more structured approach, keeping `formatvalidated` as a `Boolean`, and having a separate `formatvalidationdetail` attribute. The idea of any user code needing to check whether a value starts with `false` feels icky to me.

core/spec.md
```diff
@@ -3244,33 +3272,62 @@ the [`compatibility`](#compatibility-attribute) conformance checks, if
...
+
 #### `compatibilityvalidated` Attribute
-- Type: Boolean
+- Type: String
```
jskeet, line 3287: Same comment as above, really.

core/model.md
duglin: If "format" is missing, it's basically saying "turn off format/compat validation for this version". So, no it doesn't raise an error.

I struggled with the location for text (model vs spec), but in the core spec w.r.t. the "compatvalidated" attribute it says this:
```
  - MUST NOT be present if the Resource model's `validatecompatibility`
    attribute is `false`, or the Version's `format` value is absent, or the
    Resource's `meta.compatibility` attribute is absent.
```

core/spec.md
duglin: I wasn't sure how picky I should be about the shape of the text. I tried to stick with just "true" vs "false.*" (just first 5 chars matter). Is that ok? I was torn between human readable sentence and from machine parse-able (e.g. your `false (reason here)`), and decided to leave that as an impl choice.

core/spec.md
duglin: I share the concern, and we did talk about this a bit. For me, I think the overriding factor was that I didn't want to introduce "yet another" attribute when it pretty much goes "hand in hand" (ie required) with the "false". The other thought I had was that when it's a boolean it'll be too easy for people to only look at the boolean, ignore the text, and mentally think of "false" as the same thing as "invalid". Which it's not. "false" means "I didn't check".  And seeing the other text right next to the "false" increases the chances of people reading it and realizing the server didn't even try.

Perhaps true/false is part of the problem. Maybe we should not try to squeeze a boolean into a string - but if we don't, I think it makes it even parse for a machine to know what's going on. At least now they can parse the first few chars to get a quick answer.

jskeet: If this were a project I were more deeply invested in, I'd definitely push back on this. I don't think there's significant cost to an extra attribute, and it would be easier to use. But given that I don't have skin in the game, I'll let y'all decide :)

duglin: oh the guilt trip!

duglin:
Summary:
```
/capabilities
{
  "formats": [ "<STRING>", * ], ?          # "avro", "protobuf", "jsonSchema" ],
  "compatibilities": {
    "<STRING>", [ "<STRING>", * ], *       # "avro": [ "backward", "forward", "full" ],
  }                                        # "protobuf": [ "backward" ]
}

/model  (on RESOURCE definition)
{
  ...
  "validateformat": <BOOLEAN>,             # Default is false
  "validatecompatibility": <BOOLEAN>,      # Default is false. If true then validateformat MUST be true
  "strictvalidation": <BOOLEAN>,           # Default is false. Allow unknown format/compat values, RESOURCEurl
  "consistentformat": <BOOLEAN>,           # Default is false. All versions MUST have the same format value
  ...
}


/...RESOURCE Instance
{
  ...
  "format": "<STRING>", ?                   # Absent=no validation. Unknown values are ignored
  "formatvalidated": "<STRING>", ?          # Absent=disabled. "true" or "false, because...". Read-only
  "compatibilityvalidated": "<STRING>", ?   # Absent=disabled. "true" or "false, because...". Read-only
  ...
  "meta": {
    ...
    "compatibility": "<STRING>", ?          # Absent=no validation. Unknown values are ignored
    ...
  }
}
```

duglin:
Approved on the 4/16 call

PR #455, opened 2026-04-20
Week of Apr 20

duglin:
- typos, wording tweaks and tooling updates
- split `formatvalidated` and `compatvalidated` - e.g:

```
"formatvalidated": "false, reason we didn't check"
```
becomes:
```
"formatvalidated": false,
"formatvalidatedreason": "reason we didn't check"
```

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -947,6 +947,11 @@ following rules:
 - Unless otherwise stated, requests to update read-only attributes MUST be
   silently ignored by servers, even if the contents of those attributes are
   invalid.
+- <a id="read-only-changed"></a>Unless otherwise stated, server managed changes
+  to read-only attributes MUST NOT be treated the same as user initiated
+  changes to non-read-only attributes with respect to updating attributes such
+  as `modifiedat` and `epoch`. In other words, as these read-only attributes
+  are changed, the entity MUST NOT be considered "updated".
 
 #### Extensions
 
@@ -1204,7 +1209,7 @@ of the existing entity. Then the existing entity would be deleted.
 
 - Constraints:
   - REQUIRED.
-  - MUST be a read-only attribute.
+  - MUST be a [read-only](#read-only-changed) attribute.
   - MUST be an unsigned integer equal to or greater than zero.
   - MUST increase in value each time the entity is updated.
 
@@ -1691,7 +1696,7 @@ and the following Registry-level attributes:
     [Inline Flag](#inline-flag).
   - MUST be included in API and document views if requested via the
     [Inline Flag](#inline-flag).
-  - MUST be a read-only attribute.
+  - MUST be a [read-only](#read-only-changed) attribute.
 
 #### `modelsource` Attribute
 - Type: [Registry Model](./model.md#registry-model).
@@ -2855,7 +2860,7 @@ and the following Meta-level attributes:
 
 - Constraints:
   - REQUIRED.
-  - SHOULD be a read-only attribute.
+  - SHOULD be a [read-only](#read-only-changed) attribute.
   - When not specified, the default value MUST be `false`.
   - It MUST be a case-sensitive `true` or `false`.
   - A request to update a read-only Resource SHOULD generate an error
@@ -3123,11 +3128,7 @@ and the following Version-level attributes:
 #### `isdefault` Attribute
 - Type: Boolean
 - Description: Indicates whether this Version is the "default" Version of the
-  owning Resource. This value is different from other attributes in that it
-  might often be a calculated value rather than persisted in a datastore.
-  Thus, when its value changes due to the default Version of a Resource
-  changing, the Version itself does not change - meaning attributes such as
-  `modifiedat` remains unchanged.
+  owning Resource.
 
   See [Creating or Updating Resources and
   Versions](#creating-or-updating-resources-and-versions) for additional
@@ -3135,7 +3136,7 @@ and the following Version-level attributes:
 
 - Constraints:
   - REQUIRED.
-  - MUST be a read-only attribute.
+  - MUST be a [read-only](#read-only-changed) attribute.
   - When not specified, the default value MUST be `false`.
   - When specified, MUST be either `true` or `false`, case-sensitive.
 
@@ -3239,7 +3240,7 @@ for additional information.
 - Description: When
   [`format` validation](./model.md#groupsstringresourcesstringvalidateformat)
   is enabled, this attribute will indicate whether or not the server has
-  attempted to validated that the Version conforms to the rules defined by
+  attempted to validate that the Version conforms to the rules defined by
   its `format` attribute's value.
 
   A value of `true` indicates that the server has validated the Version and
@@ -3247,7 +3248,7 @@ for additional information.
   [`formatvalidatedreason`](#formatvalidatedreason-attribute) MUST NOT be
   present.
 
-  A value of `false` indicates that the server did not attempt to validated
+  A value of `false` indicates that the server did not attempt to validate
   the Version. This can happen when
   [`strictvalidation`](model.md#groupsstringresourcesstringstrictvalidation)`
   is set to `false` and:
@@ -3265,7 +3266,7 @@ for additional information.
  reject the update request instead.
 
   When `false`, the
-  [`formatvalidatedreason`](#formatvalidatedreason-attribute) MUST attribute
+  [`formatvalidatedreason`](#formatvalidatedreason-attribute)
   MUST be present with an explanation as to why the server was unable to
   attempt validation of the Version. Additionally, when `false`, if
   `compatibility` validation is enabled, the Version's
@@ -3279,15 +3280,10 @@ for additional information.
   [`strictvalidation`](model.md#groupsstringresourcesstringstrictvalidation)`
   model aspect.
 
-  Due to this attribute being server managed, and influenced by the state of
-  other Versions, as its value changes the Version itself MUST NOT be marked as
-  changed - meaning, attributes such as `modifiedat` and `epoch` remain
-  unchanged.
-
 - Constraints:
   - OPTIONAL
   - When specified, MUST be either `true` or `false`, case-sensitive.
-  - MUST be a read-only attribute.
+  - MUST be a [read-only](#read-only-changed) attribute.
   - MUST be present if the Resource model's `validateformat` attribute is
     `true` and the Version's `format` attribute is present.
   - MUST NOT be present if the Resource model's `validateformat` attribute is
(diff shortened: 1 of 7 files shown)


core/spec.md
```diff
@@ -3251,12 +3264,13 @@ for additional information.
  ([format_unknown](#format_unknown) or [format_external](#format_external)) and
  reject the update request instead.
 
-  When `"false"`, this value MUST include some additional text (the `<REASON>`)
-  indicating why the server was unable to attempt validation of the Version.
-  Additionally, when `"false"`, if `compatibility` validation is enabled,
-  the Version's
+  When `false`, the
+  [`formatvalidatedreason`](#formatvalidatedreason-attribute) MUST attribute
```
```suggestion
  [`formatvalidatedreason`](#formatvalidatedreason-attribute) attribute
```

core/spec.md
```diff
@@ -3251,12 +3264,13 @@ for additional information.
...
+  MUST be present with an explanation as to why the server was unable to
+  attempt validation of the Version. Additionally, when `false`, if
```
```suggestion
  perform validation of the Version. Additionally, when `false`, if
```

(or just "validate"). I suspect in every case it is *attempting* validation, but the reason is why it can't actually perform the validation.

jskeet: (As it happens, looking down to line 3300, it seems I coincidentally stumbled on the same terminology...)

core/spec.md
```diff
@@ -3272,37 +3286,49 @@ for additional information.
...
-- Examples:
-  - `"true"`
-  - `"false, unknown format"`
+#### `formatvalidatedreason` Attribute
+- Type: String
+- Description: When [`formatvalidated`](#formatvalidated-attribute) has a
+  value of `false`, this attribute MUST provide information as to why the
+  format validation check was not performed.
+
+  Due to this attribute being server managed, and influenced by the state of
```
jskeet: This feels like it lacks a bit of context - remain unchanged under what circumstances?

core/spec.md
```diff
@@ -3272,37 +3286,49 @@ for additional information.
...
+  unchanged.
+
+- Constraints:
+  - OPTIONAL
+  - When specified, MUST be a non-empty string.
+  - MUST be a read-only attribute.
+  - MUST be present when [`formatvalidated`](#formatvalidated-attribute) is
+    present with a value of `false`.
+  - MUST NOT be present when [`formatvalidated`](#formatvalidated-attribute) is
+    absent.
```
```suggestion
    absent or has a value of `true`.
```

core/spec.md
```diff
@@ -3272,37 +3286,49 @@ for additional information.
...
 
 #### `compatibilityvalidated` Attribute
-- Type: String
+- Type: Boolean
 - Description: When [`compatibility`
   validation](./model.md#groupsstringresourcesstringvalidateformat)
   is enabled, this attribute will indicate whether or not the server has
-  validated that the Version conforms to the rules defined by its Resource's
-  `meta.compatibility` attribute's value.
+  attempted to validate that the Version conforms to the rules defined by its
```
jskeet: I think it would be better to use the same wording as for formatvalidated here. (Although if formatvalidated is false, it really *hasn't* even attempted compatibility validation, let alone actually performed validation...)

core/spec.md
```diff
@@ -3220,22 +3235,20 @@ for additional information.
...
+  it adheres to the `format` attribute's rules. When `true` the Version's
+  [`formatvalidatedreason`](#formatvalidatedreason-attribute) MUST NOT be
+  present.
 
-  A value of `"true"` indicates that the server has validated the Version and
-  it adheres to the `format` attribute's rules.
-
-  A value that starts with `false` indicates that the server did not attempt
-  to validated the Version. This can happen when
+  A value of `false` indicates that the server did not attempt to validated
```
```suggestion
  A value of `false` indicates that the server did not attempt to validate
```

(And separately, as elsewhere, the "attempt" part doesn't feel quite right to me. Happy to be overruled on everything attempt-based though.)

core/spec.md
```diff
@@ -3319,10 +3345,12 @@ for additional information.
   [format_external](#format_external)) and
   reject the update request instead.
 
-  When `"false"`, this value MUST include some additional text (the `<REASON>`)
-  indicating why the server was unable to attempt validation of the Version.
+  When `false`, the
+  [`formatvalidatedreason`](#formatvalidatedreason-attribute) MUST attribute
```
```suggestion
  [`compatibilityvalidatedreason`](#compatibilityvalidatedreason-attribute) attribute
```

core/spec.md
```diff
@@ -3345,6 +3373,30 @@ for additional information.
...
+  - OPTIONAL
+  - When specified, MUST be a non-empty string.
+  - When specified, MUST be either `true` or `false`, case-sensitive.
+  - MUST be a read-only attribute.
+  - MUST be present when
+   [`compatibilityvalidated`](#compatibilityvalidated-attribute) is
+    present with a value of `false`.
+  - MUST NOT be present when
+   [`compatibilityvalidated`](#compatibilityvalidated-attribute) is
+    absent.
```
```suggestion
    absent or has a value of `true`.
```

core/spec.md
```diff
@@ -3345,6 +3373,30 @@ for additional information.
...
+
+  Due to this attribute being server managed, and influenced by the state of
+  other Versions, as its value changes the Version itself MUST NOT be marked as
+  changed - meaning, attributes such as `modifiedat` and `epoch` remain
+  unchanged.
+
+- Constraints:
+  - OPTIONAL
+  - When specified, MUST be a non-empty string.
+  - When specified, MUST be either `true` or `false`, case-sensitive.
```
```suggestion
```

jskeet (commented):
Hope these are useful. Some are definitely matters of opinion (performed vs attempted) but a few are genuinely mistakes...

core/spec.md

core/spec.md
duglin: I can live with "perform" as long as people don't read it as "did validation and failed".
To be clear, there are 3 states:
- didn't even try to check
- checked and failed
- checked and passed

These sentences are trying to talk about the first one.

jskeet: Right, but I think I'd categorize it as "tried to check, but couldn't even *do* the check due to preconditions, let alone give you a result". Is that correct, or is it *genuinely* "I made no attempt to even see whether I could check"?

core/spec.md
duglin: per the text "as its (this attribute's) value changes".

Would this read better to you:

Normally, as attribute values change, the entity itself is marked as changed (e.g. the `modifiedat` and `epoch` attributes are updated). However, since this attribute is server managed and can change due to other Versions being modified, similar to `isdefault`, changes to this attribute MUST NOT be treated as updates to the entity.

core/spec.md
duglin: good catch!

core/spec.md
jskeet: I see what you're getting at, but "changes to this attribute" still sounds like something is *trying* to change the attribute value.

Not sure of a better wording at the moment; I wonder whether some centralized notion of "computed, output-only"  would be useful.

core/spec.md
duglin: Can you elaborate? The wording already looks similar to me

duglin: Your 2nd sentence about "attempted" vs "performed" makes me wonder about your other request to use "perform" instead of "attempted".

core/spec.md
duglin: doi

core/spec.md
duglin: good catch - it's supposed to be in the previous section. Moved it.

core/spec.md
duglin: yup!

core/spec.md
```diff
@@ -3220,22 +3235,20 @@ for additional information.
   - `Avro/1.9`
 
 #### `formatvalidated` Attribute
-- Type: String
+- Type: Boolean
 - Description: When
   [`format` validation](./model.md#groupsstringresourcesstringvalidateformat)
   is enabled, this attribute will indicate whether or not the server has
   attempted to validated that the Version conforms to the rules defined by
```
jskeet: Just spotted this...
```suggestion
  attempted to validate that the Version conforms to the rules defined by
```

core/spec.md
jskeet: I'm now not quite sure what I was thinking of at the time, but I think we can probably discuss this together with the whole attempted/performed aspect.

core/spec.md
duglin: hmm I kind of like that idea. does this work for you:

- Unless otherwise stated, server managed changes to read-only attributes 
  MUST NOT be treated the same as user initiated changes to non-read-only 
  attributes with respect to updating attributes such as `modifiedat` and 
  `epoch`. In other words, as these read-only attributes are changed, the 
  entity MUST NOT be considered "updated".

Thinking about adding it to the generic Attribute and Extensions section

jskeet: Right - that sounds great to me.

core/spec.md
duglin: I don't think we draw a distinction between those two. It's a boolean thing: either we check or we don't. Whether we "don't check" because of a missing precondition or because the feature as been turned off (for the most part) doesn't really matter. The only diff is that when the feature is turned on (and strict=false) we will at least tell the user why we didn't "attempt" to check. We do this because when they have the feature enabled but the server didn't even try, the user might have a hard time following the breadcrumbs to figure out why the server didn't attempt it.

We could technically make the "reason" be present with "feature turned off" even in the "feature disabled" cases, but we're trying to avoid including flags related to disable features - to keep things from being annoyingly verbose.

jskeet: Maybe it would be simplest to remove "attempt" entirely then - just "didn't check" seems simpler than "didn't try to check".

core/spec.md
duglin: ok done - and I removed all spots (including the one for "isdefault") that said the same thing.
I did add a link to this new text though.

core/spec.md

core/spec.md

core/spec.md
duglin: That would be nice, but let me ask this... if a newbie reads:

`the server did not validate the entity`

What are the odds of someone reading that as "validation failed" ? I ask because often when someone says something like "I did not confirm that your statement is true", I think it's ambiguous as to whether they mean "I tried to confirm it but I failed to find the proof that it's true", or "I did not even try to find the proof". See my concern?

jskeet: Yup, but I think "perform validation" is then simpler than "attempt to validate". Basically the word "attempt" feels like there's more intent to it than simply "either we check or we don't".

duglin: ok check latest commit to see if it looks ok

jskeet: Yup, much better, thanks.

duglin:
Approved on the 4/23 call

PR #457, opened 2026-05-04
Week of May 4

duglin:
- lots of minor typos/fixes
- move api_not_found error into core spec so it applies to all transports, not just http


core/model.md
```diff
@@ -357,9 +357,13 @@ The following describes the attributes of the Registry model:
 - Indicates whether this attribute is REQUIRED to have a non-null value.
 - When set to `true`, this specification does not mandate how this
   attribute's value is populated (i.e. by a client, the server or via a
-  default value), just that by the end of processing any request it MUST
-  have a non-null value, and generate an error
+  default value), just that by the end of the server's processing any request
+  it MUST have a non-null value, and generate an error
   ([invalid_attribute](./spec.md#invalid_attribute)) if not.
+- Note that the previous bullet implies that a REQUIRED attribute does not
```
jskeet: There's a whole slew of negatives here. How about:

> Note that the previous bullet implies that a REQUIRED attribute does not mean that clients are mandated to include that attribute in its requests. If the attribute will automatically be populated by the server (e.g. it has a default value defined) the client may omit it.

(Also we potentially don't need an extra bullet. We could just add "Note that this implies that..." within the same bullet.)

core/spec.md
```diff
@@ -3343,7 +3353,7 @@ for additional information.
 
   Note that `false` MUST NOT be used for validation failure. In those cases
   the write operation MUST generate an error
-  ([compatibility_violation](#compatibility_violation) and reject the request
+  [compatibility_violation](#compatibility_violation) and reject the request
```
jskeet: At line 3288 you went with putting the link in parentheses - here you've removed the parentheses. Not a biggie, but I'd suggest we should be consistent.

schema/spec.md
```diff
@@ -575,9 +575,9 @@ with the declared version.
 - `Protobuf/3` is the identifier for the Protobuf syntax version 3.
 - `Protobuf/2` is the identifier for the Protobuf syntax version 2.
 
-A URI-reference, like [`schemauri`](../message/spec.md#dataschemauri that points
-to an Protobuf Schema document MUST reference an Protobuf `message` declaration
-contained in the schema document using a URI fragment suffix
+A URI-reference, like [`schemauri`](../message/spec.md#dataschemauri) that
+points to an Protobuf Schema document MUST reference an Protobuf `message`
```
jskeet:
```suggestion
points to a Protobuf Schema document MUST reference a Protobuf `message`
```

core/model.md
duglin: both suggestions adopted!

core/spec.md
duglin: good catch!

schema/spec.md
duglin: done

duglin:
Approved on the May 7 call

PR #458, opened 2026-05-06
constraint definition -  for review only

clemensv:
constraint definition for groups

### DIFF
--- a/message/model.json
+++ b/message/model.json
@@ -9,12 +9,14 @@
         "envelope": {
           "name": "envelope",
           "type": "string",
-          "description": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted. The predefined envelope format is CloudEvents/1.0"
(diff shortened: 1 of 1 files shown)


message/model.json
```diff
@@ -9,12 +9,14 @@
         "envelope": {
           "name": "envelope",
           "type": "string",
-          "description": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted. The predefined envelope format is CloudEvents/1.0"
+          "description": "Envelope format identifier that defines the common metadata information for the message. All definitions in this group share this envelope format. Mixed-envelope-format groups are not permitted. The predefined envelope format is CloudEvents/1.0",
+          "constrains": ["messages.envelope"]
```
duglin, line 13: Given this is at the model level, and in our previous chats we talked about constraints being on a per-group instance basis, are we ok with this not being group-specific?  It does satisfy the current words in the spec, but since it seems like we're no longer talking about the original usecases that causes these chats I wanted to make sure we 1) ack that we've changed topics, and 2) we're ok with no longer addressing the original discussion points. @lailabougria

duglin:
@clemensv we can close this now, right?

duglin:
closing - let me know if we need to reopen

PR #460, opened 2026-05-11
Week of May 11

### LINKED ISSUE #461: Capabilities schema
Is this ok for the schema of a Capability?

```yaml
{ 
  "<STRING>": {                    # Name of Capability attribute
    "type": "<TYPE>",
    "enum": [ <VALUE>, * ], ?      # Allowed values for scalars
    "min": <VALUE>, ?
    "max": <VALUE>, ?
    "documentation": "<URL>", ?

    "attributes": { ... }, ?       # If "type" is object
    "item": {                      # Used for non-scalar attribute types
      "type": "<TYPE>",            # Value type of map/array attributes
      "attributes": { ... }, ?     # If this item "type" is object (see above)
      "item": { ... }, ?           # If this item "type" is map,array (see above)
    }, ?
  } *
}   
```

Removed "options", and make it recursive. Similar to our model.

so now "compatibilities" looks like this:
```yaml
  "compatibilities": {
    "type": "object",
    "attributes": {
      "avro": {
        "type": "array",
        "item": { "type": "string" },
        "enum": [ "backward", "forward" ]
      },
      "protobuf": {
        "type": "array",
        "item": { "type": "string" },
        "enum": [ "backward" ]
      }
    }
  },
```

instead of this:
```yaml
   "compatibilities": {
     "type": "map",
     "item": {
       "type": "string"
     },
     "options": {
       "avro": [ "backward", "forward" ],
       "protobuf": [ "backward" ]
     }
   },
```

It used to be a "map", now it's an "object"

duglin:
- More typos
- merge "capabilities.apis" and "capabilities.mutable" into one: "capabilities.available"
  - Remove capabilities.apis & capabilities.mutable
  - Removed "options" from the capabilities schema, define it as an "object" instead of a "map" instead
- moved "api_not_found" error back to http spec. Now have "not_available" error in core (to align with "available" capability

Fixes: #461

### DIFF
--- a/cloudevents/schemas/document-schema.json
+++ b/cloudevents/schemas/document-schema.json
@@ -1486,7 +1486,7 @@
                               "type": "string",
                               "description": "The authentication/authorization type. OAuth2, Plain, APIKey, and X509Cert are well-defined"
                             },
-                            "resourceurl": {
+                            "resourceuri": {
                               "type": "string",
                               "format": "uri",
                               "description": "The resource uri for which authorization shall be granted (if applicable)"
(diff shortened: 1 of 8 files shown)


duglin:
Approved on the 5/21 call

PR #463, opened 2026-05-18
Need some text (primer?) about how we see people using doc view

### LINKED ISSUE #386: Need some text (primer?) about how we see people using doc view
single doc vs multiple doc - especially when something is checked into git.
what tooling do we expect people to create to enable this?
what tooling should WE create to help?
what's the workflow we see people using? E.g. when is RESOURCE inlined into one big file vs just URL-pointers?
How do things change in "dev mode" vs "production"? What tooling is needed for any differences? Who writes this?

### LINKED ISSUE #394: add text about "single doc docview" vs "multi-doc docview" vs "API view"
- How, and why, we support all 3?
- When to use each?
  - in particular single vs multi doc view. e.g. dev time, static file server, ...
  - When checking into git, when would single doc vs a tree of docs be better?

### LINKED ISSUE #451: How to deal with static fileservers & xreg files?
We intentionally created the xRegistry spec to support different implementation options: 
- File representation
- Static file server
- API server

While registries might be assembled in API servers using the API calls, the creation of file-based registries might happen in a more manual way, or using an AI agent. Some aspects can cause confusion in this case:

### Server-generated attributes
Several attributes in the specification, like `self` or `epoch` are not needed in this case. Nevertheless, the spec lists them as `REQUIRED`. As they are also specified as `read-only`, this implies that they are required to be set by a server implementation. 

### Redundant information
Something that may confuse someone putting together a simple single xreg file, is the redundancy that was introduced intentionally for convenience reasons. Example: attributes of the default version are replicated to resource level. 

### Document links
If an xreg file is used as central list of all kinds of metadata present in a git repository, it creates unneeded overhead to inline the documents. Instead using relative URLs to files in the repo can be useful, turning the xreg file into an `index.html` for metadata. 

The proposal is to create a dedicated description of how the spec needs to be interpreted for file-based registries. This lowers the adoption barrier for human readers and AI tools.

clemensv:
Fixes #386
FIxes #394
Fixes #451

Ths is a section about files vs directories vs APIs and federation.

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -428,7 +428,7 @@ Some xRegistry implementations, including the reference server implementation,
 might choose to provide validation capabilities for well-known formats like
 Apache Avro schema, and even version-to-version compatibility checks. These
 validation capabilities per-se are not defined in the specification, but the
-specification allows the server to provide hints about whether they are
+specification allows the server to provide indicators about whether they are
 available and whether they have been applied. Metadata declared in the models
 allows the server to validate embedded data directly.
 
@@ -470,7 +470,8 @@ kind, possibly just files in a public repo, and want to catalog those with
 xRegistry. Using the `<RESOURCE>url` patterns, all your documents can remain
 in their place. The `format` and `contenttype` information still needs to
 exist in xRegistry so that clients can understand this information before
-trying to resolve the document.
+trying to resolve the document. When a resource is registered with a URL, the
+registry will not resolve it and load the content; that is left to clients.
 
 External links also make it possible to overlay xRegistry on top of a
 completely different registry model, proxy the metadata through an xRegistry
@@ -506,13 +507,13 @@ the client fetches and caches just the branches it needs.
 
 Federation in xRegistry is built on *shadowing*: an application's view is a
 layered combination of registries, where a local registry can override or
-extend entries from an underlying one without modifying it. This lets you keep
-a local copy of selected resources that you can change freely, while still
-referring to the original registry for everything you haven't touched - with
+extend entries from an underlying one without modifying it. This allows keeping
+a local copy of selected resources that can be changed freely, while still
+referring to the original registry for everything that hasn't been touched - with
 no synchronization protocol and no risk of conflicts between different users
 or applications.
 
-If you are just dealing with files, you load files in order and merge the
+When dealing with files, applications load files in order and merge the
 content, with later files taking precedence over earlier ones.
 
 The principle applies uniformly across representations. A file can shadow
--- a/core/primer.md
+++ b/core/primer.md
@@ -16,52 +16,71 @@ normative technical details.
 
 ## 2. Table of Contents
 
-- [1. Abstract](#1-abstract)
-- [2. Table of Contents](#2-table-of-contents)
-- [3. History](#3-history)
-- [4. Value proposition](#4-value-proposition)
-  - [4.1. Why build something new?](#41-why-build-something-new)
-  - [4.2. Discovery](#42-discovery)
-  - [4.3. Vendor-agnostic](#43-vendor-agnostic)
-  - [4.4. Versioning](#44-versioning)
-  - [4.5. Schema validation](#45-schema-validation)
-  - [4.6. Payload reduction](#46-payload-reduction)
-  - [4.7. Schema-based data contract generation](#47-schema-based-data-contract-generation)
-  - [4.8. Producer and consumer generation](#48-producer-and-consumer-generation)
-  - [4.9. Basis for further developments](#49-basis-for-further-developments)
-- [5. Motivation](#5-motivation)
-- [6. Design Goals](#6-design-goals)
-  - [6.1. Non-Goals](#61-non-goals)
-- [7. Representations](#7-representations)
-  - [7.1. File](#71-file)
-  - [7.2. Static File Server](#72-static-file-server)
-  - [7.3. API Server](#73-api-server)
-- [8. Embeddings, References, and Federation](#8-embeddings-references-and-federation)
-  - [8.1. Formats and Content/Media-Types](#81-formats-and-contentmedia-types)
-  - [8.2. Embeddings and Links](#82-embeddings-and-links)
-  - [8.3. Federation](#83-federation)
-    - [8.3.1. Composing registries in memory](#831-composing-registries-in-memory)
-    - [8.3.2. Shadowing](#832-shadowing)
-    - [8.3.3. Identifiers](#833-identifiers)
-    - [8.3.4. API Servers and Proxies](#834-api-servers-and-proxies)
-- [9. Possible Use Cases](#9-possible-use-cases)
-  - [9.1. CloudEvents](#91-cloudevents)
-  - [9.2. Business objects](#92-business-objects)
-  - [9.3. Metadata files in repositories](#93-metadata-files-in-repositories)
-- [10. Design decisions or topics of interest](#10-design-decisions-or-topics-of-interest)
-  - [10.1. Resource.ID vs Resource.Version.ID](#101-resourceid-vs-resourceversionid)
-  - [10.2. Valid Characters](#102-valid-characters)
-  - [10.3. Extensions](#103-extensions)
-  - [10.4. Deleting entities](#104-deleting-entities)
-  - [10.5. Detection of Referenced Resources](#105-detection-of-referenced-resources)
-  - [10.6. Shared/Referenced Resources](#106-sharedreferenced-resources)
-- [11. The `format` attribute in the Schema spec](#11-the-format-attribute-in-the-schema-spec)
-- [12. Problematic characters in attributes](#12-problematic-characters-in-attributes)
-- [13. Why do Resources have 3 levels of data?](#13-why-do-resources-have-3-levels-of-data)
-- [14. Why are some (conditional) read-only attributes not ignored?](#14-why-are-some-conditional-read-only-attributes-not-ignored)
-- [15. Why isn't `PUT` idempotent?](#15-why-isnt-put-idempotent)
-- [16. Defining Extensions](#16-defining-extensions)
-- [17. Validation](#17-validation)
+- [xRegistry Primer](#xregistry-primer)
+  - [1. Abstract](#1-abstract)
+  - [2. Table of Contents](#2-table-of-contents)
+  - [3. History](#3-history)
+  - [4. Value proposition](#4-value-proposition)
+    - [4.1. Why build something new?](#41-why-build-something-new)
+    - [4.2. Discovery](#42-discovery)
+    - [4.3. Vendor-agnostic](#43-vendor-agnostic)
+    - [4.4. Versioning](#44-versioning)
+    - [4.5. Schema validation](#45-schema-validation)
+    - [4.6. Payload reduction](#46-payload-reduction)
+    - [4.7. Schema-based data contract generation](#47-schema-based-data-contract-generation)
+    - [4.8. Producer and consumer generation](#48-producer-and-consumer-generation)
+    - [4.9. Basis for further developments](#49-basis-for-further-developments)
+  - [5. Motivation](#5-motivation)
+  - [6. Design Goals](#6-design-goals)
+    - [6.1. Non-Goals](#61-non-goals)
+  - [7. Representations](#7-representations)
+    - [7.1. File](#71-file)
+    - [7.2. Static File Server](#72-static-file-server)
+    - [7.3. API Server](#73-api-server)
+  - [8. Embeddings, References, and Federation](#8-embeddings-references-and-federation)
+    - [8.1. Formats and Content/Media-Types](#81-formats-and-contentmedia-types)
+    - [8.2. Embeddings and Links](#82-embeddings-and-links)
+    - [8.3. Federation](#83-federation)
+      - [8.3.1. Composing registries in memory](#831-composing-registries-in-memory)
+      - [8.3.2. Shadowing](#832-shadowing)
+      - [8.3.3. Identifiers](#833-identifiers)
+      - [8.3.4. API Servers and Proxies](#834-api-servers-and-proxies)
+  - [9. Possible Use Cases](#9-possible-use-cases)
+    - [9.1. CloudEvents](#91-cloudevents)
+    - [9.2. Business objects](#92-business-objects)
+    - [9.3. Metadata files in repositories](#93-metadata-files-in-repositories)
+  - [10. Design decisions or topics of interest](#10-design-decisions-or-topics-of-interest)
+    - [10.1. Resource.ID vs Resource.Version.ID](#101-resourceid-vs-resourceversionid)
+    - [10.2. Valid Characters](#102-valid-characters)
+    - [10.3. Extensions](#103-extensions)
+    - [10.4. Deleting entities](#104-deleting-entities)
+    - [10.5. Detection of Referenced Resources](#105-detection-of-referenced-resources)
+    - [10.6. Shared/Referenced Resources](#106-sharedreferenced-resources)
+- [Default Version and Maximum Versions](#default-version-and-maximum-versions)
+- [Potential Extensions](#potential-extensions)
+- [Why Epoch?](#why-epoch)
+- [Naming and Case Sensitivity](#naming-and-case-sensitivity)
+- [Why the lower character limit on some Group and Resource type names?](#why-the-lower-character-limit-on-some-group-and-resource-type-names)
+- [Why must Group type and Resource type names be valid attribute names?](#why-must-group-type-and-resource-type-names-be-valid-attribute-names)
+- [Choosing unique Group and Resource names](#choosing-unique-group-and-resource-names)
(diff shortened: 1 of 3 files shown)


tools/dict
```diff
@@ -142,6 +141,7 @@ inlining
 interoperability
 interoperable
 isdefault
+isnt
```
duglin: let's use `isn't` instead - so drop this one

core/primer.md
```diff
@@ -246,41 +285,54 @@ Cases](#possible-use-cases) for examples outside the eventing world.
...
   depending on the registry representation.
 - **Relationships between event channels:** Focus on precisely describing a
   single event channel before standardizing the relationships between event
   channels.
 
-## Representations
+## 7. Representations
+
+An xRegistry can be represented in a few different ways: a JSON file, a static
+file server and an API server - with some further nuances. Resources will most
```
duglin: I wonder if "Resources" needs to be qualified a bit... something more like "Resource data" or "Resource's domain specific documents".  It reads to me like Resources themselves (ie. xReg Resource metadata) could live outside of the Registry and that's not true.

core/primer.md
```diff
@@ -246,41 +285,54 @@ Cases](#possible-use-cases) for examples outside the eventing world.
...
+others, like the shape of raised events and associated schemas, will have one or 
+multiple associated xRegistry files that describe those resources and that 
+these files can later be merged into another xRegistry when the module is 
+deployed, so that other participants in the system can understand the
+raised events.
+
 Writing up a registry by hand can be a great way of getting the idea, but it can
 get tedious quickly. This is when you want to use the API server.
 
-### Static File Server
```
duglin, line 283: It seems like there's a step that's missing: multiple files on disk, but not a "file server".
While having one large json file is always an option, don't we think that people will find this tedious to manage and instead create individual files for each Resource in a dir structure that matches the xReg URL structure.

core/primer.md
duglin: Maybe also include a pointer to the new section 8 which talks about "docs"?

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+However, when a model declares a resource to be of the
+["document"](spec.md#document-resources-vs-metadata-only-resources) type,
+xRegistry can indeed embed content, like schema documents. The shape of those
+documents is not declared by the model, but may be constrained by convention
+(and implementation choice) based on the [`format`](spec.md#format-attribute) attribute.
+
+### 8.1. Formats and Content/Media-Types
+
+`format` and the supported values are a convention chosen for this
+specification because the IANA "media-types" and not consistently available for
```
duglin: s/and not/are not/  maybe?

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+many document types that are in focus. There is additionally a
+[contenttype](spec.md#contenttype-attribute) attribute that SHOULD be set for
```
duglin: s/SHOULD/should/

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+each document, that provides the corresponding media-type. An example for this
+is that Apache Avro schema has no official media type, so that the content type
+might be "application/json", but the `format` is "Avro/1.12".
+
+Some xRegistry implementations, including the reference server implementation,
+might choose to provide validation capabilities for well-known formats like
+Apache Avro schema, and even version-to-version compatibility checks. These
+validation capabilities per-se are not defined in the specification, but the
+specification allows the server to provide hints about whether they are
+available and whether they have been applied.
```
duglin: We may want to say that the validation isn't just for embedded docs, but can be used to validate the xRegistry metadata itself. The "format" will define what data (Resource vs doc, or both) is checked.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+might be "application/json", but the `format` is "Avro/1.12".
+
+Some xRegistry implementations, including the reference server implementation,
+might choose to provide validation capabilities for well-known formats like
+Apache Avro schema, and even version-to-version compatibility checks. These
+validation capabilities per-se are not defined in the specification, but the
+specification allows the server to provide hints about whether they are
+available and whether they have been applied. 
+
+A registry server can only enforce validation constraints on embedded documents
```
duglin, line 436: In the case of enforcing validation constraints on embedded documents, a registry server can only act upon documents that are under its own control.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+that are under its own control. If a document is external, meaning that the
+resource entry does not embed it but links to it, the document might change at
+the external location without the registry knowing. In this case, the registry
```
duglin: Not following the "In this case..." sentence. The spec doesn't talk about following the RESOURCEurl link during an xReg GET, so we don't ever do any enforcement of externally stored docs.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+might indeed choose to refuse to return an invalid document on GET, meaning it
+performs the enforcement during retrieval.
+
+### 8.2. Embeddings and Links
```
duglin, line 441: Since the previous section talks about validation w.r.t. inlined vs externally stored docs, I wonder if it would be better to move this this section (8.2) up to be before that discussion so people fully understand "embedding and links" first.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+
+When the resource is declared to be a document, it can be stored inside the
+registry either as a directly [embedded object](spec.md#resource-attribute), a
+[binary blob](spec.md#resourcebase64-attribute), or an [external URL
+reference](spec.md#resourceurl-attribute).
+
+Strictly speaking, the choice between whether, say, a schema is represented as
+`schema` with an embedded object or `schemabase64` with base64 binary content
+depends on the encoding of the registry metadata.
```
duglin, line 450: Since you can have an embedded doc in a different format as long as it can be serialized in the registry's encoding (ie. escaped), this sentence might make it sound like you can only use today's xReg with embedded docs IFF those docs are in json. Which isn't true.

duglin: oh crap, never mind, I misread this paragraph. You may want to preface it with something like "embedded docs can be serialized using "schema" or "schemabase64"... to set the stage. Jumping start to "when to use one vs the other" w/o even mentioning there are two options feels like joining a conversation half-way thru.

core/primer.md
duglin: However, what's interesting is that our implementation does support doing a GET for the user automatically but via RESOURCEproxyurl attribute instead. Just doesn't do validation at all though.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+[cloudevents/samples](../cloudevents/samples) folder.
+
+When the registry is split across multiple files in a folder structure, the
+application can load the root document and traverse `<RESOURCE>url` links to
+individual documents on demand, building an in-memory view that combines the
+root document with only those branches that are relevant to the application.
+The same applies when the root sits behind an [API server](#73-api-server):
+the client fetches and caches just the branches it needs.
+
+#### 8.3.2. Shadowing
```
duglin, line 507: I wonder if it would make sense to use the term "proxy" in here too since that would also tie-in with the xrproxy tool.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+another API server. The application always sees the composed result as a
+single registry.
+
+#### 8.3.3. Identifiers
+
+For shadowing and cross-referencing to work, identifiers need to be stable
+across registries. Group identifiers should be treated as globally unique, so
+that the same group can be recognized across different registries. Resource
+identifiers are only unique within a group, so a cross-registry reference
+should always use the combination of group identifier and resource identifier.
```
duglin: To be clear... this "cross-registry reference" isn't a single ID (e.g. of the resource), rather you mean the URL of the Resource is a combination of globall-unique-group-ID + group-local-resource-ID, right?

duglin: Actually, when you say "Group identifiers", does that include the Group name or just the Group entity's ID? I'm assuming the group name too. We may need to make that clear so people don't confuse the two.

core/primer.md
duglin: This would then also lead into a discussion about how the Resource file on disk itself may not contain anything in it - all data is in the Version. This is actually something we may want to discuss as I think this might confuse or bother people. If I'm using "files on disk" then I would expect "something" at the Resource level and not be required to find the default version.  BUT, our ?doc flag purposely hides the duplicate info at the Resource level - which is why the file is empty in this "files on disk" case.  I'm starting to wonder if we should rethink that decision, and perhaps the Resource (on export - ?doc) should still include the default version attributes due to the usability concern.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+for the hosting registry, not a unique identifier for the resource itself. The
+same resource may be available in multiple registries, and applications can
+refer to it using the same *relative URI* regardless of which registry hosts
+it. This matters in practice: an application in a private network might not
+be able to reach a registry hosted in a public cloud, so the same content may
+need to be served from a local registry on a private DNS name. Treating the
+authority as an endpoint identifier rather than a resource identifier is what
+makes that possible.
+
+#### 8.3.4. API Servers and Proxies
```
duglin, line 557: Would it make sense to merge this with the "shadowing" section since (to me) they're all part of the same concept?

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+[contenttype](spec.md#contenttype-attribute) attribute that SHOULD be set for
+each document, that provides the corresponding media-type. An example for this
+is that Apache Avro schema has no official media type, so that the content type
+might be "application/json", but the `format` is "Avro/1.12".
+
+Some xRegistry implementations, including the reference server implementation,
+might choose to provide validation capabilities for well-known formats like
+Apache Avro schema, and even version-to-version compatibility checks. These
+validation capabilities per-se are not defined in the specification, but the
+specification allows the server to provide hints about whether they are
```
jskeet: "Hint" feels a bit vague/uncertain to me. Can a server not be definitive about it? ("Valid, invalid, I can't check" etc)

core/primer.md
jskeet: Yes, I think that would be useful.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+
+You can also refer to the target document externally using `schemaurl` and
+point to a location from where that document can be directly obtained wither
+using an HTTP GET or resolving a file system file.
+
+External links are very handy if you already have an existing registry of some
+kind, possibly just files in a public repo, and want to catalog those with
+xRegistry. Using the `<RESOURCE>url` patterns, all your documents can remain 
+in their place. The `format` and `contenttype` information still needs to 
+exist in xRegistry so that clients can understand this information before
```
jskeet, line 473: It's slightly unclear whether the intention is that when a document is represented by a link, the client is expected to fetch it (which this paragraph implies) or whether the xRegistry server should fetch and then serve it (as is somewhat implied by the next paragraph, and by the earlier comment about RESOURCEproxyurl.

Maybe it would be worth clarifying that here.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+both a document format model and an API, the integration of multiple registries
+("federation") does not rely on making copies but is intended to work via
+cross-referencing and "shadowing" across the [representations](#7-representations)
+described earlier.
+
+#### 8.3.1. Composing registries in memory
+
+A registry document, whether a single [file](#71-file) or the root of a
+[static file server](#72-static-file-server) layout, is meant to be loaded into
+an application as a local, in-memory registry. While a document technically
```
jskeet, line 493: What kind of application are we talking about here? I wouldn't expect most client applications to want to fetch "the registry".

If this is more about this being *possible*, I think some rewording is required.

core/primer.md
```diff
@@ -318,14 +370,165 @@ generate files and split them up in directories like in the static file server
...
+individual documents on demand, building an in-memory view that combines the
+root document with only those branches that are relevant to the application.
+The same applies when the root sits behind an [API server](#73-api-server):
+the client fetches and caches just the branches it needs.
+
+#### 8.3.2. Shadowing
+
+Federation in xRegistry is built on *shadowing*: an application's view is a
+layered combination of registries, where a local registry can override or
+extend entries from an underlying one without modifying it. This lets you keep
```
jskeet: The use of "you" here and below feels a little odd.

jskeet (commented):
A few comments, but I admit I got a bit lost in the discussion of shadowing. I do wonder how many readers are going to want/need that information.

duglin:
Conditionally approved on 5/2 call - will merge once comments are addressed

core/primer.md
clemensv: I'd like to keep that separate. Shadowing also applies to files.

core/primer.md
```diff
@@ -16,52 +16,71 @@ normative technical details.
...
+    - [9.2. Business objects](#92-business-objects)
+    - [9.3. Metadata files in repositories](#93-metadata-files-in-repositories)
+  - [10. Design decisions or topics of interest](#10-design-decisions-or-topics-of-interest)
+    - [10.1. Resource.ID vs Resource.Version.ID](#101-resourceid-vs-resourceversionid)
+    - [10.2. Valid Characters](#102-valid-characters)
+    - [10.3. Extensions](#103-extensions)
+    - [10.4. Deleting entities](#104-deleting-entities)
+    - [10.5. Detection of Referenced Resources](#105-detection-of-referenced-resources)
+    - [10.6. Shared/Referenced Resources](#106-sharedreferenced-resources)
+- [Default Version and Maximum Versions](#default-version-and-maximum-versions)
```
duglin: this indentation doesn't seem right, and the html rendering looks odd

core/primer.md
clemensv: We have a lot of file-based examples where "the registry" is that one file.

core/primer.md
```diff
@@ -318,14 +394,168 @@ generate files and split them up in directories like in the static file server
...
+[cloudevents/samples](../cloudevents/samples) folder.
+
+When the registry is split across multiple files in a folder structure, the
+application can load the root document and traverse `<RESOURCE>url` links to
+individual documents on demand, building an in-memory view that combines the
+root document with only those branches that are relevant to the application. The
+same applies when the root sits behind an [API server](#73-api-server) or proxy:
+the client fetches and caches just the branches it needs.
+
+#### 8.3.2. Shadowing
```
duglin, line 507: This paragraph reads like a feature of the spec, but there's nothing in the spec about this. Do we need to make it clear that this is an implementation choice?

In these cases, how are all of the various requirements enforced? E.g. with shadowing, how are the new "constraints" stuff we're added enforced and what happens when they're not valid?

core/primer.md
```diff
@@ -318,14 +394,168 @@ generate files and split them up in directories like in the static file server
...
+A registry document, whether a single [file](#71-file) or the root of a
+[static file server](#72-static-file-server) layout, is meant to be loaded into
+an application as a local, in-memory registry. While a document technically
+needs to embed a model to be valid, servers often choose to allow that the
+models they know about are implicitly available, so that a registry document
+can just contain the resources and their metadata, without the model. There
+are several examples of such documents in the
+[cloudevents/samples](../cloudevents/samples) folder.
+
+When the registry is split across multiple files in a folder structure, the
```
duglin, line 500: Has anyone actually tried to do this with git?
See: https://github.com/xregistry/spec/pull/463#discussion_r3282116252
I'm not sure what the usability looks like in these cases because the Resources themselves are basically empty, any meaningful interactions needs to be done with a Version, and I'm not sure if that's a problem or not. But it seems like one.

core/primer.md
```diff
@@ -318,14 +394,181 @@ generate files and split them up in directories like in the static file server
...
+application can load the root document and traverse `<RESOURCE>url` links to
+individual documents on demand, building an in-memory view that combines the
+root document with only those branches that are relevant to the application. The
+same applies when the root sits behind an [API server](#73-api-server) or proxy:
+the client fetches and caches just the branches it needs.
+
+#### 8.3.2. Shadowing
+
+Federation in xRegistry is intended to be built on *shadowing*: an application's
```
lailabougria, line 509: This is an interesting section. I don’t remember discussing this in the meetings, unless it was when I wasn’t around. The way it’s currently framed is quite opinionated, which is why I’m asking: is this something we decided, and if so, is that represented in the rest of the spec? If not, should it really be framed this “pressing”?

duglin:
@clemensv a rebase is needed

duglin:
@clemensv since your other PR touched the primer, a rebase is needed

duglin:
@clemensv CI is unhappy

PR #466, opened 2026-05-19
Primer: explain Groups as a single organizational dimension

clemensv:
Adds a new "Group organization" subsection to the primer's *Design decisions or topics of interest* section, explaining:

- Groups are intentionally a single organizational dimension above Resources; Groups don't contain other Groups, by design (model/URL/filter/AC/tooling complexity vs. expressive value).
- Where hierarchical organization is useful, dot-notation in Group identifiers (e.g. `Contoso.ERP.Orders`) provides hierarchy semantics over a flat collection, including prefix-based filtering.
- The orthogonal organizational tool is `labels` — free-form name/value pairs on Registries, Groups, Resources, Meta, and Versions. Server implementations MAY (and are encouraged to) support filtering and lookup by labels, complementing the single Group dimension for cross-cutting discovery.

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -464,7 +464,7 @@ The second, orthogonal organizational tool is `labels`. Labels are
 free-form name/value pairs that can be attached to Registries, Groups,
 Resources, Meta entities, and Versions, and they allow entities to be
 classified along as many independent axes as needed (for example
-`stage=dev`, `team=payments`, `tier=gold`). Server implementations MAY,
+`stage=dev`, `team=payments`, `tier=gold`). Server implementations may,
 and are encouraged to, support filtering and lookup by labels (see the
 filter examples in the core spec), which makes labels a powerful
 complement to the single Group dimension for cross-cutting discovery
--- a/core/primer.md
+++ b/core/primer.md
@@ -442,6 +442,34 @@ variable or structure property names - they're usually just stored as
 - how MIGHT someone implement multi-tenancy with xRegistry
   - and layers of xRegistries to get multi-level grouping ??
 
+### Group organization
+
+Groups are intentionally a single organizational dimension above Resources:
+the model is always Registry → Group → Resource → Version, and Groups
+themselves do not contain other Groups. This is a deliberate design choice.
+Allowing arbitrary nesting of Groups would significantly complicate the
+model, the URL space, traversal and filter semantics, access-control
+reasoning, and tooling, without adding expressive power that cannot be
+achieved by other means.
+
+Where a hierarchical organization is useful, xRegistry instead relies on
+dot-notation in Group identifiers. Because Group IDs may contain dots, an
+identifier like `Contoso.ERP.Orders` naturally expresses a logical
+hierarchy (`Contoso` → `ERP` → `Orders`) while keeping the underlying
+collection flat. This gives implementers and users the readability and
+grouping behavior of a hierarchy — including prefix-based filtering — without
+the structural complexity of nested containers.
+
+The second, orthogonal organizational tool is `labels`. Labels are
+free-form name/value pairs that can be attached to Registries, Groups,
+Resources, Meta entities, and Versions, and they allow entities to be
+classified along as many independent axes as needed (for example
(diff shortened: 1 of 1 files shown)


core/primer.md
```diff
@@ -442,6 +442,34 @@ variable or structure property names - they're usually just stored as
...
+hierarchy (`Contoso` → `ERP` → `Orders`) while keeping the underlying
+collection flat. This gives implementers and users the readability and
+grouping behavior of a hierarchy — including prefix-based filtering — without
+the structural complexity of nested containers.
+
+The second, orthogonal organizational tool is `labels`. Labels are
+free-form name/value pairs that can be attached to Registries, Groups,
+Resources, Meta entities, and Versions, and they allow entities to be
+classified along as many independent axes as needed (for example
+`stage=dev`, `team=payments`, `tier=gold`). Server implementations MAY,
```
duglin: s/MAY/may/

duglin:
As the spec stands today this text is fine.

However, it does make me wonder if people will then ask why we have Groups at all? If people can use dot-notation in Resource names, and use labels, why even have Groups if they don't provide any semantic value? For example, many systems use groups, not just for "grouping", but also for (at least) two other purposes: 1) policy enforcement, where the system will only allow items in the Group that conform with the requirements of the Group, or 2) allows for "bulk" settings of aspects of the items in the group. For example, today people would need to set each Resource's "compatibility" attribute to the desired value. Having this flexibility at the Resource level is good, but it's also a usability issue because now each time someone adds a new Resource they need to remember to set that value. If the Group defined a default value for all Resources then the usability concern decreases. 

When I look at Confluent's schema registry and Apicurio, I see Grouping concepts there that align with what I'm describing. They're not just "bags of stuff" without semantics attached to them. Instead, they allow for people to do things like define "compatibility" rules above individual Resources, but are not necessarily Registry-wide. Confluent: subjects, contexts. Apicurio: just Groups - but you can override global settings (e.g. compat) at the Group level.

Right now our Groups kind of leaves us in a weird state. They're not as good as a normal "bag of stuff" like a generic directory structure because you can only have one level (that would be ok IF they added some other value... like semantics, which they don't). But, they also don't provide some other value (like group-wide default, semantics, policy enforcement... that might be hard to support in a multi-level feature), so why have them at all when (as you state in this PR) you can (sort of) get the same value as multi-level-groups (even one level) via labels and dot-notation-names on Resources? 

If nothing else, I think we need to discuss xReg's inability to support Confluent's and Apicurio's features/usecases if we are hoping to be a replacement for those offerings.

duglin:
LGTM for now, but I think we should discuss some stuff

core/primer.md
```diff
@@ -442,6 +442,34 @@ variable or structure property names - they're usually just stored as
...
+
+Groups are intentionally a single organizational dimension above Resources:
+the model is always Registry → Group → Resource → Version, and Groups
+themselves do not contain other Groups. This is a deliberate design choice.
+Allowing arbitrary nesting of Groups would significantly complicate the
+model, the URL space, traversal and filter semantics, access-control
+reasoning, and tooling, without adding expressive power that cannot be
+achieved by other means.
+
+Where a hierarchical organization is useful, xRegistry instead relies on
```
jskeet, line 455: This sounds like it's normative, something that implementers need to be aware of. Is that actually the case, or is it just a convention.

To put it another way: if one user decided to use underscores such as Contoso_ERP_Orders, would you expect a standard implementation to make that harder than using dots?

jskeet (commented):
Just one question, generally seems fine to me. (Not *sure* whether I agree with the design decision, but that's separate to whether this PR expresses it appropriately, and I'm certainly not trying to reopen the design debate.)

duglin:
Conditionally approved on 5/2 call - will merge once comments are addressed

core/primer.md
clemensv: It's a convention for the most part. The reason I even wrote the section is that I got asked more than once why we don't support nested groups. My short answer is "too complicated". The long answer is this section.

duglin:
All ready now! Merging

PR #468, opened 2026-05-22
Remove ur*-reference types

duglin:
And make sure we support all uri/uri variants

Hot fix since our tooling breaks w/o this.


duglin:
chicken and egg problem - merging so tools are updated. Will rerun tests after new tools are available.

PR #470, opened 2026-05-27
Week of May18

duglin:
- more typos
- clarify ordering of processing when capabilities, modelsource and metadata are all present in one request
- add text about doc translation to README - similar to CloudEvents

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -1627,39 +1627,6 @@ The serialization of the Registry entity MUST adhere to this form:
 }
 ```
 
-Write operations directed to the Registry root, when it includes the
-`capabilities` or `modelsource` attributes, SHOULD follow these general rules
-to help promote consistency across implementations:
-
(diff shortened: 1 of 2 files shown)


core/spec.md
```diff
@@ -1674,6 +1674,14 @@ and the following Registry-level attributes:
     of the Registry.
   - An explicit value of `null` for this attribute MUST result in resetting the
     capabilities to the server's default values.
+  - If present, MUST be processed before any other attributes since subsequent
```
jskeet, line 1677: This sounds like it's saying "for some capabilities/attributes, use the settings before the request; for other capabilities/attributes, use the settings within the request".

That may well be appropriate, but an example would be really useful.

duglin: good idea. done

jskeet (approved):
Thanks, the example makes a big difference.

duglin:
Approved on the 5/28 call

PR #475, opened 2026-06-10
Remove some sticky fields

### LINKED ISSUE #472: Possible issue around stickyversions
In capabilities we have: `stickyversions: BOOLEAN` - indicates if the server supports sticky versions

In model.Resource we have: `setdefaultversionsticky: BOOLEAN` - indicates if this Resource allows sticky versions

In Resource instance we have: `meta.defaultversionsticky: BOOLEAN` - indicates if this Resource instance's default version is sticky.

When capabilities.stickyversion=false, then model.resource.setdefaultversionsticky=false by default AND MUST only be "false"
When capabilities.stickyversion=true, then model.resource.setdefaultversionsticky=true by default

1 - Is it weird that the default value changes based on the capability?  Proposal: remove it from capabilities

2 - when capabilities.stickyversion is set to true, existing Resource instances's meta.defaultversionsticky values will remain `false`. Only new Resource will have their `meta.defaultversionsticky` set (defaulted) to `true`. ok?

### LINKED ISSUE #473: Remove resourceModel.setdefaultversionsticky
When trying to set resourceModel.setdefaultversionsticky to "false", you can get the same net effect by modifying the "meta.defaultversionsticky" flag to have an enum set of just "false".

```
{
  "groups": {
    "dirs": {
      "singular": "dir",
      "resources": {
        "files": {
          "singular": "file",
          ...
          "setdefaultversionsticky": false,    <===
          ...
          "metaattributes": {
            "defaultversionsticky": {
              "enum": [ false ],               <===
              "type": "boolean",
              "required": true,
              "default": false
            }
          }
        }
      }
    }
  }
}
```

duglin:
- Remove the stickyversions capability and resourcemodel.setdefaultversionsticky flags.
  - Talk about to edit the "defaultversionsticky" spec attribute instead.
- Remove "consistentformat" in favor of the new "constraint" feature
  - Add "format" to the schemagroup (in schema registry) - should have been part of weekly PR.

Ignore the first commit - that's actually from the weekly PR.

Fixes #472 
FIxes #473

### DIFF
--- a/cloudevents/schemas/document-schema.avsc
+++ b/cloudevents/schemas/document-schema.avsc
@@ -715,6 +715,10 @@
               "doc": "Time of the object modification"
             },
             {
+              "type": "string",
+              "name": "format"
+            },
+            {
               "name": "Extensions",
               "type": {
                 "type": "map",
--- a/cloudevents/schemas/document-schema.json
+++ b/cloudevents/schemas/document-schema.json
@@ -2187,6 +2187,9 @@
             "format": "date-time",
             "description": "Time of the object modification"
           },
+          "format": {
+            "type": "string"
+          },
           "schemas": {
             "type": "object",
             "additionalProperties": {
--- a/cloudevents/schemas/openapi.json
+++ b/cloudevents/schemas/openapi.json
@@ -5351,6 +5351,9 @@
             "format": "date-time",
             "description": "Time of the object modification"
           },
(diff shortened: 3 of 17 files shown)


duglin:
Approved on the 6/11 call but will use #478 to decide if we really want to remove "consistentformat" or not

PR #479, opened 2026-06-12
Week of June 15

### LINKED ISSUE #477: is "equals" case insensitive for strings?
To align with "consistentformat", which we just removed, I think it should be.

### LINKED ISSUE #478: is "constraints" really what we want to use instead of "consistentformat" ?
"constraints" are group-wide while "consistentformat" is scoped to a Resource

duglin:
- add support for "matchversions" (and add to schema spec's model)
- make "constraints.equals" use "matchcase" when doing string compares
- clarity that setting maxversions=1 when there are sticky versions -> error
- tell people who to require sticky versions (set defaultversionsticky'e enum/default to "true")

Fixes #478 
Fixes #477

### DIFF
--- a/core/model.md
+++ b/core/model.md
@@ -2,7 +2,7 @@
 
 <!-- words: compat validatecompatibility validateformat strictvalidation -->
 <!-- words: matchcase compatibilityvalidated formatvalidated -->
-<!-- words: validators -->
+<!-- words: validators matchversions -->
 
 ## Abstract
 
@@ -84,6 +84,7 @@ The overall format of a model definition is as follows:
       "enum": [ <VALUE> * ], ?         # Array of scalars of type "<TYPE>"
       "strict": <BOOLEAN>, ?           # Just "enum" values or not. Default=true
       "matchcase": <BOOLEAN>, ?        # Strings case-sensitive? Default=false
+      "matchversions": <BOOLEAN>, ?    # Same for all Versions? Default=false
       "readonly": <BOOLEAN>, ?         # From client's POV. Default=false
       "immutable": <BOOLEAN>, ?        # Once set, can't change. Default=false
       "required": <BOOLEAN>, ?         # Default=false
@@ -122,7 +123,7 @@ The overall format of a model definition is as follows:
       "constraints": {
         "<RESOURCES>.<PATH>": {          # Resource-plural + attribute path
           "default": <VALUE>, ?          # Group specific default
-          "enum": [ <VALUE> * ], ?       # Allowed subset of values
+          "enum": [ <VALUE> + ], ?       # Allowed subset of values
           "equals": "<PATH>" ?           # Matching Group attribute path
         } *
       }, ?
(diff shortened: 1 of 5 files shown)


core/model.md
```diff
@@ -324,10 +325,30 @@ The following describes the attributes of the Registry model:
...
   a matching value in a case-sensitive way, or not.
-- This attribute MUST NOT be `true` if the owning attribute's `type` (or
+- This aspect MUST NOT be `true` if the owning attribute's `type` (or
   `item.type` for non-scalars) is not `"string"`.
 - When not specified, the default value MUST be `false`.
 
+### `attributes.<STRING>.matchversions`
+- Type: Boolean.
+- OPTIONAL.
+- Indicated whether all Versions of a Resource instance MUST have the same
```
jskeet:
```suggestion
- Indicates whether all Versions of a Resource instance MUST have the same
```

(I expect, anyway. It's odd to be in the past tense.)

core/model.md
```diff
@@ -324,10 +325,30 @@ The following describes the attributes of the Registry model:
...
+  ([mismatched_version_attribute](spec.md#mismatched_version_attribute)) MUST
+  be generated.
+- This aspect MUST only be `true` for attribute definitions that:
+  - Are within a [Resource type's versioned
+    attributes](#groupsstringresourcesstringattributes) section.
+  - Are for scalar typed attributes.
+  - Are statically defined. Meaning, not defined as part of a `ifvalues` clause
+    or via a `*` extension definition
+  - Are not defined within an array or map. Within an object is allowed.
+- In the case of the attribute type being a `string`, the comparison MUST take
```
jskeet, line 347: As in, if matchcase is false, then it's okay one version to have a value of "ABC" and another to have a value of "abc" without any errors being generated?

core/spec.md
```diff
@@ -3137,6 +3141,10 @@ use the following definition:
 }
 ```
 
+Likewise, forcing all Resource instances to have "sticky" Versions MAY be
+done via the same mechanism but using `true` instead of `false` in the
```
```suggestion
achieved via the same mechanism but using `true` instead of `false` in the
```

Maybe? It just feels a little more spec-suitable than "done".

jskeet (approved):
A few small comments, but fine as far as I understand it...

core/model.md

core/model.md
duglin: yes. Do you think I need to be more clear/verbose on that?

core/spec.md
duglin: done :-)

core/model.md
jskeet: Nope, I think it's fine - just thought I'd double-check, as if it *didn't* mean that, we'd need to be clearer :)

duglin: Perfect!  Thanks SOOOO much for the quick review!

duglin:
Approved off-line

PR #481, opened 2026-06-21
make rc3

duglin:


duglin:
Merging despite errors - it's due to the rc3 tag not existing yet.

PR #484, opened 2026-06-25
Week of June 22

### LINKED ISSUE #483: document our dot-notation
We use dot-notation for : filters, constraints, matchversions inline and sort

It's based on JSONPath:
```
name             # top-level attr
['name']         # same, but allows for dot in the name
name.name1       # subobject attr
name.15          # same as previous, "15" is treated like a name/string
name['name1']    # useful for when name1 includes a dot
name[int]        # for arrays, index "int" into "name" array
name[int].name1  # "name1" attr in index "int" into "name" array
```

Filters:
- doesn't support ENTITY_IDs as it traverses the hierarchy - e.g. `GROUP.RESOURCE` (not `GROUP.GROUPID.RESOURCE`)
- but, once in an entity, you need to specify a map key to traverse into the map
- don't support arrays at all right now in the impl - no special mention of arrays in the spec
- Q: should we support arrays? By index seems weird. Search over all index might be the best option
- Q: for maps, we support specifying the key (similar to objects), should we support a "search over all" variant?

Constraints & MatchVersions:
- we only support scalars and objects

Inline:
- only used for specifying xReg entities: groups, resources, versions, meta, model, RESOURCE,... but not other attributes

Sort:
- only available for xReg collections
- only points to scalar attribute in the top collection entities, not any nested xReg entities (e.g. resource.versions)
- no mention of objects, maps or arrays - but the impl supports objects, maps, array - but need to specify map key and array index ( I think ). More investigation/thinking needed here

duglin:
- minor wording tweaks to better explain the new "constraints" stuff
- document our dot-notiation patterns

Fixes #483

### DIFF
--- a/core/model.md
+++ b/core/model.md
@@ -123,7 +123,7 @@ The overall format of a model definition is as follows:
       "constraints": {
         "<RESOURCES>.<PATH>": {          # Resource-plural + attribute path
           "default": <VALUE>, ?          # Group specific default
-          "enum": [ <VALUE> + ], ?       # Allowed subset of values
+          "enum": [ <VALUE> * ], ?       # Allowed subset of values
           "equals": "<PATH>" ?           # Matching Group attribute path
         } *
       }, ?
@@ -575,6 +575,25 @@ The `constraints` map defines a set of rules that can be used to govern
 the attribute values in Resources that are added to instances of the Group
 type being defined.
 
+These restrictions are designed to only allow subsetting of the constraints
+specified by the Resource type model. They MUST NOT be used to extend the
+allowable values of the attributes being constrained.
+
+The format of a constraint map MUST adhere to the following:
+
+```yaml
+"constraints": {
+  "<RESOURCES>.<PATH>": {          # Resource-plural + attribute path
+    "default": <VALUE>, ?          # Group specific default
+    "enum": [ <VALUE> * ], ?       # Allowed subset of values
+    "equals": "<PATH>" ?           # Matching Group attribute path
+  }
+}
+```
+
+More details about each part of the constraints are defined in the following
+sections.
+
 These constraints MUST be applied to all instances of this Group type. Group
 instances MAY choose to add additional entries, or further restrict the ones
 defined here, via use of the Group instance's
@@ -588,9 +607,9 @@ detected.
 When possible these constraints are best described when defining the Resource
 attributes in question, rather than here. However, there are cases where this
 might not be possible, such as when Resources are added to a Group type via the
-`ximportresources` feature or defining a relationship between Group and
-Resource attributes - such is the case for the `equals` constraint defined
-below.
+`ximportresources` feature, Resource are added to Groups via the `xref`
+features, or defining a relationship between Group and Resource attributes -
+such is the case for the `equals` constraint defined below.
 
 Inclusion of a Resource in a Group instance via the
 [`meta.xref`](./spec.md#cross-referencing-resources) mechanism introduces
@@ -608,7 +627,7 @@ a constraint are applied to such xref'd Resources:
 
 This map key MUST reference the Resource attribute that is to be constrained.
 It MUST reference a scalar attribute (top-level, or nested within objects) but
-but MUST NOT reference items in arrays or maps). It MUST only reference
+MUST NOT reference items in arrays or maps. It MUST only reference
 statically-defined attributes, not ones that are dynamically added via an
 `ifvalues` clause or via a `*` extension definition.
 
@@ -621,9 +640,10 @@ the Resource attribute being constrained.
 ### `groups.<STRING>.constraints.<RESOURCES>.<PATH>.default`
 
 This aspect defines the default value that MUST be used for the referenced
-Resource attribute. This MUST override any default value specified in the
-model for that attribute. See the
-[attribute `default` aspect](#attributesstringdefault) for additional
+Resource attribute. This Group type value MUST override any default value
+specified in the model for that attribute.
+
+See the [attribute `default` aspect](#attributesstringdefault) for additional
 information concerning default value processing, as they apply here as well
 with one exception: adding a default value here does not mandate that the
 referenced Resource attribute's `required` aspect be set to `true`.
@@ -641,7 +661,7 @@ This aspect defines the set of values that the referenced Resource attribute
 MUST be restricted to regardless of whether the Resource's
 [`strict`](#attributesstringstrict) attribute is set to `true` or not. The
 list's values MUST be valid per the Resource attribute's model definition
-(e.g. a proper subset of any `enum` defined if the Resource's `enum` is
+(i.e. a proper subset of any `enum` defined if the Resource's `enum` is
 [`strict`](#attributesstringstrict), and of the same type).
 
 As with the [`enum` attribute](#attributesstringenum) defined for attributes,
@@ -649,14 +669,18 @@ an empty `enum` list in a constraints MUST be treated the same as no `enum`
 aspect at all and no further constraints on the allowable attribute values are
 applied beyond what the attribute itself defines.
 
-If an `enum` set is defined, but a `default` values is not, then any `default`
-value specified in the attribute itself MUST be part of the `enum` set.
+If an `enum` set is defined, but a constraint `default` values is not, then
+any `default` value specified in the attribute itself MUST be part of the
+`enum` set.
 
 ### `groups.<STRING>.constraints.<RESOURCES>.<PATH>.equals`
 
-Use of this aspect within a constraint MAY be used to ensure that all
-Resource instances within a Group instance have the same value for the
-specified attribute in all of their Versions.
+Use of this aspect within a constraint MAY be used to ensure that the specified
+attribute in all Versions of all Resources within a Group instance is present
+with the same value as the specified `equals` attribute in the Group instance.
+
+When specified with an empty string, it MUST be treated as if it were not
+specified at all.
 
 When present, this aspect MUST contain the dot (`.`) notation path in the
 Group instance that the referenced Resource attribute MUST match. In the
@@ -666,18 +690,12 @@ account the [matchcase](#attributesstringmatchcase) aspect of the attribute.
 If the referenced Group attribute does not exist, then the `equals` constraint
 enforcement for the Resource attribute MUST be silently ignored.
 
-If, after any potential `default` processing logic is performed, the Resource
-attribute (in any of its Versions) does not exist but the Group attribute does
-exist, then an error ([constraint_failure](./spec.md#constraint_failure)) MUST
-be generated.
-
 This attribute MUST reference a statically defined Group attribute. In other
 words, it can not reference an attribute defined by an `ifvalues` clause or a
 `*` extension definition. Nor can it reference an attribute within an array.
 
-Note that if a Resource type's versioned attribute has a `matchversions` aspect
-set to `false`, then this feature will have the same net affect as
-`matchversions` for that attribute being `true`.
+Note that this feature has similar results as setting the Resource attribute's
+`matchversions` aspect being set to `true`.
 
 ### `groups.<STRING>.resources`
 - Type: Map where the key MUST be the plural name (`groups.resources.plural`)
--- a/core/model.md
+++ b/core/model.md
@@ -634,7 +634,8 @@ statically-defined attributes, not ones that are dynamically added via an
 The `<RESOURCES>` portion of the map key MUST be the plural name of the
 Resource type being referenced.
 
-The `<PATH>` portion of the map key MUST be a dot (`.`) notional traversal to
+The `<PATH>` portion of the map key MUST be a
+[dot (`.`) notional](spec.md#xregistry-dot--notation) traversal to
 the Resource attribute being constrained.
 
 ### `groups.<STRING>.constraints.<RESOURCES>.<PATH>.default`
@@ -682,7 +683,8 @@ with the same value as the specified `equals` attribute in the Group instance.
 When specified with an empty string, it MUST be treated as if it were not
 specified at all.
 
-When present, this aspect MUST contain the dot (`.`) notation path in the
+When present, this aspect MUST contain the
+[dot (`.`) notation](spec.md#xregistry-dot--notation) path in the
 Group instance that the referenced Resource attribute MUST match. In the
 case of the attribute type being a `string`, the comparison MUST take into
 account the [matchcase](#attributesstringmatchcase) aspect of the attribute.
--- a/core/model.md
+++ b/core/model.md
@@ -635,7 +635,7 @@ The `<RESOURCES>` portion of the map key MUST be the plural name of the
 Resource type being referenced.
 
 The `<PATH>` portion of the map key MUST be a
-[dot (`.`) notional](spec.md#xregistry-dot--notation) traversal to
+[dot (`.`) notation](spec.md#xregistry-dot--notation) traversal to
 the Resource attribute being constrained.
 
 ### `groups.<STRING>.constraints.<RESOURCES>.<PATH>.default`
--- a/core/spec.md
+++ b/core/spec.md
@@ -558,7 +558,7 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
       "constraints": {
         "<RESOURCES>.<PATH>": {                # Resource-plural + attr path
           "default": <VALUE>, ?                # Group specific default
-          "enum": [ <VALUE> + ], ?             # Allowed subset of values
+          "enum": [ <VALUE> * ], ?             # Allowed subset of values
           "equals": "<PATH>" ?                 # Matching Group attribute path
         } *
       }, ?
@@ -1424,9 +1424,9 @@ of the existing entity. Then the existing entity would be deleted.
   allowed values for a particular attribute. This allows for Group instance
   specific restrictions without requiring the creating of new Resource types.
 
-  These restrictions are designed to only allow subsetting of the constraints
-  specified by the Resource type model and any Group model
-  [`constraints`](model.md#groupsstringconstraints) defined. They can not be
+  These restrictions are designed to only allow subsetting of the definitions
+  specified by the Resource type model and any Group type model
+  [`constraints`](model.md#groupsstringconstraints) defined. They MUST NOT be
   used to extend the allowable values of the attributes being constrained.
 
   The definition of this map is the same as the Group type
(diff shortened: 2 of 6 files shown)


core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+
+This specification defines a syntax for referencing entities and attributes
+within the xRegistry metadata. Known as "dot notation", it borrows the some
+of the core aspects of the [JSON Path](https://www.rfc-editor.org/info/rfc9535)
+specification.
+
+The following notation operators are defined:
+
+- `.NAME` : access a property (or key) of an object (or map).
+- `['NAME']` : same as previous but `NAME` contains characters that are not
```
jskeet, line 4467: This has single quotes, but in the examples there are double quotes. Are both valid, or is one of these a mistake?

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+Note: due to the serialization of maps and objects (often) being
+indistinguishable, this specification (similar to JSONPath) does not provide
+different syntaxes for each.
+
+**Examples:**
+
+- `.name`          # An attribute (or map key) called "name"
+- `['my.name']`    # An attribute (or map key) called "my.name"
+- `[2]`            # The 3rd (zero-based) item in an array
+- `["birth.data]"` # Attribute/key called "birth.date"
```
jskeet: Nit: data => date

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+  to be interpreted as an operator (e.g. `.`) or it contains just digits and
+  is not to be interpreted as an array index. Double-quotes MAY be used as
+  well.
+- `[INTEGER]` : access an array index.
+
+Note: due to the serialization of maps and objects (often) being
+indistinguishable, this specification (similar to JSONPath) does not provide
+different syntaxes for each.
+
+**Examples:**
```
jskeet, line 4481: Possibly an example of double-dereferencing, assuming that makes sense? e.g. `employee.address.state` or `employee.address["state"]`

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+
+Depending on the situation in which the notation is being used, there are
+certain special values that MAY be used, a described in the following sections.
+
+### Dot-Notation in Filters
+
+When using [filters](#filter-flag) the following special values MAY be used::
+- `.*`             # Match any item in a collection of objects/maps.
+- `[*]`            # Match any item in an array.
+- `[-1]`           # Match the last item in an array.
```
jskeet: Does this extend to other indexes, e.g. `[-2]` matching the penultimate item in an array?

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+- `stack[3]`       # The 3rd (zero-based) item in an array called "stack"
+
+Depending on the situation in which the notation is being used, there are
+certain special values that MAY be used, a described in the following sections.
+
+### Dot-Notation in Filters
+
+When using [filters](#filter-flag) the following special values MAY be used::
+- `.*`             # Match any item in a collection of objects/maps.
+- `[*]`            # Match any item in an array.
```
jskeet: Does this also match any item in an object/map?

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+    "owner": <STRING>,
+    "reviewers": [ <STRING>, * ]                # An array of strings
+    "labels": { "<STRING>": "<STRING>" * }, ?   # A map of string -> string
+  },
+  "age": <INTEGER>
+}
+```
+
+- Find all schemagroups that have an `age` of `5`:
+  `GET /?filter=schemagroups.age=5`
```
jskeet, line 4536: Everything else has `info`, but this doesn't... is that a typo? (You later refer to `info.age`.)

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+  then they would most likely have specified that as part of the request URL.
+  For example, `GET /schemagroups/mygroup/...` - in which case the ID would
+  not appear as part of the filter expression at all.
+
+As a result, when specifying a filter expression that steps between the
+xRegistry hierarchy, the `ID` portion of the path is excluded. Note that this
+does not apply to stepping through the path of objects/maps/arrays defined
+within an xRegistry entity (Group, Resource, Versions). In those cases use
+of the `*` wildcard MAY be used to indicate "any" items in that set. As
+shown in the "info" examples above.
```
jskeet: The only examples using are the last two, which are leaf matches. Perhaps we need another example where the `*` is not a leaf, if that's valid?

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+It is worth noting that if the user really does want a certain schemagroup with
+a certain id (e.g. `mygroup`), without the request URL path being
+`/schemagroups/mygroup`, then they can achieve this by using a compound
+filter:
+
+```yaml
+GET /?filter=schemagroups.schemas.name=myschema,schemagroups.schemagroupid=mygroup
+```
+
+The `,` in a single filter expression represents and `AND` operation.
```
jskeet: and => an

core/spec.md
```diff
@@ -4449,6 +4451,134 @@ However, due to the potential for semantics changes of versions with suffix
...
+While not used in this specification of a server, if tooling needs to provide
+a syntax to represent inserting an item into an array at a certain index then
+this MAY be used:
+
+- `[3:]`           # Insert an item at the 4th index (zero-based).
+
+***Examples***
+- `set myarray[3:]=5` might mean "insert `5` at index `3` in `myarray`,
+  pushing current index positions 3 (and higher) further down in the array.
+
```
jskeet, line 4625: Can -1 be used here to effectively mean "add this as a new last element"?

jskeet (commented):
Initial comments before we meet

core/spec.md
duglin: Line 4466 below says: Double-quotes MAY be used as well. 
Do I need to tweak that to make it clearer?

core/spec.md
duglin: done

core/spec.md
duglin: good catch - and the quote was misplaced

core/spec.md
duglin: I just added this:
```
Note that `['*']` is not the same as `.*`. Rather, `['*']` is a reference
to an attribute/key called `*` - assuming that it is a valid attribute/key 
name in that particular situation.
```
cover it?

core/spec.md
duglin: Just added:
```
Use of `-1` as a special index in an array does not extend to other negative
integers.
```
ok?

core/spec.md
duglin: This one is correct, I fixed the other info.age ones to be info.owner

core/spec.md
duglin: I just extended info to include a map of addresses and this example:
```
- Find all schemagroups that have an "info.address" in `CA`:
  `GET /?filter=schemagroups.info.addresses.*.state=CA`
```
does that cover it?

core/spec.md

core/spec.md
duglin: YES, but no colon is needed! I forgot to add that. Good catch.
I reworded [-1] to say "end of the array" so it can be used to "grab the latest" and "insert at end" based on the usage.

duglin: I added this:
```
Use of `-1` with a `:` (e.g. `[-1:]`) is not defined and SHOULD generate an
error ([bad_request](#bad-request)).
```

duglin:
thanks @jskeet for the review. I think we're all set for our chat in 2 hrs

duglin:
For reference, my current plan for the xr CLI is to support these variants for setting xReg metadata (if the user doesn't want to pass in json):
```
--del       attr             # scalars
--del       obj              # del entire obj (maps are same as obj)
--del       array            # delete entire array

--del       obj.key          # del just the one key
--del       array[2]         # del just the index 2, 0 for first
--del       array[-1]        # del last

--set       obj={}           # create empty obj/map, clear it present
--set       array=[]         # create empty array, clear if present

--set       attr=...         # set scalar attribute
--set       obj.key=...      # set key
--set       array[2]=...     # set index 2
--set       array[2:]=...    # insert before index 2 (0 = add to start)
--set       array[-1]=...    # add to end of array
```

core/spec.md
jskeet: I think it *may* be worth adding the quote itself:

> Double-quotes (`"`) MAY be used as well.

That just makes it stand out visually.

We should probably say that the same quote must be used in both places. Alternatively, add it as another option, e.g.

- `["NAME"]` : as above, but using double-quotes around the name. This can be useful if the name contains an apostrophe.

core/spec.md
```diff
@@ -4451,6 +4451,165 @@ However, due to the potential for semantics changes of versions with suffix
...
+                   # the first index of its 'addresses' array, then 'state'
+                   # attr/key of that address
+
+Depending on the situation in which the notation is being used, there are
+certain special values that MAY be used, a described in the following sections.
+
+### Dot-Notation in Filters
+
+When using [filters](#filter-flag) the following special values MAY be used::
+- `.*`             # Match any item in a collection of objects/maps.
```
jskeet: Not sure whether this should be "any item in a collection of objects/maps" or just "any item in an object/map".

(Failed to add this comment before - found it as a draft...)

core/spec.md
jskeet: Right, but *without* the quote, i.e. `[*]` - can I use that to match any attribute in an object/map?

core/spec.md
```diff
@@ -4451,6 +4451,165 @@ However, due to the potential for semantics changes of versions with suffix
...
+
+Note that `['*']` is not the same as `.*`. Rather, `['*']` is a reference
+to an attribute/key called `*` - assuming that it is a valid attribute/key
+name in that particular situation.
+
+Additional details about the `-1` special value:
+- Use of `-1` as a special index in an array does not extend to other negative
+  integers. Those MUST generate an error ([bad_request](#bad_request)).
+- Use of `-1` when updating an index of an array MUST be interpreted as a
+  request to add a new item to the end of the array.
```
jskeet: Ooh, thinking about it, I'm not sure it should.

Suppose an existing array has two elements, `a` and `b`.

Examples as I think of them:

- `array[0]` = 'c'` => `c`, `a`, `b`
- `array[1]` = 'c'` => `a`, `c`, `b`
- `array[2]` = 'c'` => `a`, `b`, `c`

... but elsewhere, `array[-1]` would refer to `b`, the same as `array[1]` does. So maybe `array[-1]` should be invalid when adding.

core/spec.md
jskeet: Yup, that's a good one.

core/spec.md
jskeet: Hmm... I think that could be a bit confusing, for the reasons given above. We can chat about it.

core/spec.md
duglin: ok added another row

core/spec.md
duglin: I like your wording better

core/spec.md
duglin: oh sorry, I just added this:

```
For clarity, use of square brackets (`[]`) without use of either type of
quotes around the value MUST be interpreted as accessing an array. Conversely,
use of quotes MUST be interpreted as accessing an object/map.
```
better?

core/spec.md
duglin: I think you left out the colon for ex 1 and 2. Ex 3 is ok.

As for "-1", that's why I reworded it as "end of the array" and not "last item in the array", and the context matters.

- set array[-1] would mean "set the end of the array to ... - add to the end
- get array[-1] would mean last item

core/spec.md
duglin: Yup -let's chat but i also added a bullet about delete

core/spec.md
```diff
@@ -4451,6 +4451,173 @@ However, due to the potential for semantics changes of versions with suffix
...
+
+- Find all schemagroups that have an `age` of `5`:
+  `GET /?filter=schemagroups.age=5`
+- Find all schemagroups where `owner` is `joe`:
+  `GET /?filter=schemagroups.info.owner=joe`
+- Find all schemagroups that have a labels `env` set to `prod`:
+  `GET /?filter=schemagroups.info.labels.env=prop`
+- Find all schemagroups that have any label with a value of `June`:
+  `GET /?filter=schemagroups.info.labels.*=June`
+- Find all schemagroups that have a "reviewers" value of `Steve`:
```
jskeet, line 4543: This needs fixing :)

duglin: fixed

core/spec.md
```diff
@@ -4469,47 +4469,43 @@ The following notation operators are defined:
...
-When using [filters](#filter-flag) the following special values MAY be used::
-- `.*`             # Match any item in an object/map.
-- `[*]`            # Match any item in an array.
-- `[-1]`           # Match the end of an array.
+When using [filters](#filter-flag) the following special values MAY be used:
+| | |
+| ------ | --- |
+| `.*`   | Match any item in an object/map |
+| `[*]`  | Match any item in an array |
+| `[-1]` | Match the end of an array |
```
jskeet: Perhaps "final item" rather than "end"?

duglin: works for me - thanks!

core/spec.md
```diff
@@ -3968,19 +3968,22 @@ The format of a filter expression is one of:
 ```
 
 Where:
-- `<PATH>` MUST be a dot (`.`) notation traversal of the Registry to the entity
```
duglin, line 3971: @jskeet if you get a chance, this Filter section repeated some of what's now in the "dot notation" section, see if this reads ok to you. thanks!

duglin:
Approved on the 7/2 call - just waiting for one more review before merging

core/spec.md
```diff
@@ -4469,47 +4469,43 @@ The following notation operators are defined:
...
-                   # attr/key of that address
+| Dot Notation      | Description |
+| ----------------- | --- |
+| `.name`           | An attribute (or map key) called `name` |
+| `['my.name']`     | An attribute (or map key) called `my.name` |
+| `[2]`             | The 3rd (zero-based) item in an array |
+| `["birth.date"] ` | Attribute/key called `birth.date` |
+| `["2"]`           | An attribute/key called `2` (a string, not integer)|
+| `employee.name`   | `name` attribute/key in an object/map called `employee` |
+| `stack[3]`        | The 3rd (zero-based) item in an array called `stack` |
```
jskeet: I don't like "3rd (zero-based)" here (or above)... it's the 4th item in any normal description. (If someone asked you how to access the first element of an array, I expect you'd say `array[0]`. And indeed you do so in the example below...)

"Element with index 3 (zero-based)" would be better, IMO.

core/spec.md
```diff
@@ -4569,11 +4565,11 @@ would introduce two potential problems:
...
 
 As a result, when specifying a filter expression that steps between the
-xRegistry hierarchy, the `ID` portion of the path is excluded. Note that this
+xRegistry hierarchy, the `ID` portion of the path is excluded. However, this
 does not apply to stepping through the path of objects/maps/arrays defined
 within an xRegistry entity (Group, Resource, Versions). In those cases use
-of the `*` wildcard MAY be used to indicate "any" items in that set. As
-shown in the "info" examples above.
+of the `*` wildcard would need to be used to indicate "any" items in that set.
+As shown in the "info.addresses" example above.
```
jskeet: I'd either change this to be part of the previous sentence, or change it to "This is demonstrated in" - "As shown [...]" isn't a sentence in its own right.

core/spec.md
```diff
@@ -4598,15 +4594,21 @@ The `,` in a single filter expression represents an `AND` operation.
...
 - Insert an item into the middle of an array: `[INTEGER:]`.
   - E.g. `set myarray[3:]=mary` would be used to insert "mary" at the 4th
     index (zero-based), pushing current index positions 3 (and higher) further
     down in the array.
-    Note that use of `-1` with this `:` variant SHOULD NOT be used as it is
-    ambiguous. Rather, to append to an array `[-1]` SHOULD be used instead.
+  - If the specified integer value does not exist, then an error MUST be
+    generated. Even in the case of `0` for an empty array.
+- Insert at the start of an array: `[^]`.
+  - E.g. `set myarray[^]=mary` would add "mary" to the start of the array, and
```
jskeet: It might be worth using

> insert "mary" at the start of the array

I think that's clearer than "add to the start" and it's consistent with the outer bullet point

core/spec.md
```diff
@@ -4598,15 +4594,21 @@ The `,` in a single filter expression represents an `AND` operation.
...
+    create the array first if it is not yet defined.
+- Append to the end of an array: `[$]`.
+  - E.g. `set myarray[$]=mary` would add "mary" to the end of the array, and
```
jskeet: Less bothered about changing "add" to "append" here, but you could do so

core/model.md
```diff
@@ -634,7 +634,8 @@ statically-defined attributes, not ones that are dynamically added via an
 The `<RESOURCES>` portion of the map key MUST be the plural name of the
 Resource type being referenced.
 
-The `<PATH>` portion of the map key MUST be a dot (`.`) notional traversal to
+The `<PATH>` portion of the map key MUST be a
+[dot (`.`) notional](spec.md#xregistry-dot--notation) traversal to
```
jskeet: notional => notation

jskeet (approved):
A few more comments, but approving on the grounds that they're easy to fix and - I hope - uncontroversial.

core/model.md

core/spec.md
duglin: well, first 3rd should have been 4th :-). But I went with your wording

core/spec.md
duglin: done!

core/spec.md

core/spec.md
duglin: I like "append" here. thanks

duglin:
@jskeet thanks again for the great review!
Merging once CI is done...

Issue #488, opened 2026-07-06
Remove filter error situation


duglin:
Spec says:
```
<PATH> MUST only consist of valid <GROUPS> type names, <RESOURCES> type names or
versions, otherwise an error
([bad_filter](https://github.com/xregistry/spec/blob/main/core/spec.md#bad_filter))
MUST be generated.
```

This is ambiguous since we can't tell if "xxx.yyy" means "xxx" is a group type or an attribute. 

I think we may have written this sentence back when we only had scalar attributes.

Proposal: I suggest we just remove it and rather than an error they'll just get back zero results.

duglin:
Agreed on 7/14 call

PR #489, opened 2026-07-06
Week of Jul 6

### LINKED ISSUE #496: Should ximported Resources be shown in model?
https://github.com/xregistry/spec/blob/main/core/model.md#reuse-of-resource-definitions

Right now the spec is silent on whether the Resources pulled into a Group via the ximportresources mechanism are included in the resulting "model" like normal Resources, or whether just the ximportedresources attribute is included and it's left as an exercise for the client to import it.

The server, as of today, treats this as a client-side job. I wonder if we added this before we had "modelsource"? I'm thinking that "model" should show the ximported Resources in the target Group so that clients do not need to do it themselves - and the ximportresources trigger could remain as "FYI info" if we want, but it becomes more of a modelsource pre-processor type of thing.

I think this would make client's lives easier.

Proposal:
- like $include, treat ximportresources like a modelsource preprocessor directive.
- ximportresources MUST be excluded from "model" - they can check modelsource if they want that info

duglin:
- Remove filter text about flagging bad groups/resources. Treat them like
  "unknown attributes".
- Move [-1] syntax into "other dot notation uses", not part of core spec.
- Mandate that GET /model is always supported - for interop
- s/ancestor/ancestorid/g - for consistency
- add support for xregistry well-known files
- make ximportresources be a modelsource only thing

Fixes: #496 #495 #490 #494 #493 #488

### DIFF
--- a/core/http.md
+++ b/core/http.md
@@ -1013,7 +1013,7 @@ PATCH /capabilities
 
 A server MUST support clients retrieving its full
 [model definition](./model.md#registry-model) via an HTTP `GET` directed to
-the stand-lone `model` entity.
+the stand-alone `model` entity.
 
 The request MUST be of the form:
 
@@ -3047,7 +3047,7 @@ HTTP/1.1 204 No Content
 
 As defined by the [core specification](./spec.md#xregistry-discovery), clients
 MAY query an xRegistry server for the list of additional xRegistry servers
-that might be of interest. If supported, the Registry-specific discovery file,
+that might be of interest. If supported, the Registry-based discovery file,
 for an HTTP server, MUST be available at the root of the Registry (e.g. as a
 sibling API to `/model`) via the `/.xregistry` API:
 
@@ -3065,6 +3065,7 @@ Content-Type: application/json; charset=utf-8
     "URL", *
   }
 }
+```
 
 Note that unlike the [Host-based Discovery](./spec.md#host-based-discovery)
 mechanism, this API includes a dot (`.`) before `xregistry` to avoid any
--- a/core/http.md
+++ b/core/http.md
@@ -59,6 +59,7 @@ model and semantics that apply to all protocols.
     - [`GET /<GROUPS>/<GID>/<RESOURCES>/<RID>/versions/<VID>`](#get-groupsgidresourcesridversionsvid)
     - [`PATCH` and `PUT /<GROUPS>/<GID>/<RESOURCES>/<RID>/versions/<VID>`](#patch-and-put-groupsgidresourcesridversionsvid)
     - [`DELETE /<GROUPS>/<GID>/<RESOURCES>/<RID>/versions/<VID>`](#delete-groupsgidresourcesridversionsvid)
+  - [xRegistry Discovery](#xregistry-discovery)
 - [Request Flags / Query Parameters](#request-flags--query-parameters)
 - [HTTP Header Values](#http-header-values)
 - [Error Processing](#error-processing)
@@ -1010,7 +1011,7 @@ PATCH /capabilities
 
 #### `GET /model`
 
-A server MAY support clients retrieving its full
+A server MUST support clients retrieving its full
 [model definition](./model.md#registry-model) via an HTTP `GET` directed to
 the stand-lone `model` entity.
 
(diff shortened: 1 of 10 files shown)


core/http.md
```diff
@@ -1010,7 +1011,7 @@ PATCH /capabilities
 
 #### `GET /model`
 
-A server MAY support clients retrieving its full
+A server MUST support clients retrieving its full
 [model definition](./model.md#registry-model) via an HTTP `GET` directed to
 the stand-lone `model` entity.
```
```suggestion
the stand-alone `model` entity.
```

core/model.md
```diff
@@ -118,8 +118,8 @@ The overall format of a model definition is as follows:
       "modelversion": "<STRING>", ?      # Version of the group model
       "modelcompatiblewith": "<URI>", ?  # Statement of compatibility
       "attributes": { ... }, ?           # See "attributes" above
-      "ximportresources": [ "<XIDTYPE>", * ], ?   # Include these Resources
-
+      "ximportresources": [ "<XIDTYPE>", * ], ?   # Include these Resources,
+                                         # only available for "modelsource"
```
jskeet:
```suggestion
                                                 # only available for "modelsource"
```

(If we want to align the hashes.)

core/spec.md
```diff
@@ -503,8 +507,8 @@ For easy reference, the JSON serialization of a Registry adheres to this form:
         "modelversion": "<STRING>", ?     # Version of the group model
         "modelcompatiblewith": "<URI>", ? # Statement of compatibility
         "attributes": { ... }, ?          # Group-level attributes/extensions
-        "ximportresources": [ "<XIDTYPE>", * ], ?   # Include these Resources
-
+        "ximportresources": [ "<XIDTYPE>", * ], ?   # Include these Resources,
+                                          # only available for "modelsource"
```
jskeet: Again, maybe align the hash with the line above?

core/spec.md
```diff
@@ -4623,6 +4630,100 @@ are RECOMMENDED to be used for consistency across xRegistry tooling:
...
+This specification defines the following mechanisms by which xRegistry servers
+can be discovered:
+
+### Host-based Discovery
+
+Leveraging the `/.well-known` mechanism, this specification defines an
+xRegistry file that MAY exist at the root of a host's web server:
+
+```yaml
+GET /.well-known/xregistry
```
jskeet: Having just read about `/.xregistry`, I found this a bit confusing, until I read the section below. May not be a problem, but just thought I'd mention it.

core/spec.md
```diff
@@ -4623,6 +4630,100 @@ are RECOMMENDED to be used for consistency across xRegistry tooling:
...
+
+```yaml
+<link rel="alternative" type="application/xregistry+json"
+      title="STRING" href="URL"/>
+```
+
+- Example:
+  - ```yaml
+    <link rel="alternative" type="application/xregistry+json"
+          title="Our xRegistry server" href="https://example.com/xregistry"/>
```
jskeet: Given the HTTP spec, I'd expect this to be https://example.com/.xregistry - note the dot. Again, I may be missing something.

core/http.md
duglin: I find my typing is getting worse as I'm getting older.. thanks

core/model.md

core/spec.md

core/spec.md
duglin: What about it was confusing?

core/spec.md
duglin: I'll fix the URL - I picked as bad sample URL. This uses "example.com/xregistry" as the registry root :-)

core/spec.md
duglin: Would moving the stuff about "host access" up make it less confusing?

jskeet: It was confusing that there were two different paths for what sounded like the same thing. But I may have misunderstood - let's chat in the meeting.

duglin: I redid the section trying to guess what was bugging you - see if it's any better

core/spec.md
```diff
@@ -4623,6 +4630,100 @@ are RECOMMENDED to be used for consistency across xRegistry tooling:
...
+specific host, it is not a requirement that this be the case.
+
+- Examples:
+  - ```
+    {
+      "registries": [
+        "https://example.com/schemaregistry"
+      ]
+    }
+
```
jskeet, line 4676: Nit:  missing closing backticks

jskeet (commented):
Changes in the second commit look fine, but don't quite explain the difference between `.well-known/xregistry` and `.xregistry` - again, this may well just be me though...

clemensv (approved):
Let's merge it

duglin:
Approved on the 7/16 call

PR #500, opened 2026-07-21
Week of July 20

duglin:
- misc typos
- remove "shortself" from doc view
- write href checker tool in golang
- prep for README in ce/samples dir
- add ?filter=excludeall

### DIFF
--- a/core/spec.md
+++ b/core/spec.md
@@ -4097,11 +4097,9 @@ returned. If the URL references an entity, then an error
 appear at the same time, then an error ([bad_filter](#bad_filter)) MUST be
 generated.
 
-
-There is special `filter` value (`excludeall`) that MAY be used to indicate
-that no results are to be returned from the query. This value is expected to
-only be used by server's in response to a query that produces a
-`<COLLECTION>url`
+The special `filter` value `excludeall` MAY be used to indicate that no
+results are to be returned from the query. This value is expected to only be
+used by servers in response to a query that produces a `<COLLECTION>url`.
 
 If the request references an entity (not a collection), and the expression
 references an attribute in that entity (i.e. there is no `<PATH>`), then if the
--- a/core/spec.md
+++ b/core/spec.md
@@ -3,7 +3,7 @@
 <!-- words: validatecompatibility validateformat strictvalidation matchcase -->
 <!-- words: compat formatvalidated compatibilityvalidated -->
 <!-- words: compat formatvalidatedreason compatibilityvalidatedreason -->
-<!-- words: matchversions -->
+<!-- words: matchversions excludeall -->
 <!-- words: myarray myobject -->
 <!-- words: href schemaregistry webpage -->
 
@@ -1119,8 +1119,12 @@ of the existing entity. Then the existing entity would be deleted.
 
 - Type: URL
 - Description: A server-generated unique absolute URL for an entity. This
-  attribute MUST be an alternative URL for the owning entity's `self`
-  attribute. The intention is that `shortself` SHOULD be shorter in length
+  attribute MUST be an alternative URL for the owning entity's `self` URL,
+  without any `$details` suffix. In this respect, the `shortself` URL MAY
+  have `$details` or any other flags (e.g. query parameters) that are valid
+  for use on the `self` URL.
+
+  The intention is that `shortself` SHOULD be shorter in length
   than `self` such that it MAY be used when the length of the URL referencing
   the owning entity is important. For example, in cases where the size of a
   message referencing this entity needs to be as small as possible.
@@ -1149,6 +1153,8 @@ of the existing entity. Then the existing entity would be deleted.
   - MUST be a non-empty absolute URL referencing the same entity as the `self`
     URL, either directly or indirectly via a protocol-specific redirect.
   - MUST be a read-only attribute.
+  - MUST NOT contain the `$` character as that is reserved for specification
+    defined suffixes (such as `$details`).
 
 - Examples:
   - `https://tinyurl.com/xreg123` redirects to
@@ -3845,6 +3851,7 @@ view" when serializing entities and MUST be modified to do the following:
 - MUST remove the default Version attributes from a Resource's serialization.
 - MUST remove the Version `formatvalidated` and `compatibilityvalidated`
   attributes from Version serializations.
+- MUST remove the `shortself` attribute from all entity serializations.
 - When a Resource (source) uses the `xref` feature, the target Resource's
   attributes MUST be excluded from the source's serialization.
 - Resources and Versions MUST be serialized in their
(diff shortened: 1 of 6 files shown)


core/spec.md
```diff
@@ -1119,8 +1119,12 @@ of the existing entity. Then the existing entity would be deleted.
 
 - Type: URL
 - Description: A server-generated unique absolute URL for an entity. This
-  attribute MUST be an alternative URL for the owning entity's `self`
-  attribute. The intention is that `shortself` SHOULD be shorter in length
+  attribute MUST be an alternative URL for the owning entity's `self` URL,
+  without any `$details` suffix. In this respect, the `shortself` URL MAY
+  have `$details` or any other flags (e.g. query parameters) that are valid
+  for use on the `self` URL.
```
jskeet: Okay, so `$details` can't be part of the path, but can be a query parameter, right?

core/spec.md
```diff
@@ -4056,6 +4063,46 @@ following constraints:
...
+  "schemagroupscount": 10
+}
+```
+
+If the URL references a collection then an empty collection (`{}`) MUST be
+returned. If the URL references an entity, then an error
+([not_found](#not_found)) MUST be generated. If any other filter expression
+appear at the same time, then an error ([bad_filter](#bad_filter)) MUST be
+generated.
+
```
jskeet, line 4099: Nit: double blank line

core/spec.md
```diff
@@ -4056,6 +4063,46 @@ following constraints:
...
+As specified in the [Registry Collections](#registry-collections) section,
+when a query includes a filter expression, any nested xRegistry collections
+in the response SHOULD result in the specification of a `<COLLECTION>url`
+attribute that allows clients to easily traverse into that collection with
+the appropriated sub-filter expression to yield the same results as the
+original filter, but just for that nested collection. However, if the filters
+results in zero entities for a particular collection due to the filter
+excluding that collection entirely, rather than the `<COLLECTION>url` not
+having any filter expression at all (which would result in the full collection
+being returned), servers MUST use the special filter expression `excludeall`
```
jskeet, line 4075: I really love the clarity of `excludeall` as a filter. Unambiguous, succinct... which genius came up with that?

;)

core/spec.md
```diff
@@ -4056,6 +4063,46 @@ following constraints:
...
+```
+
+If the URL references a collection then an empty collection (`{}`) MUST be
+returned. If the URL references an entity, then an error
+([not_found](#not_found)) MUST be generated. If any other filter expression
+appear at the same time, then an error ([bad_filter](#bad_filter)) MUST be
+generated.
+
+
+There is special `filter` value (`excludeall`) that MAY be used to indicate
```
```suggestion
The special `filter` value `excludeall` MAY be used to indicate
```

(Just to avoid the crucial information, the value, from being parenthesized...)

Alternatively, "There is a special `filter` value" - i.e. add "a" before "special".

core/spec.md
```diff
@@ -4056,6 +4063,46 @@ following constraints:
...
+that no results are to be returned from the query. This value is expected to
+only be used by server's in response to a query that produces a
```
```suggestion
only be used by servers in response to a query that produces a
```

core/spec.md
```diff
@@ -4056,6 +4063,46 @@ following constraints:
...
+`<COLLECTION>url`
```
jskeet: Incomplete sentence?

tools/verify.go
```diff
@@ -0,0 +1,1218 @@
...
+		{"sadnakskd bad <!--no-verify-specs--> dasdasd", "<!--no-verify-specs-->"},
+	}
+	for _, c := range cases {
+		m := reSkipText.FindString(c.given)
+		checkEqual(fmt.Sprintf("test_skip_text(%q)", c.given), m, c.expected)
+	}
+}
+
+// test_bookmark_pattern_matches_given_patterns
+func testBookmarkPatternMatches() {
```
jskeet, line 939: Possibly move these into a `verify_test.go` file? (It feels odd to me to have the production code and test code in the same file, but I'm not a Go expert by any means.)

tools/verify.go
```diff
@@ -0,0 +1,1218 @@
+// verify.go is a Go port of tools/verify.py + tools/test_verify.py.
```
jskeet, line 1: Can we delete those files in the same PR, or are we not quite ready to do that?

(Btw, I haven't reviewed this code at all...)

core/spec.md
duglin: it can be added yes, but it's not a query parameter.
So:
```
non-shortself: http://example.com/dirs/d1/files/f1
shortself: http://example.com/r/1234
```
and
```
non-shortself: http://example.com/dirs/d1/files/f1$details
shortself: http://example.com/r/1234$details
```

make sense?

core/spec.md
duglin: fixed

core/spec.md

core/spec.md

core/spec.md
duglin: forgot the trailing period. fixed

tools/verify.go
duglin: not yet - I need to make a docker image first and then hook it up for real once I trust it. As of now it's just something I run manually via: make hrefs

tools/verify.go
duglin: Definitely weird and I'll clean it up later, but during my "do I really want this?" phase I like to keep things all in one spot to avoid extra noise.

core/spec.md
jskeet: Okay, I *think* I get it now, but the "without any `$details` suffix" is ambiguous - I read it as being "the `shortself` mustn't have `$details` even if `self` does".

If you were to just remove "without any `$details` suffix" would we be losing information? If not, let's remove it - if so, let's clarify.

duglin: see if the new text is any better

core/spec.md
```diff
@@ -1120,7 +1120,11 @@ of the existing entity. Then the existing entity would be deleted.
 - Type: URL
 - Description: A server-generated unique absolute URL for an entity. This
   attribute MUST be an alternative URL for the owning entity's `self`
-  attribute. The intention is that `shortself` SHOULD be shorter in length
+  (non-`$details` suffixed) URL. When used as part of a request, the
+  `shortself` URL MAY include `$details`, or any flags (e.g. query parameters),
+  that are valid for use on the `self` URL.
```
jskeet: Ah, I think I see now. So the `shortself` *in the response* should never include `$details`, but a client can *add* `$details` to `shortself` and get the same effect as the longer URL.

I think a tweak is still needed, because the `shortself` URL itself *does not* include `$details`... when making a request, you're not saying "this is the `shortself` URL" you're just using a URL which has the value copied from the `shortself` attribute.

How about:

> When a client constructs a request based on the `shortself` URL, it may append `$details` or any flags (e.g. query parameters) that are valid for use on the `self` URL.

?

duglin: Yup - and minor tweak to what you said though:

...but a client can add $details to shortself and get the same effect as the longer URL.
should be:
...but a client can add $details to shortself and get the same effect as the longer URL with $details added.

I assume that's what you meant.

jskeet: Yup - basically "whatever you add to shortself will have the same effect as if you added it to self".

duglin:
Approved on the 7/23 call

PR #504, opened 2026-07-28
Week of July 27

duglin:
- clarify what a dangling xref resource looks like
- expand an error msg to cover other cases


duglin:
Merging early due to it blocking other work.
@jskeet if you get a chance, a review would still be nice and I can fix anything you find in a follow-on PR.

jskeet (commented):
All makes sense to me.

duglin:
@jskeet thanks!

PR #509, opened 2026-07-29
fix(endpoint): align authorization and protocol options guidance

### LINKED ISSUE #505: Endpoint bindings: make multi-role endpoints unambiguous and define protocol-specific subscription options
This issue proposes a role-safe endpoint model and binding updates for HTTP,
AMQP 1.0, MQTT 3.1.1, MQTT 5.0, Kafka, and Core NATS.

The root problem is structural:
[`usage`](https://github.com/xregistry/spec/blob/main/endpoint/spec.md#usage) is an array, but the endpoint has only one flattened/shared set of
`protocoloptions`, `endpoints`, `authorization`, `deployed`, `messagegroups`/`messages`, `envelope`, and `channel` values. Those values cannot express different addresses, credentials policy, deployment state, message contracts, envelope constraints, or option values per role.
Consequently, arbitrary role combinations are syntactically possible but cannot be interpreted reliably.

This proposal does **not** add a generic subscription resource or delivery-policy abstraction. It tightens endpoint role rules and puts
subscription behavior in each protocol binding.

It also does **not** request any new constraint language, conditional-model feature, or other capability in the xRegistry core model. All proposed
requirements are Endpoint-model semantics and should be expressed using the core model capabilities that already exist, supplemented by Endpoint-specific validation where a relationship cannot be represented directly in `endpoint/model.json`.

## Upstream surfaces to change

* `endpoint/spec.md`:
  * **Usage**: client-relative role meanings and allowed combinations;
  * **Protocoloptions**: the actual flattened shape and common-option behavior;
  * **Protocol Options**: HTTP, AMQP 1.0, MQTT 3.1.1, MQTT 5.0, Kafka, and NATS
    applicability and semantics.
* `endpoint/model.json`: express the role cardinality and protocol-option shape that can be represented with existing xRegistry core-model constructs,
  including exact option names, types, ranges, enums, and defaults.
* Endpoint samples and validation: conforming examples for all six modeled bindings and rejected examples for invalid role/option combinations.

No change to the xRegistry core specification, core model schema, or generic constraint capabilities is requested.

The baseline reviewed was Endpoint Registry 1.0-rc3 at
[`f38a3b0f`](https://github.com/xregistry/spec/tree/f38a3b0fcccfe163dc01c6ef81d1d4fa4ce0b4cd/endpoint).

## Normative role semantics

Roles are relative to the client interacting with the advertised endpoint, not to the business identity of the endpoint owner:

| `usage` | Client action | Contract asserted | Not asserted |
|---|---|---|---|
| `producer` | Sends/publishes messages **to** the endpoint (`client -> endpoint`). | The endpoint accepts conforming messages. | That the endpoint owner is the domain producer. |
| `consumer` | Receives/retrieves messages **from** the endpoint (`endpoint -> client`). | The endpoint transfers messages to a consuming client. | Permission or ability to create/change the receive interest. |
| `subscriber` | Configures a consumer endpoint or establishes protocol receive interest. | Subscription-management/setup capability. | **Message transfer. A subscriber alone is not a source, sink, or consumable stream.** |

Proposed normative rule:

1. An endpoint **MUST** have exactly one usage unless its selected protocol binding explicitly permits a combination.
2. Only `subscriber` + `consumer` **MAY** be combined, and only for MQTT, AMQP 1.0, or Core NATS when the same interface/session performs setup and receipt
   and every shared value is unambiguous.
3. HTTP and Kafka subscriber and transfer roles **MUST** use separate endpoint resources.
4. `producer` **MUST NOT** be combined with any other role. Three-role endpoints are invalid.
5. If addresses, authorization, deployment state, messages, envelope, or options differ by role, the declaration **MUST** be split even for an otherwise
   permitted pair.
6. Duplicating connection data in role-specific endpoint resources is allowed and preferred to guessing. Related resources **MAY** repeat `channel`;
   `channel` only correlates them and does not merge contracts.

## Common shape and alignment fixes

`protocoloptions` is a flattened object in `endpoint/model.json` and samples.
Protocol fields are direct children of it, not children of a prose-only
`protocoloptions.options` object. The prose and model should consistently use
that shape.

Common fields remain shared across all declared roles:

* `endpoints`: protocol addresses. Split when roles use different addresses.
* `authorization`: authorization configuration, not credentials. Split when
  roles require different grants/configuration.
* `deployed`: one deployment-state assertion for the whole logical endpoint; it
  is not per role.
* `messagegroups`, inline `messages`, and `envelope`: producer/consumer transfer
  contracts. They do not make `subscriber` a transfer role.
* `channel`: opaque correlation only.

Please also resolve these existing prose/model inconsistencies in the same
change:

* AMQP prose uses unhyphenated names while the model uses `link-properties`,
  `connection-properties`, `connection-capabilities`, `node-capabilities`, and
  `distribution-mode`. Use the exact model names everywhere.
* AMQP `durable` is described as a producer transfer flag in prose and node
  durability in the model. Deprecate it or define one precise meaning; it **MUST
  NOT** imply terminus durability or a delivery guarantee.
* The model defaults `deployed` to `false`, while prose says `true`. Select and
  document one default in both surfaces.
* Kafka model and prose disagree on the `acks` default. Select one normative
  default without silently changing existing documents.
* Document modeled Kafka `headers`, `keyserializer`, and `valueserializer` as
  producer-only fields.
* Extend Kafka `partition` prose beyond producer selection to its consumer
  explicit-assignment meaning described below.

### MQTT 3.1.1 and MQTT 5.0

For both versions:

| Field | Type/values | Usage | Rule |
|---|---|---|---|
| `topic` | string Topic Name | producer | Concrete PUBLISH destination; wildcard topic-filter tokens are invalid. |
| `topicfilter` **(new)** | string Topic Filter | subscriber | SUBSCRIBE filter; normally used on `subscriber` + `consumer`. |
| `qos` | integer `0..2` | directional | Producer: PUBLISH QoS. Subscriber: maximum requested subscription QoS. Consumer: resulting delivery QoS may be lower. A combined declaration is valid only if the scalar is intentionally unambiguous; otherwise split/profile it. |
| `retain` | boolean | producer | PUBLISH retain flag only. |
| `willtopic`, `willmessage` | existing types | connection/session | CONNECT Will configuration; declaring a Will does not add producer usage to ordinary transfers. |

For MQTT 3.1.1, `cleansession` is a boolean CONNECT session-state choice
applicable to the described connection (especially a subscriber/consumer
session, but also a publishing connection). It is not consumer-only or itself a
subscription policy. MQTT 3.1.1 adds no standard `sharedsubscriptiongroup` or
MQTT 5 properties.

For MQTT 5.0 add:

| Field | Type/allowed values | Usage | Dependency/semantics |
|---|---|---|---|
| `sharedsubscriptiongroup` | string | subscriber | Requires `topicfilter`. Serialize the wire filter as `$share/{sharedsubscriptiongroup}/{topicfilter}`; do not store that composite in `topic`. |
| `cleanstart` | boolean | connection/session | MQTT 5 CONNECT Clean Start. |
| `sessionexpiryinterval` | uinteger `0..4294967295` seconds | connection/session | MQTT 5 Session Expiry Interval. |
| `nolocal` | boolean | subscriber | MQTT 5 SUBSCRIBE No Local. |
| `retainaspublished` | boolean | subscriber | MQTT 5 SUBSCRIBE Retain As Published. |
| `retainhandling` | integer enum `0`, `1`, `2` | subscriber | MQTT 5 Retain Handling. |

`cleansession` should be deprecated/rejected for MQTT 5. Clean Start plus
Session Expiry Interval must replace it; the migration text must not claim the
representations are equivalent for every combination.

Do not add `subscriptionidentifier` to endpoint metadata. Although MQTT 5
Subscription Identifiers are selected by the subscribing client rather than
assigned by the server, they identify one concrete SUBSCRIBE operation and may
change independently of the endpoint contract. They belong in runtime
subscription state.

### AMQP 1.0

Keep existing fields, align their hyphenated names, and add the following direct
children of `protocoloptions`:

| Field | Type/allowed values | Applicability and dependency |
|---|---|---|
| `source-filters` | map of protocol-defined filter keys to values | subscriber/source setup, normally with consumer. It preserves AMQP filter descriptors and does not invent a generic selector language. |
| `dynamic` | boolean | Dynamic source for subscriber/consumer setup or dynamic target for producer. Split when source and target values differ. |
| `terminus-durability` | enum `none`, `configuration`, `unsettled-state` | Source for subscriber/consumer; target for producer. |
| `expiry-policy` | enum `link-detach`, `session-end`, `connection-close`, `never` | Lifetime of the applicable source or target terminus. |
| `timeout` | uinteger seconds | Timeout for the applicable terminus expiry policy. |
| `sender-settle-mode` | enum `unsettled`, `settled`, `mixed` | AMQP sender-link settlement, not an xRegistry producer/consumer synonym. On a producer link the client is the sender; on a consumer link the remote endpoint is the sender. |
| `receiver-settle-mode` | enum `first`, `second` | AMQP receiver-link settlement. On a consumer link the client is the receiver; on a producer link the remote endpoint is the receiver. |

Existing-field applicability must also be explicit:

* `node` is a producer target and a consumer/subscriber source. A combined
  endpoint is valid only for the same node.
* `link-properties` and `node-capabilities` are direction-dependent; split if
  sender/receiver or source/target values differ.
* `connection-properties` and `connection-capabilities` are connection-scoped
  and may be shared only when the same connection values apply.
* `distribution-mode` is consumer receive-side and accepts exactly `move` or
  `copy`; it is not subscription management.
* Settlement and distribution controls describe AMQP behavior and **MUST NOT**
  be normalized into generic “at least once” or similar guarantees.

### Kafka

Kafka subscriber and consumer endpoints are intentionally separate:

* A Kafka `subscriber` identifies the broker/topic interface through which a
  client creates a consumer group. It **MUST omit** `consumergroup`.
* A Kafka `consumer` identifies consumption through a particular group. It
  **MUST include** non-empty string `consumergroup`.
* Even when bootstrap servers and topic are identical, these are separate
  endpoint resources. The presence of `consumergroup` is the discriminator.

Field rules:

| Field | Type/allowed values | Usage | Rule |
|---|---|---|---|
| `topic` | string | subscriber, consumer, or producer | Group-creation/consumption source for subscriber/consumer; record destination for producer. |
| `consumergroup` | string | consumer only | Required on consumer; prohibited on subscriber. |
| `partition` | uinteger | producer or consumer | Producer destination selection; consumer explicit partition assignment. |
| `autooffsetreset` **(new)** | enum `earliest`, `latest`, `none` | consumer only | Used only when the consumer group has no valid committed offset. |
| `enableautocommit` **(new)** | boolean | consumer only | Controls client commits; it does not assert a delivery guarantee. |
| `acks`, `key`, `headers`, `keyserializer`, `valueserializer` | existing model types (`acks` values/default aligned in model and prose) | producer only | Producer record/serialization behavior, not subscription configuration. |

The endpoint does not carry a concrete offset. A committed offset is mutable
consumer-group runtime state, while a requested offset is an invocation-level
read position. `autooffsetreset` describes behavior only when the group has no
valid committed offset. `enableautocommit` only controls commit behavior.

Broker topic creation, partition count/replication, ACL administration,
consumer-group inventory, rebalancing/runtime membership, and live lag are
topology/admin/runtime concerns and remain outside endpoint metadata.

### HTTP

No new HTTP field is required:

* `subscriber` is a CloudEvents Subscription Manager base/API endpoint.
* `producer` is an HTTP WebHook/push sink to which the client sends events.
* `consumer` is a message source that the client polls/retrieves from.
* `method` describes one concrete operation: commonly POST for producer, GET for
  consumer, or one manager operation for subscriber. It **SHOULD** be omitted
  for a multi-method manager unless an API profile defines one value.
* `headers` and `query` are request metadata for that concrete operation.
  Different role-specific values require separate endpoint resources.

HTTP management, push, and pull declarations must not be combined.

### Core NATS

| Field | Type | Usage | Rule |
|---|---|---|---|
| `subject` | string | producer | Concrete publish subject; wildcard tokens are invalid. |
| `subjectfilter` **(new)** | string | subscriber | Subscription subject; Core NATS wildcard tokens are allowed. Normally used with consumer. |
| `queuegroup` **(new)** | string | subscriber | Queue-subscription group; requires `subjectfilter`. |

A Core NATS `subscriber` + `consumer` endpoint is permitted only when one
connection establishes interest and receives messages with unambiguous shared
values. JetStream durability, acknowledgements, replay, streams, consumers, and
offsets are not part of this binding.

## Concise conforming endpoint examples

These are complete endpoint objects to place under an xRegistry `endpoints`
collection; proposed fields require the model changes above.

### MQTT 5 shared subscription

```json
{
  "id": "orders-mqtt-receiver",
  "usage": ["subscriber", "consumer"],
  "protocol": "MQTT/5.0",
  "protocoloptions": {
    "endpoints": [{"url": "mqtts://broker.example.com"}],
    "topicfilter": "orders/+/created",
    "sharedsubscriptiongroup": "eu-workers",
    "qos": 1,
    "cleanstart": false,
    "sessionexpiryinterval": 86400,
    "nolocal": true,
    "retainaspublished": false,
    "retainhandling": 0
  }
}
```

### Kafka group creation and group-specific consumption

```json
{
  "id": "orders-kafka-group-manager",
  "usage": ["subscriber"],
  "protocol": "KAFKA",
  "protocoloptions": {
    "endpoints": [{
      "bootstrap.servers": ["kafka-1.example.com:9093"],
      "security.protocol": "SSL"
    }],
    "topic": "orders"
  }
}
```

```json
{
  "id": "orders-kafka-fraud-consumer",
  "usage": ["consumer"],
  "protocol": "KAFKA",
  "protocoloptions": {
    "endpoints": [{
      "bootstrap.servers": ["kafka-1.example.com:9093"],
      "security.protocol": "SSL"
    }],
    "topic": "orders",
    "consumergroup": "fraud-detectors",
    "partition": 3,
    "autooffsetreset": "earliest",
    "enableautocommit": false
  }
}
```

### AMQP filtered source

```json
{
  "id": "orders-amqp-receiver",
  "usage": ["subscriber", "consumer"],
  "protocol": "AMQP/1.0",
  "protocoloptions": {
    "endpoints": [{"url": "amqps://broker.example.com/orders"}],
    "node": "orders",
    "dynamic": false,
    "source-filters": {
      "apache.org:selector-filter:string": "region = 'EU'"
    },
    "terminus-durability": "unsettled-state",
    "expiry-policy": "never",
    "timeout": 0,
    "distribution-mode": "copy",
    "sender-settle-mode": "unsettled",
    "receiver-settle-mode": "first"
  }
}
```

### HTTP Subscription Manager and WebHook

```json
{
  "id": "orders-subscription-manager",
  "usage": ["subscriber"],
  "protocol": "HTTP",
  "protocoloptions": {
    "endpoints": [{"url": "https://events.example.com/subscriptions"}],
    "deployed": true
  }
}
```

```json
{
  "id": "orders-webhook",
  "usage": ["producer"],
  "protocol": "HTTP",
  "protocoloptions": {
    "endpoints": [{"url": "https://app.example.com/hooks/orders"}],
    "method": "POST",
    "headers": [{
      "name": "Content-Type",
      "value": "application/cloudevents+json"
    }],
    "deployed": true
  }
}
```

### Core NATS queue subscription

```json
{
  "id": "orders-nats-receiver",
  "usage": ["subscriber", "consumer"],
  "protocol": "NATS",
  "protocoloptions": {
    "endpoints": [{"url": "tls://nats.example.com:4222"}],
    "subjectfilter": "orders.*.created",
    "queuegroup": "eu-workers"
  }
}
```

## Validation and rejected cases

Validation should be conditional on `protocol`, protocol version, and `usage`;
strict validation should reject options outside the selected branch rather than
ignoring them.

| Rejected declaration | Reason |
|---|---|
| `usage: ["producer", "consumer"]` or `["producer", "subscriber"]` | Producer combinations are prohibited because flattened contracts/options cannot distinguish direction. |
| `usage: ["producer", "consumer", "subscriber"]` | Three-role endpoints are prohibited. |
| HTTP `["subscriber", "consumer"]` | Manager and pull-transfer interfaces must be separate. |
| Kafka `["subscriber", "consumer"]` | Group creation and group-specific consumption must be separate resources. |
| Kafka subscriber with `consumergroup` | Subscriber must omit the group discriminator. |
| Kafka consumer without `consumergroup` | Consumer must identify its group. |
| Kafka subscriber with `autooffsetreset` or `enableautocommit` | Reset and commit policies are consumer-only. |
| MQTT producer `topic: "orders/+/created"` or NATS producer `subject: "orders.*"` | Publish destinations must be concrete; wildcard syntax belongs in `topicfilter`/`subjectfilter`. |
| MQTT 3.1.1 with `cleanstart`, `sessionexpiryinterval`, `sharedsubscriptiongroup`, `nolocal`, `retainaspublished`, or `retainhandling` | These are MQTT 5-only options. |
| MQTT 5 `sharedsubscriptiongroup` without `topicfilter` | Shared-group serialization requires a filter. |
| Producer with `topicfilter`, MQTT SUBSCRIBE options, AMQP `source-filters`, NATS `subjectfilter`, or `queuegroup` | Subscription-only option used on a producer. |
| NATS `queuegroup` without `subjectfilter` | Queue group qualifies a subscription interest. |
| AMQP `distribution-mode` on producer | `move`/`copy` is receive-side. |
| Any option from a different protocol branch, such as Kafka `acks` under MQTT | `protocoloptions` must conform to exactly the selected protocol/version. |
| MQTT/AMQP/NATS subscriber+consumer with different role-specific address, authorization, deployment, message, envelope, or option values | One flattened object cannot represent those two contracts; split it. |

Samples should include both a conforming and rejected corpus for HTTP, AMQP 1.0,
MQTT 3.1.1, MQTT 5.0, Kafka, and Core NATS. In particular, add MQTT 3.1.1
examples using `topicfilter` and `cleansession`, and separate concrete publisher
examples using `topic` for MQTT and `subject` for NATS.

## Migration guidance

* Treat existing `subscriber`-only endpoints as management/setup only. Add or
  reference a `consumer`/`producer` endpoint when transfer is intended.
* Split every endpoint containing `producer`. Split HTTP and Kafka multi-role
  endpoints. Split MQTT, AMQP, and Core NATS pairs unless the new
  subscriber+consumer exception is demonstrably unambiguous.
* Repeating URLs, broker settings, authorization configuration, message
  references, and `channel` in split resources is valid and preferred.
* Move MQTT receive wildcards from `topic` to `topicfilter`. For MQTT 5, split
  `$share/{group}/{filter}` into `sharedsubscriptiongroup` and `topicfilter`.
  Leave concrete publish names in `topic`.
* For MQTT 5, migrate legacy `cleansession` deliberately to `cleanstart` and
  `sessionexpiryinterval`; do not perform a lossy one-boolean substitution.
* Existing Kafka declarations with `consumergroup` become consumer endpoints.
  Group-creation subscriber endpoints omit it; duplicate bootstrap servers/topic
  in the two resources. Do not migrate concrete offsets into endpoint metadata;
  keep committed cursors in runtime group state and requested positions with the
  consuming operation.
* Keep existing AMQP `distribution-mode`. Do not mechanically translate
  ambiguous `durable` to `terminus-durability` or to a delivery guarantee until
  its upstream definition is resolved.
* Existing producer MQTT/NATS values containing wildcard syntax must be
  reclassified as subscriber filters or corrected to concrete destinations.

## Explicitly out of scope

* New xRegistry core-model constraint, conditional-validation, dependency, or
  capability mechanisms. Endpoint-specific rules may be enforced by the Endpoint
  validator when the existing model language cannot encode them.
* A generic subscription resource or subscription XID.
* Generic delivery guarantees, retry/dead-letter policy, or normalized “at
  least/at most/exactly once” claims.
* Broker topology and administration: exchanges, queues, topics, partition
  counts, replication, group inventory, AMQP management, ACL provisioning, or
  cloud event-bus products.
* Credentials, secrets, runtime membership, lag, health, or reachability.
  `authorization` remains configuration and `deployed` remains metadata, not
  live status.
* MQTT Subscription Identifiers, which are client-selected identifiers for
  concrete SUBSCRIBE operations and belong in runtime subscription state.
* NATS JetStream durability, acknowledgements, replay, streams, consumers, and
  offsets. Those require a future, separately identified JetStream protocol
  binding.

### `endpoint/spec.md`

- [ ] Usage semantics are defined client-relatively for producer, consumer, and subscriber.
- [ ] For each protocol option, role applicability is explicitly stated as one or more of producer, consumer, subscriber, or connection-scoped/shared.
- [ ] Protocol value normalization is explicitly defined, including canonical protocol keys and accepted aliases/version spellings that map to each key.
- [ ] The `protocoloptions` object is defined as protocol-branch specific and closed: only fields defined for the selected branch are valid.
- [ ] Unknown fields under `protocoloptions` are rejected for the selected protocol branch.
- [ ] Role-applicability guidance is normative for clients. Role-inapplicable fields are baseline-interpreted per this spec and may be strictly rejected when strict endpoint validation is enabled.
- [ ] Existing prose/model mismatches called out in this issue are resolved, including defaults and naming consistency.

### `endpoint/model.json`

- [ ] The model encodes all constraints representable with existing core capabilities, including `ifvalues`-based branching.
- [ ] No new core-model conditional or expression capability is introduced.
- [ ] New and existing protocol option fields use exact names, types, enums, ranges, and defaults aligned with normative prose.
- [ ] The model remains declarative for field shape and protocol branching; usage-content-dependent enforcement is not required to be expressible in the model.

### Validation behavior

- [ ] Baseline validation mode is supported and validates structural/type/enum/range correctness.
- [ ] Strict endpoint validation mode is optional and, when enabled, rejects declared role/protocol/version dependency violations.
- [ ] Usage-dependent rules that cannot be represented in `endpoint/model.json` are enforced only in endpoint-specific validator logic, not via new core semantics.
- [ ] Validation is conditional on selected protocol branch after protocol normalization.

### Samples and tests

- [ ] Conforming samples cover HTTP, AMQP 1.0, MQTT 3.1.1, MQTT 5.0, Kafka, and Core NATS.
- [ ] Samples include clear role-applicability examples for producer, consumer, and subscriber interpretations.
- [ ] Rejected samples cover protocol-branch violations and unknown `protocoloptions` fields.
- [ ] If strict mode is implemented, rejected samples additionally cover strict role-dependency failures.

## References

* [Endpoint Registry
  specification](https://github.com/xregistry/spec/blob/main/endpoint/spec.md)
* [`usage`](https://github.com/xregistry/spec/blob/main/endpoint/spec.md#usage)
* [`protocoloptions`](https://github.com/xregistry/spec/blob/main/endpoint/spec.md#protocoloptions)
* [Protocol
  Options](https://github.com/xregistry/spec/blob/main/endpoint/spec.md#protocol-options)
* [`endpoint/model.json`](https://github.com/xregistry/spec/blob/main/endpoint/model.json)

clemensv:
## Summary
- reference RFC8414 for OAuth2 authority metadata in endpoint authorization guidance
- add SASL authorization type and mechanism guidance
- add HTTP APIKey/Plain protocol options and align pseudo-schema with model
- add inline spec-level words comment for new endpoint vocabulary

## Validation
- python tools/verify.py endpoint

Fixes #505

### DIFF
--- a/endpoint/spec.md
+++ b/endpoint/spec.md
@@ -530,7 +530,7 @@ This specification defines the following envelope options for the indicated
 ##### `CloudEvents/1.0`
 
 - `mode` : indicates whether the CloudEvent will use `binary` or `structured`
-  (mode)[https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#message].
+  [mode](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#message).
   When specified, its value MUST be one of: `binary` or `structured`, case
   sensitive. When not specified, the endpoint is indicating that either mode
   is acceptable.
@@ -596,9 +596,9 @@ This specification defines the following envelope options for the indicated
   address identifies a network host or links directly to a resource managed by
   the network host is protocol specific.
 
-  The addressing attribute is protocol specific. For `HTTP`, `AMQP/1.0`,
-  `MQTT/3.1.1`, `MQTT/5.0`, and `NATS` the address is carried in a `uri`
-  attribute. For `KAFKA` the address is carried in a `bootstrap.servers`
+  The shape of each entry is customized for the protocol in use. For `HTTP`,
+  `AMQP/1.0`, `MQTT/3.1.1`, `MQTT/5.0`, and `NATS` the address is carried in a
+  `uri` attribute. For `KAFKA` the address is carried in a `bootstrap.servers`
   attribute, which is an array of Kafka bootstrap server addresses, because
   Kafka clients are configured with a bootstrap server list rather than with a
   single destination URI.
@@ -607,10 +607,13 @@ This specification defines the following envelope options for the indicated
   - OPTIONAL.
   - Entries are ordered by preference; the first entry is the preferred
     address.
-  - For all protocols other than `KAFKA`, each object MUST contain a `uri`
-    attribute with a valid, absolute URI (URL).
-  - For `KAFKA`, each object MUST contain a non-empty `bootstrap.servers`
-    attribute. A `uri` attribute is not expected for `KAFKA` endpoints.
+  - Each object MUST carry the endpoint address in the attribute that the
+    endpoint's protocol defines for that purpose. Of the protocols defined in
+    this specification, `HTTP`, `AMQP/1.0`, `MQTT/3.1.1`, `MQTT/5.0`, and
+    `NATS` use a `uri` attribute holding a valid, absolute URI (URL), and
+    `KAFKA` uses a non-empty `bootstrap.servers` attribute. A protocol
+    defined outside of this specification MAY define a different addressing
+    attribute.
 - Examples:
   - `[ {"uri": "https://example.com" } ]`
   - ```
@@ -648,7 +651,7 @@ This specification defines the following envelope options for the indicated
 
 - Constraints:
   - OPTIONAL.
-  - MUST be an array of objects if present. It is not a map.
+  - MUST be an array of objects if present.
   - MUST only be used for authorization configuration.
   - MUST NOT be used for credential configuration.
 
@@ -657,21 +660,23 @@ This specification defines the following envelope options for the indicated
 - Type: String
 - Description: The type of the authorization configuration. The value SHOULD be
   one of the following:
-  - `OAuth2`: OAuth 2.0 authorization is used. The `authorityuri` SHOULD
-    reference authorization server metadata as defined by RFC 8414.
+  - `OAuth2`: [OAuth 2.0][RFC6749] authorization is used. The
+    [`authorityuri`](#protocoloptionsauthorizationauthorityuri) SHOULD
+    reference authorization server metadata as defined by [RFC 8414][RFC8414].
   - `Plain`: The client uses username with a plaintext password for
-    authentication and authorization. For HTTP, see protocol options
-    `plainscheme`, `plainusernamefield`, and `plainpasswordfield`; for other
-    protocols, equivalent transport-specific options MAY be defined as
-    protocol extensions.
-  - `SASL`: The client uses a SASL authentication mechanism. The
-    `mechanism` attribute SHOULD be provided when this type is selected.
-  - `X509Cert`: The client uses client certificate authentication and
-    authorization.
+    authentication and authorization. For HTTP, see the
+    [HTTP options](#http-options) `plainscheme`, `plainusernamefield`, and
+    `plainpasswordfield`; for other protocols, equivalent transport-specific
+    options MAY be defined as protocol extensions.
+  - `SASL`: The client uses a [SASL][RFC4422] authentication mechanism. The
+    [`mechanism`](#protocoloptionsauthorizationmechanism) attribute SHOULD be
+    provided when this type is selected.
+  - `X509Cert`: The client uses [X.509][RFC5280] client certificate
+    authentication and authorization.
   - `APIKey`: The client uses an API key for authentication and
-    authorization. For HTTP, see protocol options `apikeyname` and
-    `apikeyin`; for other protocols, equivalent carrier-specific options MAY
-    be defined as protocol extensions.
+    authorization. For HTTP, see the [HTTP options](#http-options)
+    `apikeyname` and `apikeyin`; for other protocols, equivalent
+    carrier-specific options MAY be defined as protocol extensions.
 
 - Constraints:
   - OPTIONAL.
@@ -692,9 +697,11 @@ This specification defines the following envelope options for the indicated
 ###### `protocoloptions.authorization.mechanism`
 
 - Type: String
-- Description: The SASL mechanism name for `authorization.type = "SASL"`
-  (for example `PLAIN`, `SCRAM-SHA-256`, `SCRAM-SHA-512`, `OAUTHBEARER`,
-  or `EXTERNAL`).
+- Description: The SASL mechanism name for `authorization.type = "SASL"`, for
+  example `PLAIN` ([RFC4616][RFC4616]), `SCRAM-SHA-256` or `SCRAM-SHA-512`
+  ([RFC7677][RFC7677]), `OAUTHBEARER` ([RFC7628][RFC7628]), or `EXTERNAL`
+  ([RFC4422][RFC4422]). Mechanism names are those registered in the
+  [IANA SASL Mechanisms registry][IANA SASL Mechanisms].
 
 - Constraints:
   - OPTIONAL.
@@ -730,10 +737,7 @@ This specification defines the following envelope options for the indicated
 - Constraints:
   - OPTIONAL.
   - If present, MUST be either `true` or `false`, case-sensitive.
-  - When not specified, the default value is MUST be `true`.
-
-Protocol-specific options are direct children of `protocoloptions`.
-There is no nested `protocoloptions.options` object.
+  - When not specified, the default value MUST be `true`.
 
 #### `messagegroups`
 
@@ -817,15 +821,16 @@ constraint for their users.
 
 #### Protocol Options
 
-The following protocol options are direct children of `protocoloptions` for
-the respective protocols. All of these are OPTIONAL.
+For each protocol specified in the following sections there is a table that
+describes the set of protocol-specific options and the roles (Producer,
+Consumer, Subscriber) to which each option applies. All protocol options are
+OPTIONAL.
 
-This specification is descriptive for the applicability of protocol options to
-roles: the role columns guide clients on how to interpret metadata, and a
-validator need not reject every role-inapplicable option. Where a
-rule below is stated with MUST, MUST NOT, or REQUIRED, it is a conformance
-requirement of this specification and an implementation MUST reject metadata
-that violates it.
+The role applicability is descriptive: it guides clients on how to interpret
+metadata, and a validator need not reject every role-inapplicable option.
+Where a rule below is stated with MUST, MUST NOT, or REQUIRED, it is a
+conformance requirement of this specification and an implementation MUST
+reject metadata that violates it.
 
 In each table below, role applicability is shown as:
 - `✓`: The option applies to that role. Clients acting in that role SHOULD
@@ -1075,6 +1080,14 @@ Addressing constraints:
 [Apache Kafka producer]: https://kafka.apache.org/31/javadoc/org/apache/kafka/clients/producer/ProducerRecord.html
 [Apache Kafka consumer]: https://kafka.apache.org/31/javadoc/org/apache/kafka/clients/consumer/ConsumerRecord.html
 [HTTP Message Format]: https://www.rfc-editor.org/rfc/rfc9110#section-6
-[RFC7617]: https://www.rfc-editor.org/rfc/rfc7617
+[IANA SASL Mechanisms]: https://www.iana.org/assignments/sasl-mechanisms/sasl-mechanisms.xhtml
+[RFC4422]: https://www.rfc-editor.org/rfc/rfc4422
+[RFC4616]: https://www.rfc-editor.org/rfc/rfc4616
+[RFC5280]: https://www.rfc-editor.org/rfc/rfc5280
 [RFC6570]: https://www.rfc-editor.org/rfc/rfc6570
+[RFC6749]: https://www.rfc-editor.org/rfc/rfc6749
+[RFC7617]: https://www.rfc-editor.org/rfc/rfc7617
+[RFC7628]: https://www.rfc-editor.org/rfc/rfc7628
+[RFC7677]: https://www.rfc-editor.org/rfc/rfc7677
+[RFC8414]: https://www.rfc-editor.org/rfc/rfc8414
 [rfc3339]: https://tools.ietf.org/html/rfc3339
--- a/endpoint/spec.md
+++ b/endpoint/spec.md
@@ -257,8 +257,8 @@ this form:
         "endpoints": [
           {
             "uri": "<URI>"                        # plus endpoint extensions
-          } *
-        ], ?
+          } *                                     # "KAFKA" uses
+        ], ?                                      # "bootstrap.servers" instead
         "authorization": [
           {
             "type": "<STRING>", ?
@@ -386,9 +386,15 @@ to the xRegistry-defined core
 
   The following values are defined for `usage`
 
-  - `subscriber`: The endpoint offers managing subscriptions for delivery of
-    messages to another endpoint, using the [CloudEvents Subscriptions
-    API][CloudEvents Subscriptions API].
+  - `subscriber`: The endpoint offers management of subscriptions that cause
+    messages to be delivered to some other endpoint. A `subscriber` endpoint
+    is a control-plane endpoint: it does not itself transfer messages.
+    Establishing subscription interest is the only interaction it describes.
+
+    Depending on the protocol, subscription interest is established by the
+    [CloudEvents Subscriptions API][CloudEvents Subscriptions API], an MQTT
+    `SUBSCRIBE` packet, establishment of an AMQP link with the endpoint node
+    as source, creation of a Kafka consumer group, or a NATS subscription.
 
     Some perspectives that might exist on a subscriber endpoint:
     - Application from which messages originate.
@@ -427,13 +433,20 @@ to the xRegistry-defined core
     - "consumer"
     - "producer"
   - MUST be an array of at least one.
-  - SHOULD declare exactly one usage value per endpoint.
-  - Mixed usage declarations SHOULD be used only when the same protocol
-    interface and the same endpoint contract unambiguously serve each declared
-    role.
-  - Declarations that mix `producer` with any other role are strongly
-    discouraged. Prefer separate endpoint declarations, connected with a shared
-    `channel` when correlation is useful.
+  - MUST declare exactly one usage value, except for the single combination
+    permitted below.
+  - MUST NOT combine `producer` with any other value.
+  - MUST NOT declare all three values.
+  - MAY declare `["subscriber", "consumer"]` only when the protocol is
+    `MQTT/3.1.1`, `MQTT/5.0`, `AMQP/1.0`, or `NATS`. In these protocols a
+    single client connection both establishes subscription interest and
+    receives the resulting messages, so the two roles are served by one
+    protocol interface and one endpoint contract.
+  - MUST NOT declare `["subscriber", "consumer"]` for `HTTP` or `KAFKA`
+    endpoints. In those protocols the two roles are distinct interfaces and
+    MUST be modeled as separate Endpoint resources.
+  - Endpoints that jointly describe more than one role SHOULD be declared
+    separately and correlated with a shared `channel` value.
 
 #### `channel`
 
@@ -576,15 +589,28 @@ This specification defines the following envelope options for the indicated
 ##### `protocoloptions.endpoints`
 
 - Type: Array of Objects
-- Description: An array of objects map where each object contains a `uri`
-  attribute with the network address to which clients can communicate with
-  the endpoint. The object MAY contain extension attributes that can be used
-  by clients to determine which URI to use, or to configure access to the
-  specific URI. Whether the URI identifies a network host or links directly to
-  a resource managed by the network host is protocol specific.
+- Description: An array of objects where each object describes one network
+  address at which clients can communicate with the endpoint. The object MAY
+  contain extension attributes that can be used by clients to determine which
+  address to use, or to configure access to the specific address. Whether the
+  address identifies a network host or links directly to a resource managed by
+  the network host is protocol specific.
+
+  The addressing attribute is protocol specific. For `HTTP`, `AMQP/1.0`,
+  `MQTT/3.1.1`, `MQTT/5.0`, and `NATS` the address is carried in a `uri`
+  attribute. For `KAFKA` the address is carried in a `bootstrap.servers`
+  attribute, which is an array of Kafka bootstrap server addresses, because
+  Kafka clients are configured with a bootstrap server list rather than with a
+  single destination URI.
+
 - Constraints:
   - OPTIONAL.
-  - Each object MUST contain a `uri` attribute with a valid, absolute URI (URL).
+  - Entries are ordered by preference; the first entry is the preferred
+    address.
+  - For all protocols other than `KAFKA`, each object MUST contain a `uri`
+    attribute with a valid, absolute URI (URL).
+  - For `KAFKA`, each object MUST contain a non-empty `bootstrap.servers`
+    attribute. A `uri` attribute is not expected for `KAFKA` endpoints.
 - Examples:
   - `[ {"uri": "https://example.com" } ]`
   - ```
@@ -610,17 +636,19 @@ This specification defines the following envelope options for the indicated
 
 ##### `protocoloptions.authorization`
 
-- Type: Map
+- Type: Array of Objects
 - Description: OPTIONAL authorization configuration details of the endpoint.
-  When specified, the authorization configuration MUST provide sufficient
-  metadata for clients to select the authorization mechanism
+  Each entry describes one authorization option that the endpoint accepts;
+  a client selects one of them. When specified, each entry MUST provide
+  sufficient metadata for clients to select the authorization mechanism
   and discover where authorization is obtained. Runtime credentials and
   deployment-specific values are expected to be supplied separately through
-  external configuration. The configuration keys below MUST be used as
-  defined. Additional endpoint-specific configuration keys MAY be added.
+  external configuration. The attribute names below MUST be used as
+  defined. Additional endpoint-specific extension attributes MAY be added.
 
 - Constraints:
   - OPTIONAL.
+  - MUST be an array of objects if present. It is not a map.
   - MUST only be used for authorization configuration.
   - MUST NOT be used for credential configuration.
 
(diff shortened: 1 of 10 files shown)


endpoint/spec.md
```diff
@@ -477,7 +530,7 @@ This specification defines the following envelope options for the indicated
 ##### `CloudEvents/1.0`
 
 - `mode` : indicates whether the CloudEvent will use `binary` or `structured`
-  [mode](https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#message).
+  (mode)[https://github.com/cloudevents/spec/blob/main/cloudevents/spec.md#message].
```
duglin: Isn't this ()/[] backwards?

endpoint/spec.md
```diff
@@ -569,14 +636,19 @@ This specification defines the following envelope options for the indicated
...
+  a client selects one of them. When specified, each entry MUST provide
+  sufficient metadata for clients to select the authorization mechanism
+  and discover where authorization is obtained. Runtime credentials and
+  deployment-specific values are expected to be supplied separately through
+  external configuration. The attribute names below MUST be used as
+  defined. Additional endpoint-specific extension attributes MAY be added.
 
 - Constraints:
   - OPTIONAL.
+  - MUST be an array of objects if present. It is not a map.
```
duglin: I think the 2nd sentence is unnecessary

endpoint/spec.md
```diff
@@ -585,16 +657,49 @@ This specification defines the following envelope options for the indicated
 - Type: String
 - Description: The type of the authorization configuration. The value SHOULD be
   one of the following:
-  - OAuth2: OAuth 2.0 authorization is used.
-  - Plain: The client uses username with a plaintext password for
-    authentication and authorization.
-  - X509Cert: The client uses client certificate authentication and
+  - `OAuth2`: OAuth 2.0 authorization is used. The `authorityuri` SHOULD
+    reference authorization server metadata as defined by RFC 8414.
```
duglin: href the rfc?

endpoint/spec.md
```diff
@@ -585,16 +657,49 @@ This specification defines the following envelope options for the indicated
...
+    `plainscheme`, `plainusernamefield`, and `plainpasswordfield`; for other
+    protocols, equivalent transport-specific options MAY be defined as
+    protocol extensions.
+  - `SASL`: The client uses a SASL authentication mechanism. The
+    `mechanism` attribute SHOULD be provided when this type is selected.
+  - `X509Cert`: The client uses client certificate authentication and
     authorization.
-  - APIKey: The client uses an API key for authentication and authorization.
+  - `APIKey`: The client uses an API key for authentication and
+    authorization. For HTTP, see protocol options `apikeyname` and
```
duglin: href'ing these attributes would be nice

duglin: I won't repeat it, but I think this is true for just about all mentions of attributes. Perhaps your intern can do them all :)

endpoint/spec.md
```diff
@@ -617,48 +722,24 @@ This specification defines the following envelope options for the indicated
...
 - Type: Boolean
 - Description: If `true`, the endpoint metadata represents a public, live
-  endpoint that is available for communication and a strict validator MAY test
-  the liveness of the endpoint.
+  endpoint that is expected to be available for communication.
 - Constraints:
   - OPTIONAL.
   - If present, MUST be either `true` or `false`, case-sensitive.
-  - When not specified, the default value MUST be `true`.
+  - When not specified, the default value is MUST be `true`.
```
duglin: I think it reads better w/o the "is"

endpoint/spec.md
```diff
@@ -617,48 +722,24 @@ This specification defines the following envelope options for the indicated
...
-- Description: Additional configuration options for the endpoint. The
-  configuration options are protocol specific and described in the
-  [protocol options](#protocol-options) section below.
-- Constraints:
-  - OPTIONAL.
-  - When specified, MUST be a map of non-empty strings to `ANY` type values.
-  - If `protocoloptions.protocol` is a well-known protocol, the options MUST be
-    compliant with the [protocol's options](#protocol-options).
+Protocol-specific options are direct children of `protocoloptions`.
+There is no nested `protocoloptions.options` object.
```
duglin: These are odd things to say. They're more like historical info. We don't need to worry about what old versions of the spec said.

endpoint/spec.md
```diff
@@ -737,232 +817,248 @@ constraint for their users.
 
 #### Protocol Options
 
-The following protocol options (`protocoloptions.options`) are defined for the
-respective protocols. All of these are OPTIONAL.
+The following protocol options are direct children of `protocoloptions` for
```
duglin: I'd remove the first sentence

endpoint/spec.md
```diff
@@ -528,39 +581,53 @@ This specification defines the following envelope options for the indicated
...
-  specific URL. Whether the URL identifies a network host or links directly to
-  a resource managed by the network host is protocol specific.
+- Description: An array of objects where each object describes one network
+  address at which clients can communicate with the endpoint. The object MAY
+  contain extension attributes that can be used by clients to determine which
+  address to use, or to configure access to the specific address. Whether the
+  address identifies a network host or links directly to a resource managed by
+  the network host is protocol specific.
+
+  The addressing attribute is protocol specific. For `HTTP`, `AMQP/1.0`,
```
duglin: Just guessing but for a non-English reader they might try to find an "addressing" attribute. I wonder if it wouldn't be better to say something like: "The `endpoints` attribute will be customized for each protocol used." Then your examples are good. But, I would also show the `bootstrap.servers` example in the pseudo json we have so people can see the exact shape of it. Similar to how we show the various protocoloptions for each protocol.

endpoint/spec.md
```diff
@@ -528,39 +581,53 @@ This specification defines the following envelope options for the indicated
...
+  attribute, which is an array of Kafka bootstrap server addresses, because
+  Kafka clients are configured with a bootstrap server list rather than with a
+  single destination URI.
+
 - Constraints:
   - OPTIONAL.
-  - Each object MUST contain a `url` attribute with a valid, absolute URL.
+  - Entries are ordered by preference; the first entry is the preferred
+    address.
+  - For all protocols other than `KAFKA`, each object MUST contain a `uri`
```
duglin: I think this needs to be loosened a bit to allow for more than just KAFKA to exclude a `uri` since another one can come along later. Even to just say "Of the ones specified in this specification, all...." would be ok too.

endpoint/spec.md
```diff
@@ -528,39 +581,53 @@ This specification defines the following envelope options for the indicated
...
+    attribute with a valid, absolute URI (URL).
+  - For `KAFKA`, each object MUST contain a non-empty `bootstrap.servers`
```
duglin: Just to repeat,let's show it in the pseudo json.

endpoint/spec.md
```diff
@@ -737,232 +817,248 @@ constraint for their users.
 
 #### Protocol Options
 
-The following protocol options (`protocoloptions.options`) are defined for the
-respective protocols. All of these are OPTIONAL.
+The following protocol options are direct children of `protocoloptions` for
+the respective protocols. All of these are OPTIONAL.
+
+This specification is descriptive for the applicability of protocol options to
+roles: the role columns guide clients on how to interpret metadata, and a
```
duglin: The phrase "role columns" is out of place until they scroll down and see the tables. I think it would flow better if it said something like:  For each protocol specified in the following sections, there will be a table that describes the set of protocol-specific options and for which "roles" (Producer, Consumer, Sender) each is applicable.

endpoint/spec.md
```diff
@@ -737,232 +817,248 @@ constraint for their users.
...
+validator need not reject every role-inapplicable option. Where a
+rule below is stated with MUST, MUST NOT, or REQUIRED, it is a conformance
+requirement of this specification and an implementation MUST reject metadata
```
duglin: "implementation" of what? xReg server? send? consumer? producer?

endpoint/spec.md
```diff
@@ -737,232 +817,248 @@ constraint for their users.
...
-      "operation": "send"
-    }
-  }
-}
-```
+The [endpoint URIs](#protocoloptionsendpoints) for "HTTP" endpoints MUST be
+valid HTTP URIs using the "http" or "https" scheme as defined in
+[HTTP Message Format].
+
+HTTP has no single connection that serves several roles, so each HTTP role is
```
duglin, line 865: This is an interesting statement. Is it not possible for one HTTP endpoint to support more than one role and distinguish which role it needs to act in based on the incoming message? E.g. an endpoint can look for a "subscribe" message vs a "pull" message. I would think it would be an impl choice/detail - as long as the message types are unambiguously differentiated. No?

clemensv (commented):
> Isn't this ()/[] backwards? 

yes. bad robot.

clemensv (commented):
>  I think the 2nd sentence is unnecessary  

agree

endpoint/spec.md
clemensv: yes. bad robot

endpoint/spec.md
clemensv: agree

endpoint/spec.md
clemensv: OpenAPI lets you freely define whatever your interaction pattern for HTTP is. 

Here, an endpoint definition effectively creates a usage profile over an HTTP resource for message passing, but we are not specific. You pass events to the resource, you retrieve events from the resource, or you establish a subscription on the resource to do the latter.

Practically, you will have different paths on the resource (a "message collection") for these actions. For instance /path to send, /path/messages to retrieve and /path/subscriptions to subscribe and those are different endpoints.

"message types" are endpoints/paths in HTTP since it has no other disambiguator. Custom methods are rightfully frowned upon.

endpoint/spec.md
clemensv: The client/reader, e.g. a code generator that acts on this. The server can't know all the constraint relationships because we can't express them well in the model at this time (as discussed).

duglin:
Approved on the 7/30 call

PR #510, opened 2026-08-01
feat(openusd): add OpenUSD Artifact Registry domain specification

marcschier:
## What this adds

A domain specification for an **OpenUSD Artifact Registry** — `workingdrafts/models/openusd/spec.md`, `workingdrafts/models/openusd/model.json` and `workingdrafts/models/openusd/README.md` — plus its row in the README artifacts table.

[OpenUSD](https://openusd.org) composes a scene from many artifacts: a root layer plus the sublayers, references, payloads, textures, MaterialX documents and volumes reachable from it, and the codeless schema plugins that let a consumer resolve a vendor's prim types instead of degrading them to untyped prims. Today every publisher answers *"where do I get these, which revision, and did I get all of them?"* differently — a shared filesystem, an object store, a vendor asset service. Those answers do not interoperate, and none of them makes the closure checkable.

This answers them with the core model rather than with new machinery:

| USD concept | xRegistry |
|---|---|
| asset container (one composable unit) | Group — `usdassetgroups` |
| artifact (layer, package, texture, MaterialX, volume) | Resource — `usdassets`, document is the artifact |
| codeless schema plugin | Group — `usdschemaplugingroups`, holding its manifest and generated schema |
| composition closure | the Group's membership, so a Consumer holding the Group can compose with no further asset resolution |

Versions, `digest` integrity, `dependson` edges and cross-registry federation all come from the core model, so a resolver gets them without this specification inventing anything.

## The part worth reviewing closely: identifiers

A USD asset identifier is the string a layer authors between `@...@`. An xRegistry `<SINGULAR>id` admits only RFC 3986 *unreserved* characters plus `:` and `@`. These are **not the same grammar** — `@./pump.usda@` is not a valid id, and `textures/albedo.png` cannot appear in one at all.

The specification does not equate them. It derives one from the other by a deterministic, closed-form construction (§5.1.1) and keeps the authored identifier as the authority in a REQUIRED `assetidentifier` attribute:

| Authored | `assetidentifier` | `usdassetid` |
|---|---|---|
| `@./pump.usda@` | `pump.usda` | `pump.usda` |
| `@textures/albedo.png@` | `textures/albedo.png` | `textures.albedo.png` |
| `@pkg.usdz[tex/a.png]@` | `pkg.usdz[tex/a.png]` | `pkg.usdz-tex.a.png` |

The forward direction is closed-form, so a resolver locates an artifact by **computation rather than by search** — which is what makes the registry usable as an `ArResolver` backend. The reverse direction is a read of `assetidentifier`, never an inversion, because the construction is deliberately lossy and disambiguated by a hash **of the identifier, not of the document**. That last point is deliberate: a Resource is the umbrella over its Versions, so an id computed from bytes would fork the artifact on every revision.

## Two things I'd like a maintainer's call on

**1. Generated schemas are omitted, and I think that's a finding rather than an oversight.**

Every other domain with a `model.json` ships `schemas/`, so I generated them — and `document-schema.avsc` is **invalid Avro**:

```
Could not make an Avro Schema object from
io.xregistry.usdschemaplugingroups.UsdassetType
```

The generator emits a forward reference to a record it never defines. The cause is that this specification deliberately uses **one resource collection (`usdassets`) in two group types** — §4.4 states that a plugin manifest and a generated schema are USD artifacts like any other, retrieved, versioned and verified identically, differing only in `assetkind` and `format`. `xr model verify` accepts the model; only the Avro projection breaks.

Rather than redesign the specification around a tooling limit, or ship a broken artifact, I left `workingdrafts/models/openusd/schemas/` out. Happy to do whichever you prefer: rename the second collection, or fix the generator to define a shared resource type once. If it's the latter I'm glad to send that as a separate PR.

**2. `digestalg` has no model `default`.** The core model requires an attribute with a `default` to be `required: true`, but `digestalg` is only meaningful when `digest` is present. It is therefore declared without a default and the fallback to `Sha256` is stated as a processing rule in §4.2. Please tell me if you'd rather see it modelled differently.

## Review feedback addressed

**Directory structure.** The specification now lives under `workingdrafts/models/openusd/`, alongside a working-drafts index. The main README links to that index rather than listing individual proposals, so future drafts do not require top-level README updates.

`tools/Makefile` needed three glob updates, because its targets match a single path segment and `models/openusd/` is now two deep. Without them the specification would have silently dropped out of verification while CI stayed green:

| target | change |
|---|---|
| `spellcheck` | also takes `workingdrafts/models/*/README.md` and `workingdrafts/models/*/spec.md` |
| `samplescheck` | also takes `workingdrafts/models/*/model.json` |
| `models` | `xr model verify` also takes `workingdrafts/models/*/model.json` |

`tabcheck`, `badhrefs` and the link checkers walk the tree and were unaffected.

**Version identifier.** The `1.0-rc3` in the document title was raised and left as is, per the discussion on that thread — it tracks the core specification revision this is written against.

## Verification

`make -f tools/Makefile verify` passes in full — spellcheck, tabcheck, ticks, badhrefs, errorcheck, samplescheck, hrefs, 52 pytest tests, and links (**"Spec verification succeeded"**). `xr model verify */model.json workingdrafts/models/*/model.json` reports `models/openusd/model.json: Model verified`.

New vocabulary went into the document's own `<!-- words: ... -->` list rather than `tools/dict`, since it is all domain-local. Four lowercase RFC 2119 keywords the link checker flagged were **reworded rather than capitalized** — each was descriptive prose ("a registry need not host every artifact"), and capitalizing them would have turned description into normative text.

## Context

An OPC UA projection of this same model exists and is being standardized separately; §1.2 notes it as informative only, and this specification neither requires nor depends on it. A conformant instance document and an independent validator (which re-derives every identifier from `assetidentifier` rather than trusting the emitter) live in the drafts repository this was authored in: <https://github.com/marcschier/opcua-drafts>.


models/openusd/spec.md
```diff
@@ -0,0 +1,854 @@
+# OpenUSD Artifact Registry Service - Version 1.0-rc3
```
marcschier, line 1: change to rc1 instead of rc3

duglin:
1 - if I'm following it sounds like it's an avro tooling bug not a model (core spec) issue, so I think we should see if we can fix the tool.
2 - "...but `digestalg` is only meaningful when `digest` is present." - that's an interesting situation. It almost sounds like you want `digest` to have an `ifvalue` but not for a specific value, more of an "if present" check (or an `ifvaue` of `*`). Do you think that would solve your issue?

models/openusd/spec.md
duglin: why?

clemensv:
@duglin In terms of timeline, we should create an extensions folder as we do with CloudEvents and, in that, create a bindings folder (for #511) and a models folder (for this one), so that we can move v1 forward and have the in-development extensions run on a different timeline. 

We should also discuss how we distinguish between very general models that have broad scope (message, schema, endpoint) and these more narrow models that are for a specific domain, in terms of where they go in the repo. 

@marcschier you should definitely join the xReg subgroup calls.

models/openusd/spec.md
marcschier: It was arbitrarily chosen based on pre-review, but we can keep as is.

marcschier:
> 1 - if I'm following it sounds like it's an avro tooling bug not a model (core spec) issue, so I think we should see if we can fix the tool. 2 - "...but `digestalg` is only meaningful when `digest` is present." - that's an interesting situation. It almost sounds like you want `digest` to have an `ifvalue` but not for a specific value, more of an "if present" check (or an `ifvaue` of `*`). Do you think that would solve your issue?

On 1, happy to help.  on 2. yes that would solve it, but up for discussion.  IMO it would be better than to make up something like "none" but if so, this would equally need to be spec'ed in core.

Agree, I see the opc ua binding spec absolutely as extension like in CE,   For openusd, I would like it at the same level as other domain instantiations, as the OpenUSD artifact registry is not something esoterical IMO.  But for me the highest prio is to announce a draft for discussion at this point, not so worried about structure/prio against other domain registries.

duglin:
@marcschier can you update the PR to create dir structure as:
- /models/openusd/model.json
- /models/openusd/spec.md
- /models/openusd/README.md

models/openusd/spec.md
marcschier: Closing this out: no change made — `1.0-rc3` stays.

It tracks the core specification revision this document is written against, and the version line moved with the file to `models/openusd/spec.md` untouched. I've recorded the decision in the PR description so it isn't lost when the thread collapses. Happy to reopen if you'd rather domain specs carry their own independent version track.

marcschier:
> can you update the PR to create dir structure as:
> - /models/openusd/model.json
> - /models/openusd/spec.md
> - /models/openusd/README.md

Done — pushed in ddf8275. All three files are at those paths.

`model.json` and `spec.md` moved unchanged (git records both as 100% renames). The only relative link the spec carries is to `model.json`, which travels with it; everything else is absolute. The `README.md` is new and follows the shape of `core/README.md` and `pagination/README.md`. The root README artifact table now points at `models/openusd/spec.md`.

One thing the move surfaced, worth flagging since it will hit every future `models/*` addition: `tools/Makefile` matches a single path segment, so `models/openusd/` at two levels deep falls out of three targets. A bare `git mv` would have quietly dropped this spec from verification while CI stayed green.

| target | change |
|---|---|
| `spellcheck` | also takes `models/*/README.md` and `models/*/spec.md` |
| `samplescheck` | also takes `models/*/model.json` |
| `models` | `xr model verify` also takes `models/*/model.json` |

`tabcheck`, `badhrefs` and the link checkers walk the tree and were unaffected. Verified against the real tooling rather than assumed — `xr` now reports `models/openusd/model.json: Model verified`, spellcheck reads both new files, `tools/verify.py` reports "Spec verification succeeded", and the 52 pytest tests pass.

marcschier:
As per discussion, we will move this to /workingdrafts/models/openusd/... (@duglin, @clemensv FYI)

marcschier:
Moved this into `workingdrafts/` as agreed in 1a6a136 and c3f5e7f:

- `models/openusd/{README.md,model.json,spec.md}` → `workingdrafts/models/openusd/...`
- added `workingdrafts/README.md`
- added a **Working Drafts** section to the root artifact table and repointed the OpenUSD link
- repointed the `spellcheck`, `samplescheck`, and both `xr model verify` globs so the draft remains fully covered

`make -f tools/Makefile verify` passes, including 52 pytest tests and the link checker. The `xr` model validator also reports `workingdrafts/models/openusd/model.json: Model verified`.

Heads-up: #511 adds its binding glob on the same `spellcheck` command. Whichever PR merges second may need a trivial conflict resolution that keeps both continuation lines.

workingdrafts/README.md
```diff
@@ -0,0 +1,5 @@
+# Working Drafts
+
+The specifications in this directory are under active development and can
+change independently of the xRegistry v1 specification. They are included for
```
duglin: let's remove the "v1" reference

README.md
```diff
@@ -40,6 +40,9 @@ with a lowercase `x`, even when the first word in a sentence.
 | Message Definitions Registry  | [v1.0-rc3](https://github.com/xregistry/spec/blob/v1.0-rc3/message/spec.md) | [WIP](message/spec.md)                         |
 | Schema Registry               | [v1.0-rc3](https://github.com/xregistry/spec/blob/v1.0-rc3/schema/spec.md) | [WIP](schema/spec.md)                         |
 |                               |
+| **Working Drafts:**           |
+| OpenUSD Artifact Registry     | n/a | [WIP](workingdrafts/models/openusd/spec.md)           |
```
duglin: Rather than having to update the main README each time, let's just say something here like: "Upcoming specs", or "Proposed new specs". And then point to the workingdrafts dir or README in the dir

workingdrafts/README.md
marcschier: Addressed in commit `16835f2`: removed the version-specific reference and now describe drafts as changing independently of released xRegistry specifications.

README.md
marcschier: Addressed in commit `16835f2`: replaced the individual OpenUSD row with a generic Proposed New Specifications section linking to `workingdrafts/README.md`, so future drafts do not require updates to the main README.

workingdrafts/README.md
marcschier: Correction: the final DCO-signed fix commit is `a5222ec`, which supersedes `16835f2` after the required sign-off was added.

README.md
marcschier: Correction: the final DCO-signed fix commit is `a5222ec`, which supersedes `16835f2` after the required sign-off was added.

PR #511, opened 2026-08-01
feat(bindings): add an OPC UA protocol binding for the xRegistry API

marcschier:
## What this adds

`workingdrafts/bindings/opcua.md` — an **OPC UA protocol binding** for the xRegistry API, as a peer of [`core/http.md`](https://github.com/xregistry/spec/blob/main/core/http.md) rather than a tunnel over it.

## Why a second binding, and why this one

OPC UA is the interoperability layer of industrial automation, and it already has the two things a registry binding needs:

- **Its AddressSpace is a browsable graph of typed nodes**, so a registry *is* a folder of files rather than a resource tree behind a URL space. Listing a collection is Browse; there is no path grammar to parse.
- **FileTransfer already moves document bytes** (`Open`/`Read`/`Write`/`Close`), so a resource document needs no new transfer mechanism, no base64 in JSON, and no size ceiling from a string limit.

The practical consequence: a registry served this way is reachable from a controller, a gateway or a historian that has an OPC UA stack and no HTTP client — which is most of the installed base this is aimed at.

## It is a complete mapping, not a subset

Every xRegistry operation has an entry:

| xRegistry | OPC UA |
|---|---|
| list a collection | Browse the folder (BrowseNext for continuations) |
| read/write attributes | Read / Write Properties |
| read/replace a document | `Open`/`Read`/`Write`/`Close` on the resource file |
| create | `CreateGroup` / `GetOrCreateGroup` / `CreateResource` / `GetOrCreateResource` |
| delete | `Delete(ExpectedEpoch)` on the entity itself |
| labels and extensions | `Labels.AddAttribute` / `Labels.RemoveAttribute` |

Request flags, value encoding, serialisation, federation and error handling each have a clause. **Annex A** is an operation-by-operation correspondence to the HTTP binding, so a reader who knows that one can read this by lookup.

## Two design points worth review

**Identity is never carried by an endpoint.** No URL authority is involved: identity is the xRegistry identifier attributes and `xid`, while the OPC UA session, endpoint and NodeIds say only *where those entities are currently served*. A federated resource therefore keeps one identity across the endpoints that serve it, and a local proxy for a remote resource MUST retain the remote identity rather than adopting the local one (§9).

**Optimistic concurrency is uniform.** Every mutating method takes an `ExpectedEpoch` argument that can be omitted, so a client gets compare-and-swap semantics without a locking service. §4.5.1 mentions coarser locking only as something a server MAY additionally expose.

## Dependency I want to be explicit about

The OPC UA **information model** this binding operates on — the node types `RegistryType` / `GroupType` / `ResourceType` / `AttributesType` — is a companion specification currently in OPC Foundation standardization. It is cited as a normative reference with an absolute link, the same way this document cites the OPC UA service specifications (OPC 10000-3/-4/-5/-20).

I would rather state that plainly than claim self-containment the document cannot have: like the HTTP binding, this defines *how a protocol realizes the registry model*, not the model itself. If you would prefer the binding to carry its own model definition inline, say so and I will restructure.

## Review feedback addressed

Per the review feedback on this PR, the binding now lives at **`workingdrafts/bindings/opcua.md`** under `workingdrafts/bindings/`, where proposed protocol bindings can be reviewed before release. The move is path-only; every link the document carries is absolute, and no other file referenced the old path.

One consequence needed handling: `tools/Makefile` selects files by directory, not by name, so the `spellcheck` and `ticks` targets — both of which enumerate `core/*.md` — would have silently stopped covering the binding. Both now also enumerate `workingdrafts/bindings/*.md`, which keeps the checks that were passing on `core/opcua.md` running against its new home and extends them to any binding added later. `tabcheck`, `badhrefs` and the link checkers walk the tree and were unaffected.

## Verification

`make -f tools/Makefile verify` passes in full — spellcheck, tabcheck, ticks, badhrefs, errorcheck, samplescheck, hrefs, 52 pytest tests, links (**"Spec verification succeeded"**).

Notes on how it was brought to house style, since the source was written in OPC UA conventions:

- **131 normative keywords converted** to `MUST` / `MUST NOT` / `SHOULD` / `SHOULD NOT` / `MAY`. Each was read rather than substituted: one `may` was descriptive rationale and became "can", and **every** lowercase `optional` / `required` the checker flagged was reworded rather than capitalized. Most of those referred to the OPC UA *Optional ModellingRule* — the same word, a different concept — and capitalizing them would have asserted RFC 2119 requirements the document does not make.
- Prose re-wrapped at 80 columns. The lines still over are ToC entries, link definitions and tables, which cannot be wrapped without changing what renders.
- New vocabulary went into the document's own `<!-- words: ... -->` list rather than `tools/dict`, since it is all OPC UA identifiers (StatusCode, Method and Property names).
- I verified mechanically that the conversion changed **only** what it was meant to: comparing the normalized word stream against the source gives 0.9895 similarity, and every difference is a heading dot, the conventions paragraph, the `can` reword, an `optional` removal or a keyword case change. No content was lost.

`workingdrafts/bindings/opcua.md` is not added to the README artifacts table, since `core/http.md` and `core/events.md` are not listed there either. Happy to add a row if bindings should be indexed.

## Context

Authored in <https://github.com/marcschier/opcua-drafts>, where the companion information model and a worked domain registry live. Related: #510 proposes an OpenUSD Artifact Registry domain specification.


duglin:
Sorry - other comments (now deleted) were wrong - it didn't register that this is a protocol binding and not a new model. We may need a "bindings" dir though. We'll discuss

duglin:
@marcschier can you update the filename/path to be: /bindings/opcua.md

marcschier:
> can you update the filename/path to be: /bindings/opcua.md

Done — pushed in 3279ee8.

The document is now at `bindings/opcua.md`, which creates the `bindings/` directory as the home for protocol bindings. Git records it as a 100% rename: the content is byte-identical, every link the document carries is absolute so nothing needed rewriting, and no other file in the repo referenced the old path.

marcschier:
As per discussion, we will move this to /workingdrafts/bindings/opcua.md. (@duglin, @clemensv FYI)

marcschier:
Moved this into `workingdrafts/` as agreed in bc96bda:

- `bindings/opcua.md` → `workingdrafts/bindings/opcua.md`
- added the same `workingdrafts/README.md` used by #510
- repointed the `spellcheck` and `ticks` globs so the binding remains fully covered

The companion-model URL used by the binding had started returning 404 after that source moved into OPC Foundation review. I pinned its nine references to the last public revision (`ff22f224`) so the cited sections remain accessible. `make -f tools/Makefile verify` now passes, including 52 pytest tests and the link checker.

I left the root artifact table unchanged, consistent with the HTTP and Events bindings not being listed there. I can add a binding row under **Working Drafts** if maintainers want one.

Heads-up: #510 adds its model globs on the same `spellcheck` command. Whichever PR merges second may need a trivial conflict resolution that keeps both continuation lines.

workingdrafts/README.md
```diff
@@ -0,0 +1,5 @@
+# Working Drafts
+
+The specifications in this directory are under active development and can
+change independently of the xRegistry v1 specification. They are included for
```
duglin: let's remove the "v1" reference

marcschier: Addressed in commit `612f533`: removed the `v1` reference and now describe working drafts as changing independently of released xRegistry specifications.

Issue #512, opened 2026-08-02
core/model.schema.json rejects the valid schema/model.json: missing resource validation keys and wrong constraints shape


clemensv:
`core/model.schema.json` rejects `schema/model.json`, which is a valid
spec-defined model. It is the only model in the repository that fails
`tools/validate-models.py`:

```
FAIL: schema\model.json
  ... is not valid under any of the given schemas (at ['groups', 'schemagroups'])
```

There are two independent defects, both in `core/model.schema.json`.

### 1. `ResourceDefinition` is missing three spec-defined keys

`validateformat`, `validatecompatibility` and `strictvalidation` are defined in
[`core/model.md`](../blob/main/core/model.md) — see
`groups.<STRING>.resources.<STRING>.validateformat` and the two sections
following it — and they appear in the pseudo-JSON in both `core/model.md`
(lines 146-148) and `core/spec.md` (lines 527-529):

```jsonc
"validateformat": <BOOLEAN>, ?        # Do Version format checks. Default=false
"validatecompatibility": <BOOLEAN>, ? # Do Version compat checks. Default=false
"strictvalidation": <BOOLEAN>, ?      # Block unknown format/compat. Default=false
```

None of the three appear in `ResourceDefinition` in `core/model.schema.json`,
which sets `"additionalProperties": false`. Any model using them is therefore
rejected. `schema/model.json` uses all three.

### 2. `GroupDefinition.constraints` is modelled one level too shallow

`core/model.md` defines `constraints` as a *map* keyed by
`<RESOURCES>.<PATH>`:

```jsonc
"constraints": {
  "<RESOURCES>.<PATH>": {          # Resource-plural + attribute path
    "default": <VALUE>, ?          # Group specific default
    "enum": [ <VALUE> * ], ?       # Allowed subset of values
    "equals": "<PATH>" ?           # Matching Group attribute path
  } *
}, ?
```

`core/model.schema.json` instead declares `default`, `enum` and `equals` as
direct properties of `constraints`, omitting the keyed level entirely. So the
spec-conformant

```json
"constraints": { "schemas.format": { "equals": "format" } }
```

used by `schema/model.json` does not validate.

### Suggested fix

Add the three boolean keys to `ResourceDefinition`, and change
`GroupDefinition.constraints` to use `patternProperties` keyed on
`<RESOURCES>.<PATH>` with the existing `default`/`enum`/`equals` object as the
value schema.

### Note on discoverability

[...]

PR #514, opened 2026-08-03
Fix invalid JSON in core/model.schema.json (two missing commas)

clemensv:
## Problem

`core/model.schema.json` is not valid JSON. Two object members in the
`GroupDefinition.constraints` block are not separated by commas:

- between the `properties` object and its sibling `description`
- between the `constraints` object and its sibling `resources`

```
json.decoder.JSONDecodeError: Expecting ',' delimiter: line 103 column 15
```

## Impact

`tools/validate-models.py` loads this schema before it validates any model, so
the syntax error makes it abort with an unhandled `JSONDecodeError`. The effect
is not cosmetic: **no model in the repository is validated at all**, and the
failure surfaces as a Python traceback rather than a validation message.

With the commas restored, the tool runs and the core models validate:

```
PASS: cloudevents\model.json
PASS: core\model.json
PASS: endpoint\model.json
PASS: message\model.json
FAIL: schema\model.json
```

## Scope

Two characters, one file. This is a pure syntax repair -- no schema semantics
are changed. The comma placement matches the surrounding style, and
`constraints` remains a sibling of `ximportresources` and `resources`.

## Relationship to #512

This is a **prerequisite** for #512, not a fix for it. #512 notes the
unparseable file under "Note on discoverability"; this PR is that repair
upstreamed. The remaining `FAIL: schema\model.json` above is exactly the
condition #512 describes, and its two defects -- the missing `validateformat` /
`validatecompatibility` / `strictvalidation` keys on `ResourceDefinition`, and
the `constraints` map being modelled one level too shallow -- are untouched
here and left for a separate change.

### DIFF
--- a/core/model.schema.json
+++ b/core/model.schema.json
@@ -99,9 +99,9 @@
                 "equals": {
                   "type": "string"
                 }
-              }
+              },
               "description": "Group level constraints over resources"
-            }
+            },
             "resources": {
               "type": "object",
               "description": "Resource types within this group",


duglin:
@clemensv needs to be signed

duglin:
Approved on 8/4 - merging

PR #517, opened 2026-08-03
docs: update endpoint.messagegroups and message.basemessage in spec files to use correct type

thirschel-microsoft:
A recent change (https://github.com/xregistry/spec/commit/0e76232e6d5d125a771cf8c662fefd8fbf24a03e#diff-4790cbfb37d699e736007395731e1952053fa95a56d9a84605b94e66aa1252c5) seems to replaced an update to the spec we made back in https://github.com/xregistry/spec/pull/467

Updating the documentation to be compliant with the spec

### DIFF
--- a/cloudevents/schemas/document-schema.avsc
+++ b/cloudevents/schemas/document-schema.avsc
@@ -604,7 +604,7 @@
                 }
               },
               "name": "messagegroups",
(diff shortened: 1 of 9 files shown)


duglin:
Executive decision - merging due it being a revert of previous work by mistake.

PR #521, opened 2026-08-05
Week of Aug 3

### LINKED ISSUE #503: Is "manual" versionmode a bit of a lie?
Today, versionmode of "manual" will set new version's ancestorid to that of the current default version. And then when its ancestor is deleted it becomes its own root.

Is this inconsistent that it's really not "manual" in the create flow but it is in the "ancestor deleted" flow?

Options - not mutually exclusive:
1 - leave it as is
2 - don't point to the current latest during creation - make it fully manual
3 - create a new special 'ancestorid' value of "latest" so in "manual" mode they can choose to get today's behavior, if we do 2
4 - rename today's version mode to "semimanual", define a "manual" mode to support fully manual mode
5- keep the create flow as is, but change the delete flow so that the version inherits the deleted version's ancestorid

### LINKED ISSUE #513: Filtering pains
Registry has:
- `dirs/files` : 10 of them
- `schemagroups/schemas`: 5 of them

Filter:`?filter = schemagroups.schemas.epoch > 0 OR ( epoch=5 AND dirs.files.epoch>10 )`

Result:
if `registry.epoch=5`:
- `schemagroupsurl: ?filter = schemas.epoch > 0`
- `schemagroupcount: 5`
- `dirsurl: ?filter = dirs.files.epoch > 10`
- `dirscount: 10`

if `registry.epoch<>5`:
- `schemagroupsurl: ?filter = schemas.epoch > 0`
- `schemagroupcount: 5`
- `dirsurl: ?filter = dirs.files.epoch > 10`  ??? or `dirsurl: ?filter=excludeall`  ???
- `dirscount: 0`

Discussions:
- what should `dirsurl` show?
  - `> 10` will not yield the same results when run in isolation
  - `excludeall` will yield the correct results per "this" result set, but not a reusable query
- abstractly: when do we use `excludeall`?
  - when this branch isn't accessible?
  - whenever the results set is zero regardless of the reason? (inaccessible branch, sub-tree doesn't match filter)

- - -

Filter:`?filter = schemagroups.schemas.epoch > 10 OR epoch`

Result: entire registry

What should `schemagroupsurl` show for its filter?
- `schemas.epoch > 10` ?
- no filter ?

First only requires syntax/model evaluation. But is incomplete/wrong.
Second requires data analysis and understanding which entity was returned as a result of each `OR`,  and then only include the proper `OR` expressions on the sub-filter. E.g.
  - for schemas with epoch < 11 -> 2nd expr
  - for schemas with epoch > 10 -> 1st and 2nd expr
  - Result: using expr that appears on ALL returned entities -> 2nd expr (but a subset, so "no filter/expr" in this case)
  - If the result set is empty, what should `schemagroupsurl` show? This is related to the `excludeall` questions above.
    - E.g. `?filter = (schemagroups.schemas.epoch > 10 AND epoch<0) OR epoch`. There is no filter for `schemagroupsurl` that'll yield the same results as "this query" except `excludeall`.

- - -

Options - not mutually exclusive:
1 - leave as is - ambiguous
2 - Remove `OR` support. This makes ALL of this go away, but puts the burden of merging of results on the client
3 - ALL collections with zero results get a filter of: `?filter = excludeall` regardless of why they're zero.
4 - For non-zero collections, the server MAY (not MUST) show a filter to yield "this" result set. Leaves it as an impl choice as to whether they want to deal with the complexity of the problem or not.
5 - include the full query on each sub-tree's filter, even though it'll reference entities/attributes that not within scope of that sub-tree.  E.g. `versionsurl=?filter=schemagroups.schemas.versions.versionsid=v1` instead of just `versionsurl=?filter=versionid=v1`
6 - drop the entire idea of `filter` being in the `<COLLECTION>url` and treat it more like we treat `self`, a static URL reference to the item irrespective of the "view" or how they got there

Proposal:
- either: "3 and 4" or 6

### LINKED ISSUE #520: spec needs to say that the Registry's epoch/modifiedat are updated upon a model update
Already says this for capabilities

duglin:
- make it clear the Registry's epoch/modifiedat are updated when model and capabilities are updated
- clarify 'manual' mode and make it clear '/model' must be supported and exposed via capabilities
- require the use of 'excludeall' for all empty collections when using filtering

Fixes #520 
Fixes #503
Fixes #513

### DIFF
--- a/core/model.md
+++ b/core/model.md
@@ -868,6 +868,12 @@ Note that this feature has similar results as setting the Resource attribute's
     - Deleted Ancestor: if a Version's ancestor is deleted, then this Version
       MUST become a root, and its `ancestorid` value MUST be set to its
       `versionid` value.
+    - The name of this versionmode is a bit of a misnomer in that it is not
+      100% "manual". As stated above, as a convenience, by default new
+      Versions will point to the "latest" Version as its ancestor. If this
+      behavior is not desired, then during the "create" operation, the
+      `ancestorid` can be set to point to itself (or any other Version) if
+      desired.
 
   - `createdat`
     - Newest Version: MUST be determined by finding the Version with the
--- a/core/spec.md
+++ b/core/spec.md
@@ -4096,12 +4096,11 @@ when a query includes a filter expression, any nested xRegistry collections
 in the response SHOULD result in the specification of a `<COLLECTION>url`
 attribute that allows clients to easily traverse into that collection with
 the appropriated sub-filter expression to yield the same results as the
-original filter, but just for that nested collection. However, if the filters
-results in zero entities for a particular collection due to the filter
-excluding that collection entirely, rather than the `<COLLECTION>url` not
-having any filter expression at all (which would result in the full collection
-being returned), servers MUST use the special filter expression `excludeall`
-to indicate no results are to be returned for that URL.
+original filter, but just for that nested collection. However, if a collection
+has zero entities (regardless of whether the empty list is due to the filter
+or due to the collection being empty), server MUST use a special filter
+expression, `excludeall`, to indicate no results are to be returned for that
+URL.
 
 For example, using HTTP:
 ```yaml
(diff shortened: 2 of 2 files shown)


duglin:
Approved on the 8/6 call

PR #538, opened 2026-08-11
Fix grammar and typographical errors in endpoint spec

### LINKED ISSUE #534: Fix grammar and typographical errors in endpoint/spec.md
## Summary

The following 14 grammar and typographical issues remain in
`endpoint/spec.md`, grouped into 13 findings. Each item includes sentence
context and a minimal correction reviewed to preserve technical meaning. Line
links refer to commit
[`af777bc`](https://github.com/xregistry/spec/commit/af777bc25778cc7759b007e99bede6bf7d371a27).

1. **[Line 123](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L123)**

   > For describing a message broker queue, the best approach is to define a
   > message group that defines the messages that are be allowed to flow on the
   > queue...

   Suggested: "...the messages that are allowed to flow..."

2. **[Lines 142-143](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L142-L143)**

   > A possible application of the latter are Endpoint definitions that define
   > endpoint patterns as they are common for applications of the MQTT protocol.

   Suggested: "The latter can be used for Endpoint definitions that define
   endpoint patterns common in applications of the MQTT protocol."

3. **[Lines 144-149](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L144-L149)**

   > For instance, the Eclipse Sparkplug protocol is a convention that defines
   > endpoint roles and messages and schemas for MQTT, and the SparkPlugB
   > example that illustrates how Registry Endpoints can model the convention in
   > formal terms.

   Suggested: remove `that` before `illustrates`.

4. **[Lines 162-163](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L162-L163)**

   > Specification-defined, and server-known extension, attributes MUST
   > generate an error if the corresponding feature is not supported or
   > enabled.

   Suggested: "Specification-defined attributes and server-known extension
   attributes..."

5. **[Lines 164-166](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L164-L166)**

   > However, as with all attributes, if accepting the attribute would result in
   > a bad state (such as exceeding a size limit, or results in a security
   > issue), then the server MAY choose to reject the request.

   Suggested: change the parenthetical to "(such as exceeding a size limit or
   resulting in a security issue)".

6. **[Lines 187-190](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L187-L190)**

   > The Endpoint Registry is a registry of metadata definitions for abstract
   > and concrete network endpoints to which messages can be produced, from
   > which messages can be consumed, or which makes messages available via
   > subscription and delivery to a consumer-designated endpoint.

   Suggested: "...or which make messages available..."

7. **[Line 438](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L438)**

   > MUST be an array of at least one.

   Suggested: "MUST be an array of at least one value."

8. **[Lines 510-512](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L510-L512) - two occurrences**

   > Additionally, if a Group does not have a `envelope` attribute then there
   > are no constraints on its Resources `envelope` attributes.

   Suggested: "Additionally, if a Group does not have an `envelope` attribute,
   then there are no constraints on its Resources' `envelope` attributes."

9. **[Lines 573-575](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L573-L575)**

   > SHOULD follow the naming convention `<NAME>/<VERSION>`, whereby `<NAME>` is
   > the name of the protocol and `<VERSION>` is the version of protocol.

   Suggested: "...the version of the protocol."

10. **[Lines 584-586](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L584-L586)**

    > Configuration details of the endpoint related to the protocol used to
    > transmit the messages. An endpoint MAY be defined without detail
    > configuration. In this case, the endpoint is considered to be "abstract".

    Suggested: "...without detailed configuration."

11. **[Lines 674-675](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L674-L675)**

    > `Plain`: The client uses username with a plaintext password for
    > authentication and authorization.

    Suggested: "...uses a username with a plaintext password..."

12. **[Lines 821-824](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L821-L824)**

    > Using the same `messageid` value is not possible (even though the
    > CloudEvent `type` attribute would be the same for both runtime messages),
    > so one (or both) message definition's `messageid` values might not match
    > the runtime message's `type` value.

    Suggested: "...so the `messageid` value of one message definition (or the
    values of both) might not match the runtime message's `type` value."

13. **[Lines 830-833](https://github.com/xregistry/spec/blob/af777bc25778cc7759b007e99bede6bf7d371a27/endpoint/spec.md?plain=1#L830-L833)**

    > Implementations MAY choose to generate an error if it detects duplicate
    > `messageid` values across the `messages` collection message definitions
    > and the `messagegroups` referenced message definitions, if that is the
    > desired constraint for their users.

    Suggested: "...if they detect duplicate `messageid` values..."

clemensv:
## Summary

- correct the 14 retained grammar and typographical issues in `endpoint/spec.md`
- preserve xRegistry terminology, Markdown formatting, and normative requirements

## Validation

- `tools/spellcheck endpoint/spec.md`
- `tools/tabcheck endpoint/spec.md`
- `python tools/verify.py endpoint`

Closes #534

### DIFF
--- a/endpoint/spec.md
+++ b/endpoint/spec.md
@@ -120,7 +120,7 @@ a message registry, and which have the advantage of being shareable across
 multiple Endpoints.
 
 For describing a message broker queue, the best approach is to define a
-message group that defines the messages that are be allowed to flow on the
+message group that defines the messages that are allowed to flow on the
 queue and for that message group to then be referenced from the `producer` and
 the `consumer` endpoints of the queue, whereby both of those endpoints share
 the same `channel` identifier.
@@ -139,13 +139,13 @@ the endpoint definition is to be treated like a template where configuration
 elements like the endpoint URI will have to be supplied by external
 configuration for the client to become functional.
 
-A possible application of the latter are Endpoint definitions that define
-endpoint patterns as they are common for applications of the MQTT protocol.
+The latter can be used for Endpoint definitions that define endpoint patterns
+common in applications of the MQTT protocol.
 For instance, the [Eclipse Sparkplug](https://sparkplug.eclipse.org/) protocol
 is a convention that defines endpoint roles and messages and schemas for MQTT,
 and the
 [SparkPlugB](../cloudevents/samples/scenarios/mqtt-sparkplugB.xreg.json)
-example that illustrates how Registry Endpoints can model the convention in
+example illustrates how Registry Endpoints can model the convention in
 formal terms.
 
 ## Notations and Terminology
@@ -159,10 +159,11 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 For clarity, OPTIONAL attributes (specification-defined and extensions) are
 OPTIONAL for clients to use, but the servers' responsibility will vary.
 Server-unknown extension attributes MUST be silently stored in the backing
-datastore. Specification-defined, and server-known extension, attributes MUST
-generate an error if the corresponding feature is not supported or enabled.
+datastore. Specification-defined attributes and server-known extension
+attributes MUST generate an error if the corresponding feature is not supported
+or enabled.
 However, as with all attributes, if accepting the attribute would result in a
-bad state (such as exceeding a size limit, or results in a security issue),
+bad state (such as exceeding a size limit or resulting in a security issue),
 then the server MAY choose to reject the request.
 
 In the pseudo JSON format snippets `?` means the preceding attribute is
@@ -186,7 +187,7 @@ available via subscription for delivery to a consumer-designated endpoint.
 
 The Endpoint Registry is a registry of metadata definitions for abstract and
 concrete network endpoints to which messages can be produced, from which
-messages can be consumed, or which makes messages available via subscription
+messages can be consumed, or which make messages available via subscription
 and delivery to a consumer-designated endpoint.
 
 As discussed in the [CloudEvents Registry overview](../cloudevents/spec.md),
@@ -435,7 +436,7 @@ to the xRegistry-defined core
     - "subscriber"
     - "consumer"
     - "producer"
-  - MUST be an array of at least one.
+  - MUST be an array of at least one value.
   - MUST declare exactly one usage value, except for the single combination
     permitted below.
   - MUST NOT combine `producer` with any other value.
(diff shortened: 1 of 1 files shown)


endpoint/spec.md
```diff
@@ -159,10 +159,11 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
 For clarity, OPTIONAL attributes (specification-defined and extensions) are
 OPTIONAL for clients to use, but the servers' responsibility will vary.
 Server-unknown extension attributes MUST be silently stored in the backing
-datastore. Specification-defined, and server-known extension, attributes MUST
-generate an error if the corresponding feature is not supported or enabled.
+datastore. Specification-defined attributes and server-known extension
```
duglin, line 162: if we want this new text then we should apply it to all specs.

endpoint/spec.md
```diff
@@ -159,10 +159,11 @@ interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
...
+attributes MUST generate an error if the corresponding feature is not supported
+or enabled.
 However, as with all attributes, if accepting the attribute would result in a
-bad state (such as exceeding a size limit, or results in a security issue),
+bad state (such as exceeding a size limit or resulting in a security issue),
```
duglin, line 166: same here

duglin:
LGTM we just need to decide if we want to keep that the RFC text and apply it to all specs, or not

duglin:
Agreed on the 8/12 call - we'll do the RFC text in anther PR across all specs

PR #547, opened 2026-08-12
First pass at core spec review

duglin:


core/spec.md
```diff
@@ -3977,11 +3950,11 @@ value for the purpose of the server's
...
 The `filter` flag MAY be used to indicate that the response MUST include only
 those entities that match the specified filter criteria expressions. This
-means that any Registry Collection's attributes MUST be modified to match the
-resulting subset. In particular:
+means that any Registry Collection's `<COLLECTION*` attributes will be
+modified to match the resulting subset. In particular:
 - If the collection is inlined, it MUST only include entities that match the
   filter expression(s).
-- The collection `url` attribute MUST include the appropriate filter
+- The collection `url` attribute MAT include the appropriate filter
```
clemensv: s/MAT/MAY

core/spec.md
```diff
@@ -3977,11 +3950,11 @@ value for the purpose of the server's
 
 The `filter` flag MAY be used to indicate that the response MUST include only
 those entities that match the specified filter criteria expressions. This
-means that any Registry Collection's attributes MUST be modified to match the
-resulting subset. In particular:
+means that any Registry Collection's `<COLLECTION*` attributes will be
```
clemensv: `<COLLECTION*` as a notation is a little strange

core/spec.md
```diff
@@ -4680,7 +4646,7 @@ syntax guidelines are RECOMMENDED for other common situations:
   - E.g. `set myarray[3:]=mary` would be used to insert "mary" at the 4th
     index (zero-based), pushing current index positions 3 (and higher) further
     down in the array.
-  - If the specified integer value does not exist, then an error MUST be
+  - If the specified integer value does not exist, then an error would be
```
clemensv: correct change?

clemensv (approved):
LGTM after addressing the nits

core/spec.md

core/spec.md
duglin: missed the > 
done

core/spec.md
duglin: this goes to what I mentioned yesterday... we have no control over the cli/tooling so the best we can do is make recommendations - so in this case I felt like a MUST was inappropriate. If we want to phrase this entire section as "here's our guidance and IF you choose follow it, then ... MUST....".  Would you prefer that? or leave it as is?

PR #548, opened 2026-08-13
sync rfc text

duglin:


duglin:
Yesterday, we agreed to sync the text once we agreed to what's in the core spec... so merging.

PR #556, opened 2026-08-18
Week of Aug 17

### LINKED ISSUE #544: Is auto-creation of parent an HTTP thing or generic?
Core spec has:

```
### Design: Implicit Creation of Parent Entities

To reduce the number of interactions needed when creating an entity, all
nonexisting parent entities specified as part of `<PATH>` to the entity MUST
be implicitly created. Each of those entities MUST be created with the
appropriate `<SINGULAR>id` specified in the `<PATH>`. If any of those
entities have REQUIRED attributes, then they cannot be implicitly created, and
would need to be created directly. This also means that the creation of the
original entity would fail and generate an error
([required_attribute_missing](./spec.md#required_attribute_missing)) for the
appropriate parent entity.
```

Is this http specific or generic? Either move it or make it less HTTP specific... the PATH part seems HTTP specific to me.

I'm leaning towards making it generic.

### LINKED ISSUE #551: group constraint - matchcase from group or resource attributes?
```
### `groups.<STRING>.constraints.<RESOURCES>.<PATH>.equals`
...
When present, this aspect MUST contain the
[dot (`.`) notation](spec.md#xregistry-dot--notation) path in the
Group instance that the referenced Resource attribute MUST match. In the
case of the attribute type being a `string`, the comparison MUST take into
account the [matchcase](#attributesstringmatchcase) aspect of the attribute.
```
Not sure if we intended it to be the Group's attr or Resource's?

Options:
1 - use Resource
2 - use Group
3 - define it per constraint
4 - remove the entire notion of "matchcase" from the spec and always do things case-insensitively

duglin:
- mention some readonly attributes are inspected by server
- make auto-creation of parents a non-http specific thing
- make it clear that even when shortself is disabled, the attr is still reserved
- drop support for matchcase

Fixes: #544
FIxes: #551

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,10 @@ The project includes a "core" specification that defines the basic model and
 APIs of a Registry and a set of domain-specific specifications that extend that
 core for particular use cases. By leveraging the same "core" model/APIs,
 generic tooling and common patterns of Registry access can be developed which
-help create an interoperable (and standard) interface.
+help create an interoperable (and standard) interface. See
+[A Quick Introduction to
+xRegistry](core/primer.md#3-a-quick-introduction-to-xregistry) for more
+information.
 
 It is expected that further specifications will be developed, both as part
 of the xRegistry project and outside, that will leverage this framework for
--- a/core/model.json
+++ b/core/model.json
@@ -9,7 +9,6 @@
     },
     "registryid": {
       "type": "string",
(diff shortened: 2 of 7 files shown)


lailabougria (approved):

duglin:
Approved by sub-team during 8/19 call

PR #558, opened 2026-08-20
fix(events): correct editorial errors in core/events.md

thirschel-microsoft:
Some small changes to the examples and phrasing

The note excluding events said "created and updated", but registry.updated is defined in detail immediately above. Changed to "created and deleted", matching the operations actually outside the spec.

A few questions as well:

- Resource created requires attributes in changed, while the data section says changed MUST NOT appear on created or deleted events.
- Resource create/delete say they generate registry.updated carrying <RESOURCES>, but those attributes belong to the containing Group, and group.updated is already defined for that case.

### DIFF
--- a/core/events.md
+++ b/core/events.md
@@ -326,7 +326,7 @@ events.
     `deprecated` sub-object that were modified.
 
   - An `io.xregistry.group.updated` event will also be generated where the
-    `deprecated` attribute will be included in `changed`, if present.
+    `deprecated` sub-object will be included in `changed`, if present.
 
 - Action: `deleted`
   - MUST be generated when a Group is deleted.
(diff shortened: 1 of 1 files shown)


core/events.md
thirschel-microsoft, line 192: This seems to contradict with Line 341 "MUST be generated when a new Resource is created, where each modified attribute MUST be included in changed, if present."

thirschel-microsoft, line 347: The confusion here is that <RESOURCES> collections are attributes of the containing Group rather than the Registry. Lines 306-313 already define io.xregistry.group.updated for this exact situation: "MUST be generated when a Resource is created or deleted, where changed, if present, includes (at least) epoch, modifiedat, <RESOURCES> and <RESOURCES>count attribute names."

I read lines 347-349 and 396-398 as intending group.updated, with registry.updated carried down from the Group section by mistake. Can you confirm?

duglin: Yup, sure looks like a typo to me.

duglin: per your previous comment, I think this line should be dropped.

duglin: seems weird we refer to it as "meta attribute" instead of "meta object" (or sub-object)

duglin: If you're up for it, go ahead and remove the stuff after the comma.
We discussed this today and agreed it's wrong.

duglin: Agreed on today's call: s/attribute/sub-object/g in the entire doc. If you don't want to do this let me know and I'll open a separate PR.

duglin: Agreed this is a typo on today's call.

duglin: Yup - go ahead and change 341

thirschel-microsoft:
@duglin Thanks for the quick feedback. Updated the PR based on your comments.

core/events.md
thirschel-microsoft: This seems like a bigger change so I'd leave it a followup PR

duglin:
Approved on the 8/21 call

Issue #525, opened 2026-08-11
core/primer - repetitive content


clemensv:
### “4. Value proposition” and “5. Motivation”

- “5. Motivation” repeats the discovery, validation, versioning, serialization, extension-discovery, and vendor-neutrality arguments already developed under:

    - “4.2. Discovery”
    - “4.3. Vendor-agnostic”
    - “4.4. Versioning”
    - “4.5. Schema validation”

### “4.7. Schema-based data contract generation,” “7.1. File,” and “9.3. Metadata files in repositories”

All three discuss storing xRegistry metadata beside source code, keeping it in source control, and using a file as a repository-level manifest.

### “7. Representations,” “7.2. Static File Server,” and “7.3. API Server”

The comparison between file-based representations and API servers is repeated: direct writes, server-side search and filtering, export, hosting cost, rebuilding static content, and suitability for larger shared registries.

### “8. Embeddings, References, and Federation,” “8.1. Formats and Content/Media-Types,” and “8.2. Embeddings and Links”

These sections repeatedly establish that xRegistry catalogs metadata, can embed documents, can reference external documents, and cannot validate externally mutable content. Within “8.2. Embeddings and Links,” the schemaurl paragraph and the following external-links paragraph also overlap.

### “4.1. Why build something new?”, “8.2. Embeddings and Links,” and “8.3.4. API Servers and Proxies”

All three explain that existing registries can remain authoritative while xRegistry projects, catalogs, overlays, or proxies them through a common model.

### “10.7.7. Choosing unique Group and Resource names” and “16. Defining Extensions”

Both recommend domain- or company-specific names to avoid collisions with names from other registries or future xRegistry work.

### “10.7.1. Default Version and Maximum Versions,” “10.7.11. Multi-root ancestor hierarchies,” and “10.7.13. Pruning Versions with singleversionroot enabled”

The basic version-pruning mechanism is explained repeatedly. “10.7.11” and “10.7.13” need the pruning context, but they restate more of the general algorithm than necessary.

### “4.5. Schema validation,” “8.1. Formats and Content/Media-Types,” “10.7.15. Can validatecompatibility and validateformat be true for file-servers?”, “11. The format attribute in the Schema spec,” and “17. Validation”

These sections repeatedly introduce format validation, compatibility validation, capability indicators, and the distinction between validation performed by a server and validation asserted by a representation.

### “10.2. Valid Characters,” “10.3. Extensions,” and “10.7.4. Naming and Case Sensitivity”

[...]

PR #559, opened 2026-08-21
Reduce repetition in the core primer

clemensv:
## Summary

- consolidate repeated explanations in the core primer
- remove sections that no longer carry independent technical value
- correct ID casing and shadowing explanations
- preserve historical heading links with explicit compatibility anchors

## Validation

- `py -3 tools/verify.py core`
- repository-equivalent spelling and whitespace checks
- `git diff --check`

Relates to #525.

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -1,6 +1,6 @@
 # xRegistry Primer
 
-<!-- words: validatecompatibility validateformat strickvalidation -->
+<!-- words: validatecompatibility validateformat -->
 <!-- words: strictvalidation matchversions readme upserts -->
 
 <!-- no verify-specs -->
@@ -29,7 +29,6 @@ normative technical details.
     - [3.2.6. Errors](#326-errors)
   - [3.3. Defining a Basic Model](#33-defining-a-basic-model)
     - [3.3.1. Adding your own attributes (extensions)](#331-adding-your-own-attributes-extensions)
-  - [3.4. Where to go next](#34-where-to-go-next)
 - [4. History](#4-history)
 - [5. Value proposition](#5-value-proposition)
   - [5.1. Why build something new?](#51-why-build-something-new)
@@ -38,61 +37,56 @@ normative technical details.
   - [5.4. Versioning](#54-versioning)
   - [5.5. Schema validation](#55-schema-validation)
   - [5.6. Payload reduction](#56-payload-reduction)
-  - [5.7. Schema-based data contract generation](#57-schema-based-data-contract-generation)
-  - [5.8. Producer and consumer generation](#58-producer-and-consumer-generation)
-  - [5.9. Basis for further developments](#59-basis-for-further-developments)
-- [6. Motivation](#6-motivation)
-- [7. Design Goals](#7-design-goals)
-  - [7.1. Non-Goals](#71-non-goals)
-- [8. Representations](#8-representations)
-  - [8.1. File](#81-file)
-  - [8.2. Static File Server](#82-static-file-server)
-  - [8.3. API Server](#83-api-server)
-- [9. Embeddings, References, and Federation](#9-embeddings-references-and-federation)
-  - [9.1. Formats and Content/Media-Types](#91-formats-and-contentmedia-types)
-  - [9.2. Embeddings and Links](#92-embeddings-and-links)
-  - [9.3. Federation](#93-federation)
-    - [9.3.1. Composing registries in memory](#931-composing-registries-in-memory)
-    - [9.3.2. Shadowing](#932-shadowing)
-    - [9.3.3. Identifiers](#933-identifiers)
-    - [9.3.4. API Servers and Proxies](#934-api-servers-and-proxies)
-- [10. Possible Use Cases](#10-possible-use-cases)
-  - [10.1. CloudEvents](#101-cloudevents)
-  - [10.2. Business objects](#102-business-objects)
-  - [10.3. Metadata files in repositories](#103-metadata-files-in-repositories)
-- [11. Design decisions or topics of interest](#11-design-decisions-or-topics-of-interest)
-  - [11.1. Resource.ID vs Resource.Version.ID](#111-resourceid-vs-resourceversionid)
-  - [11.2. Valid Characters](#112-valid-characters)
-  - [11.3. Extensions](#113-extensions)
-  - [11.4. Group organization](#114-group-organization)
-  - [11.5. Deleting entities](#115-deleting-entities)
-  - [11.6. Detection of Referenced Resources](#116-detection-of-referenced-resources)
-  - [11.7. Shared/Referenced Resources](#117-sharedreferenced-resources)
-  - [11.8. Default Version and Maximum Versions](#118-default-version-and-maximum-versions)
-  - [11.9. Potential Extensions](#119-potential-extensions)
-  - [11.10. Why Epoch?](#1110-why-epoch)
-  - [11.11. Naming and Case Sensitivity](#1111-naming-and-case-sensitivity)
-  - [11.12. Why the lower character limit on some Group and Resource type names?](#1112-why-the-lower-character-limit-on-some-group-and-resource-type-names)
-  - [11.13. Why must Group type and Resource type names be valid attribute names?](#1113-why-must-group-type-and-resource-type-names-be-valid-attribute-names)
-  - [11.14. Choosing unique Group and Resource names](#1114-choosing-unique-group-and-resource-names)
-  - [11.15. Are `self` and `shortself` attributes static?](#1115-are-self-and-shortself-attributes-static)
-  - [11.16. Why does an unknown query parameter not generate an error?](#1116-why-does-an-unknown-query-parameter-not-generate-an-error)
-  - [11.17. Updating attributes with `ifvalues`](#1117-updating-attributes-with-ifvalues)
-  - [11.18. Multi-root ancestor hierarchies](#1118-multi-root-ancestor-hierarchies)
-  - [11.19. `singleversionroot` Policy Enforcement](#1119-singleversionroot-policy-enforcement)
-  - [11.20. Pruning Versions with `singleversionroot` enabled](#1120-pruning-versions-with-singleversionroot-enabled)
-  - [11.21. What's the oldest/newest Version of a Resource?](#1121-whats-the-oldestnewest-version-of-a-resource)
-  - [11.22. Can `validatecompatibility` and `validateformat` be `true` for file-servers?](#1122-can-validatecompatibility-and-validateformat-be-true-for-file-servers)
-  - [11.23. Detecting circular references between Versions](#1123-detecting-circular-references-between-versions)
-  - [11.24. Optional required fields in requests](#1124-optional-required-fields-in-requests)
-  - [11.25. Deprecation of entities in an xRegistry](#1125-deprecation-of-entities-in-an-xregistry)
-- [12. The `format` attribute in the Schema spec](#12-the-format-attribute-in-the-schema-spec)
-- [13. Problematic characters in attributes](#13-problematic-characters-in-attributes)
-- [14. Why do Resources have 3 levels of data?](#14-why-do-resources-have-3-levels-of-data)
-- [15. Why are some (conditional) read-only attributes not ignored?](#15-why-are-some-conditional-read-only-attributes-not-ignored)
-- [16. Why isn't `PUT` idempotent?](#16-why-isnt-put-idempotent)
-- [17. Defining Extensions](#17-defining-extensions)
-- [18. Validation](#18-validation)
+  - [5.7. Schema-based contract and client generation](#57-schema-based-contract-and-client-generation)
+  - [5.8. Basis for further developments](#59-basis-for-further-developments)
+- [6. Non-Goals](#7-design-goals)
+- [7. Representations](#8-representations)
+  - [7.1. File](#81-file)
+  - [7.2. Static File Server](#82-static-file-server)
+  - [7.3. API Server](#83-api-server)
+- [8. Embeddings, References, and Federation](#9-embeddings-references-and-federation)
+  - [8.1. Formats and Content/Media-Types](#91-formats-and-contentmedia-types)
+  - [8.2. Embeddings and Links](#92-embeddings-and-links)
+  - [8.3. Federation](#93-federation)
+    - [8.3.1. Composing registries in memory](#931-composing-registries-in-memory)
+    - [8.3.2. Shadowing](#932-shadowing)
+    - [8.3.3. Identifiers](#933-identifiers)
+    - [8.3.4. API Servers and Proxies](#934-api-servers-and-proxies)
+- [9. Possible Use Cases](#10-possible-use-cases)
+  - [9.1. CloudEvents](#101-cloudevents)
+  - [9.2. Business objects](#102-business-objects)
+  - [9.3. Metadata files in repositories](#103-metadata-files-in-repositories)
+- [10. Design decisions or topics of interest](#11-design-decisions-or-topics-of-interest)
+  - [10.1. Resource.ID vs Resource.Version.ID](#111-resourceid-vs-resourceversionid)
+  - [10.2. Valid Characters](#112-valid-characters)
+  - [10.3. Extensions](#113-extensions)
+  - [10.4. Group organization](#114-group-organization)
+  - [10.5. Deleting entities](#115-deleting-entities)
+  - [10.6. Detection of Referenced Resources](#116-detection-of-referenced-resources)
+  - [10.7. Shared/Referenced Resources](#117-sharedreferenced-resources)
+  - [10.8. Default Version and Maximum Versions](#118-default-version-and-maximum-versions)
+  - [10.9. Why Epoch?](#1110-why-epoch)
+  - [10.10. Naming and Case Sensitivity](#1111-naming-and-case-sensitivity)
+  - [10.11. Why are Group and Resource type names more restricted?](#1112-why-the-lower-character-limit-on-some-group-and-resource-type-names)
+  - [10.12. Choosing unique Group and Resource names](#1114-choosing-unique-group-and-resource-names)
+  - [10.13. Are `self` and `shortself` attributes static?](#1115-are-self-and-shortself-attributes-static)
+  - [10.14. Why does an unknown query parameter not generate an error?](#1116-why-does-an-unknown-query-parameter-not-generate-an-error)
+  - [10.15. Updating attributes with `ifvalues`](#1117-updating-attributes-with-ifvalues)
+  - [10.16. Multi-root ancestor hierarchies](#1118-multi-root-ancestor-hierarchies)
+  - [10.17. `singleversionroot` Policy Enforcement](#1119-singleversionroot-policy-enforcement)
+  - [10.18. Pruning Versions with `singleversionroot` enabled](#1120-pruning-versions-with-singleversionroot-enabled)
+  - [10.19. What's the oldest/newest Version of a Resource?](#1121-whats-the-oldestnewest-version-of-a-resource)
+  - [10.20. Can `validatecompatibility` and `validateformat` be `true` for file-servers?](#1122-can-validatecompatibility-and-validateformat-be-true-for-file-servers)
+  - [10.21. Detecting circular references between Versions](#1123-detecting-circular-references-between-versions)
+  - [10.22. Optional required fields in requests](#1124-optional-required-fields-in-requests)
+  - [10.23. Deprecation of entities in an xRegistry](#1125-deprecation-of-entities-in-an-xregistry)
+- [11. Problematic characters in values](#13-problematic-characters-in-values)
+- [12. Why do Resources have 3 levels of data?](#14-why-do-resources-have-3-levels-of-data)
+- [13. Why are some (conditional) read-only attributes not ignored?](#15-why-are-some-conditional-read-only-attributes-not-ignored)
+- [14. Why isn't `PUT` idempotent?](#16-why-isnt-put-idempotent)
+- [15. Validation edge cases](#18-validation-edge-cases)
+  - [15.1. Empty format capabilities](#181-empty-format-capabilities)
+  - [15.2. Constraints and `matchversions`](#182-constraints-and-matchversions)
 
 
 ## 3. A Quick Introduction to xRegistry
@@ -126,7 +120,7 @@ Think of it like a filesystem exposed over HTTP:
   or more Versions over time. If you don't care about versioning, just
   ignore Versions and treat the Resource as if it has one version.
 
-A picture is worth it:
+The entity hierarchy is:
 
 ```
 Registry
@@ -177,7 +171,9 @@ defines (e.g. `endpoints`, `messages`) - they're not literal.
 
 #### 3.2.1. Reading things
 
-A plain `GET` on any entity returns its metadata as JSON:
+A plain `GET` on a Registry, Group, metadata-only Resource, or Version returns
+its metadata as JSON. Document Resources are the exception described in
+section 3.2.3.
 
 ```
 GET /endpoints/ep1
@@ -233,10 +229,9 @@ HTTP/1.1 201 Created
 
 #### 3.2.3. Resources: metadata vs. the actual document
 
-This is the one twist compared to a plain REST API. A Resource (e.g. a
-`message` or a `schema`) can have its own domain-specific "document" (its
-actual content) in addition to its xRegistry metadata (`epoch`, `name`,
-etc).
+Document Resources differ from metadata-only Resources. A Resource such as a
+`message` or `schema` can have domain-specific content in addition to its
+xRegistry metadata (`epoch`, `name`, etc.).
 
 - `GET /endpoints/ep1/messages/msg1` → returns the Resource's **document**
   (its raw content) directly in the HTTP body.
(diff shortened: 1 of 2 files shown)


core/primer.md
```diff
@@ -177,7 +171,9 @@ defines (e.g. `endpoints`, `messages`) - they're not literal.
 
 #### 3.2.1. Reading things
 
-A plain `GET` on any entity returns its metadata as JSON:
+A plain `GET` on a Registry, Group, metadata-only Resource, or Version returns
+its metadata as JSON. Document Resources are the exception described in
```
duglin: I wonder if "Document Resources" would read better as "Resource that have domain-specific documents" would read better - otherwise it sounds like we have two types of Resource entities

core/primer.md
```diff
@@ -233,20 +233,18 @@ HTTP/1.1 201 Created
...
   (its raw content) directly in the HTTP body.
 - `GET /endpoints/ep1/messages/msg1$details` → returns the Resource's
   **metadata** as JSON instead (with the document's metadata such as
   `contenttype`, but not the document body itself).
 
-Same idea applies for `PUT`: `PUT .../msg1` with a raw body sets the
-document content; `PUT .../msg1$details` with a JSON body sets/updates just
-the metadata.
+For `PUT`, `PUT .../msg1` with a raw body sets the document content, while
+`PUT .../msg1$details` with a JSON body updates only the metadata.
```
duglin: I know the original has this issue too, but we should probably say...  with a JSON body allows for updating both the metadata and the document content.

core/primer.md
```diff
@@ -359,20 +357,9 @@ Common `type` values you'll use most often: `string`, `boolean`, `integer`,
...
-- [`spec.md`](./spec.md) - the full core specification (concepts, entities,
-  capabilities).
-- [`http.md`](./http.md) - the full HTTP binding (every API, header, and
-  query parameter in detail).
-- [`model.md`](./model.md) - the full model specification (every model
-  attribute and option).
-
+This model is sufficient to create Groups and Resources. The full model
+specification covers constraints, versioning policies, `xid` targets,
+`ifvalues`, and other advanced scenarios; see [`model.md`](./model.md).
```
duglin, line 401: We should add back in a link to the spec itself - that's probably the most important one. I actually prefer the old list.

core/primer.md
```diff
@@ -429,29 +416,23 @@ that can be used to manage and discover the full metadata graph of a system,
...
-consume events from endpoints have a single point of truth to refer to. In
-addition, xRegistry allows specifying extensions and metadata that can further
-describe events on more detailed levels. As an example, it is possible to define
-an event, and mark specific metadata that are required. Due to its rich metadata
-model, metadata can also be expressed in human language, making it easier for
-developers, data-analysts, or even business analysts to understand.
+consumers from other teams or even other organizations to find the information
+they need. A centralized registry gives teams one place to discover and query
+endpoints and the events they produce or consume. Extensions can describe these
+events in more detail, including which metadata are required. Descriptions can
```
duglin: "Extensions can describe these events in more detail, including which metadata are required" is worded such that it could be interpreted as "create an extension to indicate which fields are required". I think the 2nd part of the sentence is actually a separate topic from the first part.

core/primer.md
```diff
@@ -476,67 +457,44 @@ across these offerings.
...
-Further, building upon the idea of schema-based data contract generation, even
-client-code can be generated. When the schema is defined in a vendor-agnostic,
-well-defined format, tooling can be built upon this foundation to generate
-broker-specific consumers that can consume events from endpoints defined in the
-registry. This is not only limited to event consumers, but may also include
-generation of event store inserts. This can further speed up and streamline the
-development of both producers and consumers, significantly reducing development
-and testing time and minimizing the risk of introducing bugs at this level.
+context. Applications therefore often use serialization formats such as JSON
+that carry structural information in each payload. With that information
```
duglin: The "Applications therefore...." sentence is tripping me up. I'm reading it as "you can include 'structural info' in your message"... which implies "more data" in the message, but we're in a "payload reduction" section. Something feels off. Is this trying to say: "payloads can include references to schemas (stored in xReg) to allows the  payload to remain small while the structural definition is stored/available elsewhere" ?

core/primer.md
```diff
@@ -476,67 +457,44 @@ across these offerings.
...
+available from a centralized registry, applications can use lighter-weight
+serialization formats and reduce bandwidth without giving up contextual
+analysis.
+
+### 5.7. Schema-based contract and client generation
+
+The [file and API representations](#8-representations) can both supply versioned
```
duglin: This paragraph focuses on the "generate your code 'as needed' approach", but it seems to have dropped the "generate the code once and store it next to the model version so people can share the code w/o having to generate it themselves" option that the old text mentioned. Did you mean to drop that? I would think that's an important usecase because not everyone wants to deal with the tooling to generate the code - they just want to download a pre-built library.

core/primer.md
```diff
@@ -476,67 +457,44 @@ across these offerings.
...
+interact with endpoints instead of maintaining shared packages for those
+contracts by hand. A schema in a
+vendor-agnostic, well-defined format can drive tools that generate
+broker-specific consumers or event-store inserts. Generated producers and
+consumers reduce the amount of contract code that teams must write and test.
 Clemens Vasters built a proof-of-concept that shows this approach in
 the [Avrotize VS Code Extension](https://github.com/clemensv/avrotize).
 
-### 5.9. Basis for further developments
+<a id="59-basis-for-further-developments"></a>
```
duglin: ok this is weird. Does this mean that as section 5.8 moves around (e..g tomorrow it becomes 5.10 and then in 6 months it becomes 5.11) we'll need to add more `<a id=....>` tags for each AND EVERY possible toc number that this header ever lived at? And we need to do that for ALL headers for all time? This doesn't seem realistic. Wouldn't it be better to just remove the "59-" from the custom `<a>` tag?  And how do people grab this link? I thought there was going to be some kind of "link" icon thingy - and not just for ones that have moved, but for all.

core/primer.md
```diff
@@ -545,98 +503,46 @@ further builds upon this by providing a specification to define resources that
...
-Harmonize another piece of technology in the messaging / eventing ecosystem to
-increase interoperability and decouple event handling from broker products and
-protocols.
-
-Even though xRegistry was built with eventing in mind, the concept of this
-registry specification is far more powerful and can be used for the exchange for
-any type of message. Take a look at the [Possible Use
-Cases](#10-possible-use-cases) for examples outside the eventing world.
-
-## 7. Design Goals
```
duglin, line 587: I'm a bit surprised to see this section removed. Some of these might appear in the new section 5 (just w/o a "Goals" heading), which is fine, but I'm not sure I see all of these points below listed. For example, support for various protocols. We do mention the CE and schema-reg supports them but I didn't notice it being said about xReg.

core/primer.md
```diff
@@ -545,98 +503,46 @@ further builds upon this by providing a specification to define resources that
...
+instead of reconstructed from manually maintained documentation or observed
+runtime behavior.
+
+<a id="7-design-goals"></a>
+<a id="71-non-goals"></a>
+## 6. Non-Goals
 
 - **Authentication and Authorization:** Rely on established security standards
   depending on the registry representation.
 - **Relationships between event channels:** Focus on precisely describing a
```
duglin: we may want to make it clear this is an Endpoints spec issue, not a core xReg issue

core/primer.md
```diff
@@ -545,98 +503,46 @@ further builds upon this by providing a specification to define resources that
...
 resource metadata might also point to an external location where such data
 exists; also see the discussion about [embeddings and
 references](#9-embeddings-references-and-federation).
 
-When building up this spec, the API server had the highest priority, followed
-by the file representation and the static file server as an option to
-transition between the first two. To enable this transition, one of the
-design goals is the symmetry between all representations.
+The representations are symmetric so that data can move between them. Choose a
+representation according to the required write path and deployment model.
```
duglin: the phrase "required write path" feels awkward. Maybe "...according to the required data management and deployment models" ?? or s/model/needs/ ???

core/primer.md
```diff
@@ -645,25 +551,22 @@ these files can later be merged into another xRegistry when the module is
...
+This representation allows clients to retrieve individual entities instead of
+the entire registry as one JSON document. It can also serve data exported from
+an API server.
 
-Another benefit of this representation is that you only have to rent storage,
-but no compute units. You could spin up an xRegistry using GitHub Pages, just
-like a Helm registry. You save money and don't have to worry about maintaining a
-compute unit.
+It requires storage but no continuously running compute. For example, an
+xRegistry can be hosted on GitHub Pages, much like a Helm registry.
```
duglin: Perhaps we should point to one of our Github pages hosted xregistries as an example instead of Helm. hub.xregistry.io/xreg or xregistry.io/xreg

core/primer.md
```diff
@@ -716,20 +620,26 @@ each document, that provides the corresponding media-type. An example for this
...
 
 A registry server can only enforce validation constraints on embedded documents
 that are under its own control. If a document is external, meaning that the
 resource entry does not embed it but links to it, the document might change at
 the external location without the registry knowing.
 
-### 9.2. Embeddings and Links
+<a id="12-the-format-attribute-in-the-schema-spec"></a>
+The Schema specification requires `format` when `compatibility` is anything
+other than `none`, even when an implementation supports only one format. This
```
duglin: I believe we got rid of "none" and now it's just "absent".

core/primer.md
```diff
@@ -746,46 +656,42 @@ At present, all examples and implementations use JSON and so it might appear
...
 
-External links are very handy if you already have an existing registry of some
-kind, possibly just files in a public repo, and want to catalog those with
-xRegistry. Using the `<RESOURCE>url` patterns, all your documents can remain
-in their place. The `format` and `contenttype` information still needs to
-exist in xRegistry so that clients can understand this information before
-trying to resolve the document. When a resource is registered with a URL, the
-registry will not resolve it and load the content; that is left to clients.
-
-External links also make it possible to overlay xRegistry on top of a
```
duglin, line 768: I wonder if this usecase is still important to call-out

core/primer.md
```diff
@@ -746,46 +656,42 @@ At present, all examples and implementations use JSON and so it might appear
...
 
 A registry document, whether a single [file](#81-file) or the root of a
 [static file server](#82-static-file-server) layout, is meant to be loaded into
-an application as a local, in-memory registry. While a document technically
-needs to embed a model to be valid, servers often choose to allow that the
-models they know about are implicitly available, so that a registry document
-can just contain the resources and their metadata, without the model. There
-are several examples of such documents in the
+an application as a local, in-memory registry. A registry document used on its
+own must embed a model to be valid. A server may nevertheless accept source
```
duglin: use of "must" here is misleading given the next sentence talks about it being optional

core/primer.md
clemensv: The protocol-neutrality point got lost when the Goals section was consolidated. I added explicit text to section 5.8 stating that the core xRegistry resource model is independent of messaging protocols and that domain-specific specifications can describe resources associated with MQTT, HTTP, Kafka, AMQP, or other protocols without changing the common model.

core/primer.md
```diff
@@ -545,98 +542,46 @@ further builds upon this by providing a specification to define resources that
...
-- **Simplicity:** Allow simple use cases that do not need versioning of
-  resources or a full-blown schema registry on the server.
-
-### 7.1. Non-Goals
+Correlating events and endpoints across a system also provides the information
+needed to define event flows. Those flows can then be discovered directly
+instead of reconstructed from manually maintained documentation or observed
+runtime behavior.
+
+<a id="design-goals"></a>
```
duglin, line 576: old?

core/primer.md
```diff
@@ -545,98 +542,46 @@ further builds upon this by providing a specification to define resources that
...
+<a id="non-goals"></a>
```
duglin, line 577: I don't understand these. How is someone supposed to this "non-goals" URL? And why do they only appear for some headers and not all of them?

duglin: May need more testing, I think this might give you what you want:

```
# 6. <a id="docs" href="#docs">Docs</a>
```
It turns the text into a non-numbers href, but the numbered href is still available via the normal link-icon thingy I think

core/primer.md
```diff
@@ -1451,37 +1357,24 @@ that the use of this feature might not be useful at the Version-level or even
...
-tooling might reject such cases, or force uses to use non-problematic
-characters.
-
-## 14. Why do Resources have 3 levels of data?
+defined in the core specification for consistency. The Endpoint
+[specification](../endpoint/spec.md) does exactly this to indicate when an
+Endpoint (i.e. a Group) is deprecated.
+
+<a id="problematic-characters-in-values"></a>
+## 11. Problematic characters in values
```
duglin, line 1386: I think we should just move everything else from here under section 10. Not sure why this starts section 11

core/primer.md
```diff
@@ -545,98 +542,46 @@ further builds upon this by providing a specification to define resources that
...
+<a id="design-goals"></a>
+<a id="non-goals"></a>
+## 6. Non-Goals
 
 - **Authentication and Authorization:** Rely on established security standards
   depending on the registry representation.
-- **Relationships between event channels:** Focus on precisely describing a
-  single event channel before standardizing the relationships between event
-  channels.
+- **Relationships between event channels:** The Endpoint specification focuses
```
duglin, line 582: This one feels a bit out of place - I wonder if we should move it into the Endpoints spec so people don't miss it as they read that spec.

core/primer.md
```diff
@@ -545,98 +542,46 @@ further builds upon this by providing a specification to define resources that
...
-and then use its export function to generate files for the file or static file
-server representation.
+The representations are symmetric so that data can move between them. Choose a
+representation based on how the registry will be updated and deployed.
 
-### 8.1. File
+<a id="file"></a>
+### 7.1. File
 
 The registry is represented as a single JSON file.
```
duglin, line 602: Feels like this should be something more like "one or more files JSON files" - especially since 3 paragraphs down we specifically mention this as an option.

core/primer.md
```diff
@@ -672,26 +615,22 @@ filtering, you might consider using the API server instead. Adding server logic
 to the static file server to make up for the features of the API server is
 considered an anti pattern.
```
duglin, line 642: s/anti pattern/anti-pattern/

core/primer.md
duglin, line 663: xRegistry is extensible such that domain-specific catalog entries can be defined to allow for customized resources to meet a variety of needs.

Or something like that. The current sentence reads kind of weird to me.

core/primer.md
```diff
@@ -702,11 +641,16 @@ quite obviously not exist inside of it.
...
-documents is not declared by the model, but may be constrained by convention
-(and implementation choice) based on the [`format`](spec.md#format-attribute) attribute.
+xRegistry can embed content such as schema documents. The following sections
+separate document format metadata from the choice to embed, encode, or link the
+document.
 
-### 9.1. Formats and Content/Media-Types
+The model declares document metadata but does not define the document's
+internal shape. Implementations may interpret and validate that shape according
+to `format`.
```
duglin, line 676: s/according to `format`/through use of the `format` attribute/

core/primer.md
```diff
@@ -716,20 +660,26 @@ each document, that provides the corresponding media-type. An example for this
...
 might be "application/json", but the `format` is "Avro/1.12".
 
-Some xRegistry implementations, including the reference server implementation,
-might choose to provide validation capabilities for well-known formats like
-Apache Avro schema, and even version-to-version compatibility checks. These
-validation capabilities per-se are not defined in the specification, but the
-specification allows the server to provide indicators about whether they are
-available and whether they have been applied. Metadata declared in the models
-allows the server to validate embedded data directly.
+Some xRegistry implementations, including the reference server, validate
```
duglin, line 689: change "reference server" to be an href to the server repo

core/primer.md
```diff
@@ -716,20 +660,26 @@ each document, that provides the corresponding media-type. An example for this
...
+The Schema specification requires `format` when `compatibility` is present,
+even when an implementation supports only one format. This keeps the format
+explicit in exports so that another server, including one that supports several
+formats, can import them without ambiguity.
+
+<a id="embeddings-and-links"></a>
+### 8.2. Embeddings and Links
 
 When the resource is declared to be a document, it can be stored inside the
 registry either as a directly [embedded object](spec.md#resource-attribute), a
```
duglin, line 711: s/embedded object/embedded JSON object/

core/primer.md
duglin, line 720: When a document is embedded within the registry metadata, its encoding needs to align with the server's serialization format. In other words, if the server serializes the xRegistry metadata in JSON, then only documents that are valid JSON documents can be embedded "as is". In all other cases, when the document is stored within the registry, the `base64` attribute variable would be used to serialize the document.

core/primer.md
```diff
@@ -746,69 +696,66 @@ At present, all examples and implementations use JSON and so it might appear
 that this is a JSON-related feature, but it's very feasible to encode the entire
 registry in Avro binary and there is a formal schema for Avro.
 
+The core specification defines only JSON serialization. Another registry
+encoding requires a separately defined serialization binding.
+
 In JSON, we can embed JSON-data like an Avro Schema into `schema` and we can
```
duglin, line 728: In JSON, for the schema registry model, we can ...

core/primer.md
```diff
@@ -746,69 +696,66 @@ At present, all examples and implementations use JSON and so it might appear
...
 application can load the root document and traverse `<RESOURCE>url` links to
 individual documents on demand, building an in-memory view that combines the
 root document with only those branches that are relevant to the application. The
-same applies when the root sits behind an [API server](#83-api-server) or proxy:
+same applies when the root sits behind an [API server](#api-server) or proxy:
 the client fetches and caches just the branches it needs.
 
-#### 9.3.2. Shadowing
+<a id="shadowing"></a>
+#### 8.3.2. Shadowing
```
duglin, line 779: I'd prefer if we remove this until we get more clarity on it. I've read this several times and I'm not sure I fully understand what it's proposing or that two people would actually implement the same thing based just on what this says. Maybe a more concrete example would help - I dunno. 

I also am not sure if this is meant to be something a static file server can do (which it kind of implies) or requires running code - especially when I think about things like the "group constraints". How are those enforced? And if they fail, what happens? Are some entities discarded?

core/primer.md
```diff
@@ -826,7 +773,8 @@ composition and shadowing is the best approach to federation, but the project
 will need to gain more experience with it in practice before it can be
 formalized.
 
-#### 9.3.3. Identifiers
+<a id="identifiers"></a>
+#### 8.3.3. Identifiers
 
 For shadowing and cross-referencing to work, identifiers need to be stable
 across registries. Group identifiers should be treated as globally unique, so
```
duglin: "Group identifiers" being "globally unique" would seem to go against the idea of the same resource being available from different registries mentioned below (via a different host) since doesn't "globally unique" mean we can't reuse the same group identifier in more than one registry?

core/primer.md
```diff
@@ -177,7 +191,10 @@ defines (e.g. `endpoints`, `messages`) - they're not literal.
 
 #### 3.2.1. Reading things
 
-A plain `GET` on any entity returns its metadata as JSON:
+A plain `GET` on a Registry, Group, metadata-only Resource, or Version returns
+its metadata as JSON. Resources with domain-specific documents, called
```
Fire-furo, line 204: As per http.md, ```GET  .../versions/<VID>``` for Resources with domain-specific documents returns its document and ```GET  .../versions/<VID>$details``` or Resources without domain-specific documents returns metadata as JSON. So this line isn't accurate for Versions of document Resources , I think this should also be mentioned as an exception.

duglin: I think the wording is ambiguous: `meta-only Resources, or Versions` could be read as "meta-only" applies to Resources, or it applies to both Resource AND Versions. Perhaps we could write it as: `meta-only Resources/Version`, then the grouping of Resources and Versions is more clear. Would that help?

Fire-furo: Yea, ```meta-only Resource/Version``` is clearer and would help, but shouldn't it also be mentioned that Versions can return a domain-specific document?

duglin: We could do Resources/Versions in the next sentence too.

core/primer.md
```diff
@@ -826,7 +773,8 @@ composition and shadowing is the best approach to federation, but the project
 will need to gain more experience with it in practice before it can be
 formalized.
 
-#### 9.3.3. Identifiers
+<a id="identifiers"></a>
+#### 8.3.3. Identifiers
 
 For shadowing and cross-referencing to work, identifiers need to be stable
```
duglin: what do you mean by "identifiers" here? The various "ID" attributes or "self" or both? I'm asking because below you talk about them being unique (which isn't right for IDs) but then also talk about them being relative (which isn't right for "self").

core/primer.md
duglin: We can't really control what a client does, so from that perspective this paragraph is ok. However, at the same time the implication of this paragraph (and this entire section) seems to be that it's ok to dynamically build a registry based on lots of different data sources - and that's fine. However, when I think about what an implementation of these actually means it's much more than just things like "layering" of entities. These implementations needs to fully support the spec and all of that entails. For example, "constraints", "xref"s, proper "count" values, etc...  Not necessarily rocket science, but those are usually reserved for 'servers' and all of this logic is more like building a specialized persistent store (even if it's never actually stored on disk). Again, ok, but this section is written as if all of this logic can be trivially done by clients or a very thin layer that's more like a layered virtual filesystem - and that's where I think it's misleading people because I don't think you can do this by just "layering" or "merging" things. Logic is needed.

core/primer.md
```diff
@@ -871,24 +819,26 @@ specific use-case even though the original spec marks them as optional. In
...
-### 10.3. Metadata files in repositories
+<a id="metadata-files-in-repositories"></a>
+### 9.3. Metadata files in repositories
 
-Open source repositories could provide an xRegistry represented as a simple JSON
-file on the top level of the folder structure and list all metadata files this
-repository contains like a `package.json`. This would allow machine-readable
-access to project dependencies and configurations as well as consistent
-management of metadata across different tools and environments.
+The [file representation](#file) supports a repository-level manifest similar
```
duglin: The word 'supports' here comes across (to me) like there's a feature of "the single xreg json file" that supports a "manifest". I __think__ it's just saying that a single json file (export) of a registry can be viewed/managed as metadata for a project and stored in a git repo along-side code. RIght? If so, isn't this just repeating what's said in the "file representation" section?

core/primer.md
```diff
@@ -907,77 +857,43 @@ xRegistry model. As a result, implementations might want to avoid using `id`
...
-  - and layers of xRegistries to get multi-level grouping ??
-
-### 11.4. Group organization
+<a id="valid-characters"></a>
+### 10.2. Valid Characters
+
+Attribute names appear in JSON objects, HTTP headers, URLs, and generated code.
+The specification therefore uses a restricted character set that works across
+those contexts. Section [10.10](#naming-and-case-sensitivity) explains the
+casing rules. Section 10.11 explains the additional restrictions on Group and
```
duglin: turn 10.11 into an href

core/primer.md
```diff
@@ -907,77 +857,43 @@ xRegistry model. As a result, implementations might want to avoid using `id`
...
+<a id="extensions"></a>
+### 10.3. Extensions
+
+Extensions SHOULD be defined in the model. A model can also define a `*`
+extension to allow attributes supplied at runtime. Extension names follow the
+[valid character](#valid-characters) and
+[casing](#naming-and-case-sensitivity) rules for all attributes.
+
+Required aspects apply at each level of a nested extension. In particular,
+`clientrequired=true` also requires `serverrequired=true`; otherwise the model
```
duglin: these are not longer in the spec

core/primer.md
```diff
@@ -907,77 +857,43 @@ xRegistry model. As a result, implementations might want to avoid using `id`
...
-### 11.3. Extensions
-
-- it's RECOMMENDED that all extensions be defined in advance as part of the
-  model. However, the spec does support an extension of "\*" to allow for
-  unknown/runtime-provided extensions. Since extensions can appear in case-
-  insensitive situations (e.g. http header) we can't know the case of them
-  when storing in the backend. As a result, all attributes and key names MUST
-  be lowercase.
-
-- "default" Resource is just a pointer, NOT a set of default values
```
duglin, line 957: why did we drop this? I think we added it because some people thought the Resource was "default" values for the versions.

core/primer.md
```diff
@@ -907,77 +857,43 @@ xRegistry model. As a result, implementations might want to avoid using `id`
...
+Extensions SHOULD be defined in the model. A model can also define a `*`
+extension to allow attributes supplied at runtime. Extension names follow the
+[valid character](#valid-characters) and
+[casing](#naming-and-case-sensitivity) rules for all attributes.
+
+Required aspects apply at each level of a nested extension. In particular,
+`clientrequired=true` also requires `serverrequired=true`; otherwise the model
+is invalid.
+
+<a id="potential-extensions"></a>
```
duglin, line 910: No header?

core/primer.md
clemensv: Clarified the terminology and scope: `<SINGULAR>id` is parent-scoped, `xid` is Registry-scoped, and `self`/`shortself` are service-specific retrieval URLs.

core/primer.md
clemensv: Kept the useful composition examples and added the required caution: fetched or combined content must be processed under the effective model and xRegistry rules, including constraints, `xref`, and partial-view count semantics.

core/primer.md
clemensv: Removed the redundant metadata-files-in-repositories use case; the file representation section already covers it.

core/primer.md
clemensv: Linked section 10.11 using its stable explicit anchor.

core/primer.md
clemensv: Removed the obsolete `clientrequired`/`serverrequired` paragraph.

core/primer.md
clemensv: Restored the clarification that `defaultversionid` selects the default Version and that Resource attributes are not inherited defaults for Versions.

PR #562, opened 2026-08-25
Add text to primer describing the use of relative URLs 

### LINKED ISSUE #432: Allow (relative) links to resources instead of embedding the content
Especially for file based or static xRegistries, it sometimes feels strange to embed the content of documents into the registry. 

Example:
The realtime source provided by @clemensv contain one xreg file per source. For the GTFS source, there is an xreg file that contains all the Avro schemas. In addition, the schemas are also available as separate file.
[https://github.com/clemensv/real-time-sources/tree/main/gtfs/xreg](https://github.com/clemensv/real-time-sources/tree/main/gtfs/xreg
)

Wouldn't it be nice to just have links in the xRegistry file? This could look like the following:

```
        "GeneralTransitFeedRealTime.Trip.TripUpdate": {
          "schemaid": "GeneralTransitFeedRealTime.Trip.TripUpdate",
          "format": "Avro",
          "versions": {
            "1": {
              "schemaid": "GeneralTransitFeedRealTime.Trip.TripUpdate",
              "versionid": "1",
              "format": "Avro",
              "schemaUrl": "./gtfs-rt-tripupdate.avsc",
              "createdat": "2024-10-18T13:48:02.088029",
              "epoch": 0,
              "modifiedat": "2024-10-18T13:48:02.088029"
            }
          }
        }
```

In the case of a file-based xRegistry, the xreg file could play the role of an `index.html` file for all the metadata in a repo or project.

deissnerk:
Add section to primer describing the use of relative URLs in single file registries.

Fixes #432

### DIFF
--- a/core/primer.md
+++ b/core/primer.md
@@ -86,6 +86,7 @@ normative technical details.
   - [11.23. Detecting circular references between Versions](#1123-detecting-circular-references-between-versions)
   - [11.24. Optional required fields in requests](#1124-optional-required-fields-in-requests)
   - [11.25. Deprecation of entities in an xRegistry](#1125-deprecation-of-entities-in-an-xregistry)
+  - [11.26. Relative Resource URLs in the File representation](#1126-relative-resource-urls-in-the-file-representation)
 - [12. The `format` attribute in the Schema spec](#12-the-format-attribute-in-the-schema-spec)
 - [13. Problematic characters in attributes](#13-problematic-characters-in-attributes)
 - [14. Why do Resources have 3 levels of data?](#14-why-do-resources-have-3-levels-of-data)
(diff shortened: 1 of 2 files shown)


core/primer.md
```diff
@@ -1455,6 +1455,37 @@ defined in the core specification for consistency. It is worth noting that
...
+            schema2.json
+        ...
+```
+
+For a repository structured as shown above, a schema version could contain
+`"schemaurl": "../schemas/schema1.json"` to point to the schema document.
+The specification requires resource URLs to be resolvable which, in the
+context of the files of the repository, is possible. When uploading the
+registry content to a server though, the relative URLs must be resolved and
+the referenced documents included in the upload.
```
duglin: I think it would be good to show some of the contents of xreg.json to show what those relative links would look like.

Also, the last sentence may want to say that the "resolving" step is a client-side tooling issue, not a server one.

core/primer.md
```diff
@@ -1479,12 +1479,33 @@ Repository structure:
...
-the referenced documents included in the upload.
+For a repository structured as shown above, a schema version in xRegistry
+could look like:
+
+```yaml
+  "schemagroups": {
+      "MySchemaGroup": {
+          "schemas": {
+              "MySchema": {
+                  "schemaid": "MySchema",
```
duglin: I'd remove this since it's redundant

core/primer.md
```diff
@@ -1479,12 +1479,33 @@ Repository structure:
...
+                  "defaultversionid": "1",
```
duglin: if you want to include this then it needs to be under "meta"

core/primer.md
```diff
@@ -1479,12 +1479,33 @@ Repository structure:
...
+                  "versions": {
+                      "1": {
+                          "schemaid": "MySchema",
```
duglin: both schemaid and versionsid are optional (like I mentioned above), if you view this as input for an "import" action. If you want this to be a 100% valid xreg file like you'd get from an "export" then I agree with keeping them. What do you consider the purpose of this file? A user-friendly view of the data (good for "import") or a 100% spec compliant output of the data?

duglin:
approved on the 9/4 call

PR #563, opened 2026-08-26
Week of Aug 24

duglin:
- mention how to deal with `SINGULAR` doc updates in events (e.g. `schema` vs `schemabase64`)
- misc wording tweaks
- define registry.created/delete events
- changed "deprecation" event to "deprecated" and "undeprecated"


core/events.md
```diff
@@ -182,7 +182,11 @@ When present, and serialized in JSON, the `data` MUST be of the form:
 
 The `changed` attribute, when present, has the following constraints:
 - MUST be a list of the top-level attribute names (not values) that were
-  added, modified, or deleted for the `subject` entity.
+  added, modified, or deleted for the `subject` entity. When a Resource
+  has its `hasdocument` aspect set to `true`, the domain-specific document
+  is stored within the Registry, and it was modified, then the Resource's
+  `SINGULAR` attribute would be included in this list not `SINGULARbase64` -
```
lailabougria:
```suggestion
  `SINGULAR` attribute would be included in this list, not `SINGULARbase64` -
```

core/events.md
```diff
@@ -182,7 +182,11 @@ When present, and serialized in JSON, the `data` MUST be of the form:
 
 The `changed` attribute, when present, has the following constraints:
 - MUST be a list of the top-level attribute names (not values) that were
-  added, modified, or deleted for the `subject` entity.
+  added, modified, or deleted for the `subject` entity. When a Resource
+  has its `hasdocument` aspect set to `true`, the domain-specific document
+  is stored within the Registry, and it was modified, then the Resource's
```
lailabougria: I can't find a better way to phrase that to split this up in a list, but I found this really hard to understand.

This would be easier:
When:
- a Resource has its `hasdocument` aspect set to `true`,
- the domain-specific document is stored within the Registry, and,
- the Resouce was modified
then the Resource's `SINGULAR` attribute would be included in this list....

It may not be worth fixing, but just pointing out I had to read it a couple of times to make sense of it

core/events.md
```diff
@@ -223,6 +228,17 @@ value.
 
 ### `registry` Events
 
+- Action: `created`
+  - STRONGLY RECOMMENDED to be generated when a new Registry is created.
+    This event isn't mandated since the action of creating a new Registry
+    is not defined the the xRegistry specification, and therefore (technically)
```
lailabougria:
```suggestion
    is not defined in the xRegistry specification, and therefore (technically)
```

core/events.md
```diff
@@ -223,6 +228,17 @@ value.
...
+    out of scope. However, for interoperability, the suggested event is defined
+    here.
+
+  - If during the process of creating a new Registry the model, modelsource
```
lailabougria:
```suggestion
  - If, during the process of creating a new Registry, the model, modelsource
```

core/events.md
```diff
@@ -223,6 +228,17 @@ value.
...
+    or entities are modified/created then events for those actions MUST be
```
lailabougria:
```suggestion
    or entities are modified/created, then events for those actions MUST be
```

core/events.md
```diff
@@ -239,10 +255,17 @@ value.
...
     updated and the appropriate attribute MUST appear in `changed`, if present.
 
-The `io.xregistry.registry.created` and `io.xregistry.registry.deleted` events
-are not defined as part of this specification as those operations are not
-defined in the xRegistry specification. It is expected that the tooling that
-performs those operations will define those events, if needed.
+- Action: `deleted`
+  - STRONGLY RECOMMENDED to be generated when a Registry is deleted.
+    This event isn't mandated since the action of deleting a new Registry
+    is not defined the the xRegistry specification, and therefore (technically)
```
lailabougria:
```suggestion
    is not defined in the xRegistry specification, and therefore (technically)
```

core/events.md
duglin: done

core/events.md
duglin: Does this work ok?
```
- When a domain-specific document that is stored within the Registry is
  modified, then the Resource's `SINGULAR` attribute would be included in this 
  list, not `SINGULARbase64` - even if the `base64` variant is used during
  serialization of the Resource.
```

core/events.md
duglin: done

core/events.md
duglin: done

core/events.md
duglin: done

core/events.md
duglin: done

core/events.md
lailabougria: Much better

lailabougria (approved):

duglin:
Approved on the 8/27 call

core/events.md
```diff
@@ -176,38 +178,61 @@ When present, and serialized in JSON, the `data` MUST be of the form:
...
-- MUST be a list of the top-level attribute names (not values) that were
-  added, modified, or deleted for the `subject` entity.
+where:
+- The `epoch` attribute MUST be included in `created` and `updated` events
+  for Registries, Groups, Resources and Versions; and it MUST be the `subject`
+  entity's `epoch` value as seen at the end of the interaction.
+
+  - MUST NOT be included in `deleted` events.
+
+- The `meta.epoch` attribute MUST be include in Resource `created` and
```
lailabougria: *included

core/events.md
```diff
@@ -176,38 +178,61 @@ When present, and serialized in JSON, the `data` MUST be of the form:
...
+  `updated` events and it MUST be the `subject` Resource's `meta.epoch` value
+  as seen at the end of the interaction.
+
+  - MUST only be included for Resource related events.
+
+  - MUST NOT be included in Resource `deleted` events.
 
-- This attribute MUST NOT appear on the `created` or `deleted` events.
+  - Note that Resource `created` and `updated` events will include both
+    `epoch` and `meta.epoch` attributes even one of them didn't change for
```
lailabougria: even *if/when* one of them didn't change

core/events.md
```diff
@@ -176,38 +178,61 @@ When present, and serialized in JSON, the `data` MUST be of the form:
...
+    the interaction.
 
-- This specification does not mandate any particular order for the attribute
-  names in the array.
+- The `changed` attribute MAY be included to indicate which attribute of the
```
lailabougria: attributes

core/events.md
```diff
@@ -210,7 +211,8 @@ where:
   - MUST be a list of the top-level attribute names (not values) that were
     added, modified, or deleted for the `subject` entity.
 
-  - This attribute MUST NOT appear on the `created` or `deleted` events.
+  - This attribute MUST NOT appear on the `created`, `deleted`, `deprecated,
```
lailabougria: > `deprecated`
missing quote

core/events.md
```diff
@@ -352,21 +354,25 @@ events.
...
-    attribute names.
+    `changed`, if present, MUST include `epoch`, `modifiedat` and
+    `deprecated` attribute names.
+
+- Action: `undeprecated`
+  - MUST be generated when a Group's `deprecated` sub-object is deleted.
+
+  - An `io.xregistry.group.updated` event MUST also be generated where
+    `changed`, if present, MUST include `epoch`, `modifiedat` and
+    `deprecated` attribute names.
```
lailabougria, line 375: Are we forcing servers to support undeprecated? I could see how a server could choose not to implement it, but I understand we want to support it in the spec.

duglin: If we want undeprecated to be optional to support then that's a core spec issue, not an eventing one. I'll open an issue so we can discuss whether the core spec should allow for it. I actually think it might already since just about all operations are optional. I think the real issue is whether people can advertise whether it's supported or not... we don't have a mechanism to specify which ops are not supported. Could get kind of verbose.

duglin:
approved off-line

PR #571, opened 2026-09-10
Add registry federation specifications and transport bindings working drafts

marcschier:
## Proposed Changes

This PR proposes working-draft specifications for **read-only xRegistry
federation**. A Registry combines local Resources with cataloged sources
using HTTP, OCI, Git, File or OPC UA. Views identify whether the consumer or
producer performs resolution; absence of the signal defaults to the consumer.
Producers can serve a combined view on demand or publish it as stored files.
The drafts also define OCI packaging and a directory mapping that preserves
existing file layouts.

## Specifications

Start with **Registry of Registries** and **Federation**, then use the [binding comparison](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/README.md#choosing-a-binding) to choose the relevant binding.

| Specification | Description |
| --- | --- |
| [Registry of Registries](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/models/registry/spec.md) | Catalog entries describing Registries, access methods and relationships. |
| [Federation](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/federation/spec.md) | Producer/consumer resolution, local shadows and source selection, with a worked example of label selection after shadowing. |
| [Directory Mapping](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/bindings/mapping.md) | Map existing files to a Registry. Includes a complete one-file walkthrough with every metadata file, matching hashes and expected read results. |
| [OCI Binding](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/bindings/oci.md) | Package and read Registry snapshots through OCI Distribution, with artifact-content and resolution-sequence diagrams. |
| [HTTP Federation](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/bindings/http-federation.md) | Access source Registries through the existing HTTP API, including pagination and live-read consistency. |
| [Git Binding](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/bindings/git.md) | Read directory mappings from a pinned Git commit using stored object bytes. |
| [File Binding](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/bindings/file.md) | Read mappings or OCI layouts on local storage or OS-mounted file shares, with containment and integrity checks. |
| [OPC UA Binding](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/bindings/opcua.md) | Native Registry access, reference resolution and document transfer over OPC UA. |
| [OpenUSD Alignment](https://github.com/marcschier/xregistry-drafts/blob/main/workingdrafts/models/openusd/spec.md#53-federation) | Clarify the distinction between local aliases and remote document links in the existing draft. |

Models, schemas, samples and tests are **supporting material**, not the
starting point for review. The complete examples are readable in the specs
without running the test suite. These remain unreleased working drafts.

Schema-generator fixes and released-schema artifacts are handled separately
in xregistry/spec#574. The release-schema CI check still depends on
reconciling that work.

**Release Note**

```release-note
Added Registry-of-Registries and federation working drafts, native access
bindings, resolution-owner signaling, OCI packaging, directory mapping and
complete worked examples.
```


core/primer.md
```diff
@@ -829,12 +829,13 @@ formalized.
 
 #### 9.3.3. Identifiers
 
-For shadowing and cross-referencing to work, identifiers need to be stable
-across registries. Group identifiers should be treated as globally unique, so
-that the same group can be recognized across different registries. Resource
-identifiers are only unique within a group, so a cross-registry reference should
-always use the combination of group identifier (group name and id) and resource
-identifier (resource name and id).
+For shadowing to recognize corresponding entities, participants need an agreed
```
marcschier: While working draft, it might not make sense to add links/changes to non-working draft documents, - remove this, but we might want an issue filed to add this _eventually_

cloudevents/schemas/document-schema.avsc
marcschier, line 1: Undo the changes under cloudevents/schemas.  Create a separate PR IFF these are bugs, then create clear PR description and an issue describing the bug

core/primer.md
marcschier: Addressed in 178042c86cb7e2d78958a411c199b8568c704807. Reverted all changes to core/primer.md, including the working-draft link, so it is identical to this PR's merge base. Filed xregistry/spec#572 to track the identity clarification and links after the federation drafts stabilize. No released-primer changes remain in this PR.

cloudevents/schemas/document-schema.avsc
marcschier: Addressed in 178042c86cb7e2d78958a411c199b8568c704807. Restored every file under cloudevents/schemas to the PR merge base; that directory now has no PR diff. The generator defects are independently reproducible on upstream a154439 (8 of 9 new regression cases fail), so I filed xregistry/spec#573 and opened the separate upstream-based draft PR xregistry/spec#574 with fix commit cb1476cb5a36894b9824277b9e1cd6eaa088af3a. Its complete tools suite passes 76 tests and all 16 affected generated artifacts reproduce byte-for-byte. This draft passes 1,097 tests; its full release-schema regeneration/cleanliness check remains dependent on reconciling the separate generator fix, as now stated in the PR description. I have not reintroduced the reverted CloudEvents artifacts to conceal that dependency.

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
+# xRegistry Federation
+
+<!-- words: applicationuri federationprofiles nfc nodeid nodeids opc opcua rebasing registrygroups registryroot reparse standalone symlinks transportprofileuri ua website weburl xregurl -->
+
+## Abstract
+
+This specification defines read-only resolution across independently
+administered xRegistries. A Registry of Registries describes access methods;
+a resolver chooses a method, establishes a Registry context and interprets
+that Registry using xRegistry Core. Protocol bindings define retrieval, not a
```
marcschier: Remove ", not a new ... cross-references"

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+new identity system or changes to Core cross-references.
+
+**Status:** Unreleased working draft. This specification and its companion
+bindings are not part of a released xRegistry specification.
+
```
marcschier: Remove

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+- [Snapshots and Completeness](#snapshots-and-completeness)
+- [Errors](#errors)
+- [Security Considerations](#security-considerations)
+- [Conformance](#conformance)
+- [Source Reconciliation](#source-reconciliation)
+- [References](#references)
+
+## Overview
+
+Federation permits a consumer to find and read entities through a catalog
```
marcschier: replace "catalog" with "registry", because from a consumer perspective it does not have to be a catalog, its access point is the xregistry based "registry".  This is important because there might already be resources (shadow resources) in the registry, which are served, without retrieving from the federated registry (as described in the catalog).  The view for a consumer is the registry, the catalog is a means for a registry to federate with subservient registries to service the user request

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+without requiring all content to be copied into that catalog. The
+[Registry domain](../models/registry/spec.md) retains the deployed hub's
```
marcschier: Replace "retains the deployed hub's" with e.g. "contains"

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+[HTTP](../bindings/http-federation.md), [OCI](../bindings/oci.md),
+[Git](../bindings/git.md), [File](../bindings/file.md) and
+[OPC UA](../bindings/opcua.md) define how to perform those reads. The
+[document-tree format](document-format.md) is shared by Git and File.
+Native bindings do not require an HTTP facade.
+
+This specification does not define synchronization, replication, write-through,
+global entity identifiers, arbitrary merge precedence or discovery of
+dependencies embedded inside domain documents. OCI publication is a separate
+producer operation; describing it does not grant a reader publication rights.
```
marcschier: Not sure what is meant with "; describing it does not grant a reader publication rights", remove.

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+| Snapshot | A captured Registry state, independent of earlier snapshots. |
+| Linked snapshot | A complete internal package graph allowing explicit external document references. |
+| Offline-complete snapshot | A snapshot also carrying all declared xRegistry documents in its selected scope. |
+
+The catalog's `registryid`, its entry's `registryid` and the described
+Registry root's `registryid` have different scopes. Their values need not
+match. A catalog-description `versionid`, target Resource `versionid`, OCI
+digest, Git commit and endpoint address MUST NOT be treated as interchangeable.
+
+## Discovery and Binding Selection
```
marcschier, line 460: Jumps to discovery and resolution, but this needs some introduction first (motivation, use case, what a compliant registry does upon requests (resolving what, etc.)), ideally with a 1. a topology diagram 2. a resolution sequence diagram (in this section), both as mermaid.

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+Endpoints MUST be absolute URIs and MUST NOT embed credentials. A selected
+built-in advertisement with an unknown parameter MUST produce
+`unsupported_operation`; silently ignoring a parameter could select a
+different Registry or revision. An unknown profile name remains valid catalog
+metadata, but MUST NOT be executed as a known profile.
+
+An absent `priority` means zero; lower unsigned integer values are preferred.
+An absent `parameters` means an empty object. Selection MUST proceed as follows:
+
+1. Validate the catalog's advertised legacy/explicit HTTP consistency as
```
marcschier: Remove the use of "legacy" everywhere when referring to xregurl

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+   specified by the Registry domain.
+2. Apply the caller's explicit profile choice and access policy.
+3. Remove unsupported profile names from consideration.
+4. Order remaining advertisements by ascending priority, preserving array
+   order for ties.
+5. Choose the first remaining advertisement and validate its parameters.
```
marcschier, line 500: What happens if there are multiple candidates?  How are they ordered and what considers "found"?

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+| `capabilities` | `/` | Capabilities of the selected access method. |
+| `collection` | Typed collection path | Collection members, with explicit completeness or continuation information. |
+| `entity` | Registry, Group, Resource, Meta or Version XID | xRegistry metadata, not raw domain content. |
+| `document` | Resource or Version XID | Domain bytes, or an explicit failure or external-document locator. |
+
+A Resource document request selects its Core default Version. An explicit
+Version request selects exactly that Version, case-sensitively. Neither
+operation selects the greatest Version string, latest OCI tag, newest Git
+commit or most recently modified file. Metadata-only Resources have no domain
+document; requesting one MUST produce `unsupported_operation`.
```
marcschier: Rewrite all semi-colon delimited sentences into real sentences seperated by fullstop/period (".") across the entire PR

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+
+| Term | Meaning |
+| --- | --- |
+| Catalog | A Registry using the Registry domain to describe other Registries. |
+| Catalog entry | A `registry` Resource under a `category` Group. |
+| Advertisement | One access method in a catalog-description Version. |
+| Registry context | The selected logical Registry, model and access context in which an XID is interpreted. |
+| Revision pin | A binding's immutable snapshot selection, when supported. |
+| Origin | The catalog entry, advertisement and revision used for a result. |
+| Resolver | A consumer implementing this specification and a named binding. |
```
marcschier: There needs to be a section re how the registry is hosted and who does the resolution.  E.g. the normal case is that a consumer "resolves" using the catalog, but there shall be an API server implementation that does the resolution for the consumer calling standard HTTP or other API binding endpoints and the API server hosting the xregistries does the resolution for them.  Structure this the same as the section in core.md that describes the different hosting models.  Add it early before ##Discovery and binding section

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+7. Return the result with its origin and revision context, or a specific
+   failure.
+
+Origin information MUST identify the chosen advertisement, catalog-description
+Version when applicable, target XID and revision pin or the absence of an
+immutable pin. It MUST travel separately from Core identity attributes.
+A standalone endpoint supplied by the caller MAY bypass catalog discovery,
+but MUST still establish all other context.
+
+If a model is captured, the package MUST preserve enough model source to
```
marcschier, line 563: Define or replace the term "package".

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+retain Resource type sharing. An expanded model alone can obscure
+`ximportresources` provenance. Included model documents MUST be resolved
+within the captured scope, or supplied with immutable checked references;
+offline-complete interpretation MUST NOT fetch mutable external includes.
+Unknown Core versions MUST NOT be accepted merely because a document parses.
+
+## Identity and References
+
+An XID identifies an entity only within one Registry. Resolvers MUST preserve
```
marcschier, line 572: This needs more info. E.g. a resolver resolves XIDs, if they are not found a resolver tries to resolve the same xid in the catalog registries, and serves what is found, is that correct?  If so, this needs to be made clear, if not, explain.

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+both its Registry context and typed collection path. Equal XIDs in unrelated
+Registries are not evidence that the entities or their bytes are identical.
+Core IDs are unique case-insensitively within their parent, but are looked up
+case-sensitively. Storage bindings MUST NOT weaken either rule.
+
+### Local Cross-References
+
+Core [`meta.xref`](../../core/spec.md#cross-referencing-resources) remains
+an intra-Registry Resource XID. It MUST NOT contain an endpoint URL, OCI
```
marcschier: "It MUST NOT contain an endpoint URL, OCI locator or remote Registry selector."  

Why would it?

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+locator or remote Registry selector. Its source and target MUST have the
+same Resource model type; structurally identical independent definitions
+are insufficient. Sharing between Group types uses
+[`ximportresources`](../../core/model.md#reuse-of-resource-definitions).
+
+For example, this model source permits a local alias from a mirror Group
```
marcschier: Link to place where it is defined or define here alias and mirror group concepts

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+conform separately to [Core HTTP](../../core/http.md). Neither `oci://`
+locators nor UA NodeIds become Core XIDs.
+
+Formats MUST preserve extension attributes and domain bytes. A zero-byte
+document is present content and MUST NOT be replaced by `{}`, `null` or a
+missing document. Binary documents MUST NOT pass through text conversion.
+
+## Snapshots and Completeness
+
+Every snapshot MUST have a complete internal containment graph. A linked
```
marcschier, line 745: Introduces snapshot without a link or explanation, fix!

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+| `#` JSON Pointer navigation | The returned JSON document. |
+
+A resolver MUST NOT substitute one of these bases for another. A binding
+that cannot preserve or supply the base needed to interpret a relative
+document URI MUST report `invalid_package` or `unsupported_operation`
+rather than guess.
+
+## Selectors
+
+A label selector is scoped to one named collection. Its JSON illustration,
```
marcschier: Introduces selectors without a link or explanation, fix!

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,449 @@
...
+| Catalog: `registrygroups` and localized mandatory labels | Replaced by the deployed `categories` hierarchy and ordinary OPTIONAL Core labels. |
+| Catalog: endpoints, model descriptions and relationships | Additive Registry domain Version attributes. |
+| Profiles: remote reference resolution | Explicit Registry context, not a widening of Core `xref`. |
+| Profiles: selectors and common failures | Selectors and errors in this specification. |
+| OCI: packaging and selective retrieval | Bounded standard descriptor graph in the OCI binding. |
+| Git/File: independently shaped directories | One shared document-tree format with separate transport selection. |
+| HTTP: transport retrieval | Existing Core HTTP API, with federation selection and consistency rules. |
+| OPC UA: native access | Existing draft binding, with corrected addressing and serialization. |
+| Synchronization, writes and inferred payload dependencies | Outside this read-only federation specification. |
+
```
marcschier: Remove

workingdrafts/federation/document-format.md
```diff
@@ -0,0 +1,436 @@
...
+<!-- words: modelbase resolvedmodelsource sha256 symlinks reparse xids -->
+<!-- words: modelsource hasdocument defaultversionid defaultversionsticky -->
+<!-- words: ancestorid isdefault formatversion href readonly ximportresources -->
+<!-- words: casefold subtrees filesystem filenames bytecode smudge -->
+<!-- words: federationerror namespace namespaces py reserialized -->
+<!-- words: standalone workingdrafts -->
+<!-- words: gitattributes checkout -->
+<!-- words: formatvalidated compatibilityvalidated -->
+
+## Abstract
```
marcschier: Move this and related docs to "bindings" folder (side by side with git/file)

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,436 @@
...
+
+- [Entity record schema](schemas/document-record.schema.json).
+- [Collection index schema](schemas/document-index.schema.json).
+
+Both schemas use JSON Schema Draft 2020-12. Their URNs identify this draft's
+schemas; they are not network retrieval endpoints. Producers and consumers
+MUST also implement the semantic constraints below. Schema validity alone
+does not establish Core model compliance or graph completeness.
+
+## Root and Storage Paths
```
marcschier, line 442: Before this we need an introduction with how this fits into the other documents and why it exists (motivation) and how it is supposed to be used, ideally with a very short "example". It must be delineated from the already defined "file" serialization of an xregistry in core.md.

workingdrafts/federation/document-format.md
```diff
@@ -0,0 +1,436 @@
...
+| --- | --- |
+| Git proposal, sections 2, 3 and 5 | Indexed tree and commit capture retained; type-less paths replaced by typed XIDs and opaque allocations. |
+| File proposal, sections 2, 3 and 5 | One selected `registry.json` root; filesystem containment and explicit layout selection. |
+| Both proposals, section 4 | Ordinary OPTIONAL Core labels and case-insensitive values replace mandatory language keys and exact localized matching. |
+| Both proposals, section 6 | Read-only access and distinct selection failures; concrete schemas and exact-byte rules added. |
+| Core document view and model reuse | Navigation is materialized, defaults are not duplicated, and `xref` is not remote federation. |
+
+The preliminary proposals are informative design inputs. Local Core and the
+approved shared federation contract take precedence over their examples.
+
```
marcschier: Remove

workingdrafts/bindings/git.md
```diff
@@ -0,0 +1,230 @@
...
+
+| Git proposal clause | Disposition |
+| --- | --- |
+| Section 1, `git+https` and URI fragment | Replaced by HTTPS endpoint plus REQUIRED structured `revision`. |
+| Sections 2 and 3, indexed commit snapshot | Retained, with REQUIRED pinning and object-only access. |
+| Section 4, exact localized labels | Replaced by shared Core-compatible literal label selection. |
+| Section 5, example paths | Replaced by the shared typed index and collision-free filename allocation. |
+| Section 6, read-only conformance | Retained with explicit indirection, integrity and consistency failures. |
+
+- [Shared federation](../federation/spec.md).
```
marcschier, line 248: Add a ## References header

workingdrafts/bindings/git.md
```diff
@@ -0,0 +1,230 @@
...
+
+## Source Crosswalk and References
+
+| Git proposal clause | Disposition |
+| --- | --- |
+| Section 1, `git+https` and URI fragment | Replaced by HTTPS endpoint plus REQUIRED structured `revision`. |
+| Sections 2 and 3, indexed commit snapshot | Retained, with REQUIRED pinning and object-only access. |
+| Section 4, exact localized labels | Replaced by shared Core-compatible literal label selection. |
+| Section 5, example paths | Replaced by the shared typed index and collision-free filename allocation. |
+| Section 6, read-only conformance | Retained with explicit indirection, integrity and consistency failures. |
```
marcschier: Remove - remove all "crosswalks" but leave references in the specs in this PR

workingdrafts/bindings/git.md
```diff
@@ -0,0 +1,230 @@
...
+## Conformance
+
+A producer MUST publish a complete shared document-tree snapshot in one
+commit at the advertised path. A native resolver MUST implement locator
+validation, commit pinning, object-only reads, the common operations and
+the rejection rules above. Offline-complete conformance additionally requires
+all declared document bytes locally in that snapshot.
+
+The shared [offline helper](../../tools/document_examples.py) accepts an
+explicit local Git object directory and a revision. It performs no network
```
marcschier: Remove ". It performs no network
acquisition and is not a production Git client"

workingdrafts/bindings/file.md
```diff
@@ -0,0 +1,228 @@
...
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
+"SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this
+document are to be interpreted as described in
+[RFC2119](https://www.rfc-editor.org/rfc/rfc2119).
+
+The [federation contract](../federation/spec.md), Core 1.0-rc4 and
+Model 1.0-rc4 define the read semantics. This binding does not define
+filesystem mutation, synchronization, a file watcher or an HTTP facade.
+
+## Advertisement and File URI
```
marcschier, line 71: For all bindings here, add a motivational example and introduction.

schema/schemas/document-schema.avsc
marcschier, line 1: Move endpoint/schemas, message/schemas and schema/schemas changes also into #574, and out of this PR

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed the trailing 'not a new identity system or changes to Core cross-references' wording from the Abstract. The sentence now states that protocol bindings define retrieval. Leaving the thread unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed the Status paragraph from the federation specification. Its placement under workingdrafts and the directory's status notice remain unchanged. Leaving the thread unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Reframed the consumer-facing access point as a Registry, with the catalog serving the resolver. Added local Resource shadow precedence before remote source lookup. A found local Resource owns its entire Version set, so a missing Version is not borrowed from another origin. The new source-selection examples test that remote reads are not made for local shadows. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. The overview now says the Registry domain 'contains' the categories/registries hierarchy and explains how it describes source Registries. Removed the deployed-hub justification from that paragraph. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed the publication-rights disclaimer. The scope now simply distinguishes OCI publication as a separate producer operation. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Added an organizational local/supplier use case, hosting motivation, a Mermaid topology diagram and a Mermaid resolution sequence before Discovery and Binding Selection. The sequence distinguishes local hits, ordered source misses, a selected Resource's Version/document read and terminal errors. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed 'legacy' descriptions of xregurl throughout the draft prose, model/sample descriptions and related helper/test terminology. It is described as the HTTP endpoint or implicit HTTP advertisement. Its priority-zero, append-after-explicit selection behavior is unchanged. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Separated two candidate sets: source Registries follow an explicitly configured composition order, while advertisements for one source use ascending priority and original array order. Defined 'found' as the exact Resource existing with interpretable metadata under the selected context/model policy. Only Resource not_found advances to the next source. A reachable endpoint, catalog entry or empty collection does not count as found. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Replaced semicolon-separated prose sentences throughout the changed specifications and READMEs with periods, with capitalization/wording adjustments. Audited changed Markdown for remaining prose semicolons. Kept semicolons that are data or syntax in fenced/inline examples, including UA NodeIds, HTTP Link headers and media types. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Added Design: Hosting Models before discovery, using the Core views/no-code-server structure. It covers client-side resolution, a federating API server resolving ordinary binding requests for consumers, and preassembled no-code/stored views. Server-facing models, capabilities, identity/navigation and source provenance obligations are explicit. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Defined Package as a stored representation of snapshot metadata and declared documents in the document-tree or OCI format. The snapshot section explains the relationship between package, captured Registry state and Resource Version, with links to both concrete formats. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Yes, under an explicit composition policy the resolver checks the local Resource and then the same typed Resource XID in ordered configured sources. It stops at the first found Resource and uses only that origin for all of its Versions/documents. It does not search arbitrary catalog entries or infer global identity from an equal XID. Added the complete algorithm and tests for source misses, local shadows, missing Versions and non-not_found failures. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed the redundant list of forbidden endpoint/OCI/remote-selector values. The text instead explains the distinction: Core xref supplies an intra-Registry alias, while the catalog/composition policy chooses source Registries. The existing Core rule is linked rather than restated as a new federation restriction. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Defined Alias in the terminology table with a direct Core cross-reference link. The example now explicitly says mirrors is just a Group-type name importing assets, not a standard mirror-Group or replication feature. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Added linked Snapshot terminology, introduced captured views in the hosting section, linked the representation section's use of snapshot, and expanded Snapshots and Completeness with motivation and its distinction from a Resource Version. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Added Selector to the glossary, introduced it with the request-resolution model, and expanded Selectors to explain collection criteria and link Core filtering. The example and later-page uniqueness rules remain. Leaving unresolved as requested.

workingdrafts/federation/spec.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed Source Reconciliation and its table-of-contents entry. Kept the specification's References section. Leaving unresolved as requested.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Added Motivation and Use before Root and Storage Paths, with a short directory example and links to the complete fixture. It explicitly distinguishes Core single-/multiple-document views and no-code HTTP files from the indexed storage allocation and digest references used by Git/File. The format, schemas and exact-byte fixture now sit alongside the bindings, with all helper/test/navigation paths updated. Leaving unresolved as requested.

workingdrafts/federation/document-format.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed the document-format Source Crosswalk, the proposal-precedence paragraph and the corresponding TOC entry. Preserved the normative References section. Leaving unresolved as requested.

workingdrafts/bindings/git.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Git now has a standalone References heading and matching TOC link. File uses the same structure. Leaving unresolved as requested.

workingdrafts/bindings/git.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed source crosswalk/reconciliation sections across federation, document-tree, Git, File and OCI, including TOC entries and private-input provenance discussion. Retained the normative references and corrected the OCI References section numbering. Leaving unresolved as requested.

workingdrafts/bindings/git.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Removed the specified no-network/not-a-production-client sentence from the Git helper paragraph. It now introduces the local object-directory/revision inputs directly before the example command. Leaving unresolved as requested.

workingdrafts/bindings/file.md
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Added motivational introductions and concrete examples for File, Git, OCI, HTTP federation and OPC UA. Each explains its use case, source access and relationship to consumer-side or API-server resolution. Added a separate motivation section to the shared document-tree format as well. Leaving unresolved as requested.

schema/schemas/document-schema.avsc
marcschier: Addressed in 321b4de6b931bbfe2e3da04baa896d8727d54980. Restored endpoint/schemas, message/schemas and schema/schemas to the PR merge base, alongside the already restored cloudevents/schemas. None of the 16 released schema artifacts now has a diff in this draft PR. The regenerated artifacts and latest generator review fixes are in xregistry/spec#574 at 36af874b748665262fcbbfb0fed82a0866fcf235, whose local suite passes 80 tests. The draft's documented regeneration dependency on that PR remains explicit. Leaving unresolved as requested.

workingdrafts/bindings/mapping.md
marcschier, line 1: Rename file to something shorter, e.g. linking.md (but research whether there are better names)

workingdrafts/bindings/document-format.md
```diff
@@ -0,0 +1,465 @@
...
+The document tree adds a concrete storage allocation, complete typed
+collection indexes and exact size/digest references. These permit selective
+reads and integrity checks without deriving filesystem names from Core IDs.
+Its stored records are not Core API responses. The reader assembles them
+into the Core document view defined below.
+
+For example, a Registry with one asset can be captured as:
+
+```text
+snapshot/
```
marcschier: What I would like here and how it should work is that you can make any file system or folder a registry by dropping in registry.json and other files here, with final versions/resources pointing to a file anywhere in the scope, i.e. the folder structure is not required to be complying, the .json files added (with registry.json at the root) where required allow another registry to access it, federate it, and serve it as an xregistry.  The documentation is not clear about it but shows a folder structure that implies all must be structured like this.  Either clarify or rewrite to support above.

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,465 @@
...
+shows a valid tree.
+
+Git chooses the directory through `parameters.path` and pins its commit.
+File selects that same directory through a file URI. The shared
+[hosting and resolution model](../federation/spec.md#design-hosting-models)
+determines whether the consumer or an API server performs these reads.
+
+## Root and Storage Paths
+
+The selected directory MUST contain `registry.json`. Its record MUST have
```
marcschier, line 444: a file starting with "." and without file extension might be more expected in the context of git, consider pros and cons, do not change anything yet.

workingdrafts/bindings/document-format.md
marcschier: Addressed in ea81603d0d360bd16e8056d3076a22e5bc9bb062. Rewrote the format as a mapping added to an existing directory. `registry.json` stays at the root. Other metadata can be placed beside existing files or in a separate metadata directory. Version document references can point to existing files anywhere within the selected root, without imposing `records/`, `indexes/`, `documents/` or file-extension conventions.

Updated the schema and both File/Git readers, not just the prose. The new tests add only metadata to a project containing unchanged JSON, binary and empty files, read the same mapping from Git objects, preserve Unicode names, allow multiple Versions to reference the same file, and retain hash/size and containment checks. Unreferenced project files remain outside the Registry. Three regressions failed before the fix. The complete suite now passes 1,149 tests. The previous allocated layout remains a valid example, not a requirement.

workingdrafts/bindings/mapping.md
marcschier: Considered the tradeoffs and kept `registry.json` unchanged in ea81603d0d360bd16e8056d3076a22e5bc9bb062, as requested. Added a regression confirming that an extensionless `.xregistry` file does not silently replace the existing entry point.

A name such as `.xregistry` marks metadata clearly and reduces visible project clutter. However, it is hidden by common Unix listing conventions, can be missed by some discovery/glob workflows, and loses the `.json` cue used by editors and some content-type tooling. Git itself does not automatically ignore dotfiles: [ignore rules determine what is ignored](https://git-scm.com/docs/gitignore). Extensionless JSON can work, but [editor language/schema associations](https://code.visualstudio.com/docs/languages/json) need explicit handling. Retaining `registry.json` also avoids introducing fallback order and compatibility rules for two root names.

Recommendation: keep the current entry point for now and make any later rename an explicit format/discovery decision. No dotted entry-point name or alias has been introduced.

workingdrafts/bindings/mapping.md
marcschier: Addressed in ea81603d0d360bd16e8056d3076a22e5bc9bb062. Renamed the document to `workingdrafts/bindings/mapping.md`, titled **xRegistry Directory Mapping Format**, and updated all links.

I compared `linking.md`, `directory.md`, `manifest.md` and `mapping.md`. `mapping.md` best distinguishes the metadata-to-existing-file mapping from the `file.md` transport, Core `xref` linking and OCI manifests. This is a naming recommendation, not a name mandated by another standard. [RO-Crate 1.2's root-metadata/existing-folder model](https://www.researchobject.org/ro-crate/specification/1.2/structure.html) provides relevant precedent for the structure. `registry.json` and the serialized format identifier remain unchanged.

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+
+Each Version record MUST have exactly one `document` descriptor:
+
+| `document.kind` | REQUIRED fields | Meaning |
+| --- | --- | --- |
+| `local` | `href`, `size`, `sha256` | Exact bytes in an existing or newly stored file inside the selected root. |
+| `external` | `uri` | Explicit external domain-document reference. |
+| `none` | No others | The Resource model has `hasdocument: false`. |
+
+For `local`, `base` and `origin` are OPTIONAL absolute, credential-free URIs.
```
marcschier: Review sentence structure

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+`<RESOURCE>url` on the now-local Version.
+
+For `external`, `uri` MUST equal the Version's `<RESOURCE>url` exactly.
+A relative URI MUST have an explicit absolute `base` on this descriptor.
+Neither the containing file, catalog URL, Git repository URL nor `source.uri`
+is an implicit base. Known integrity is represented by OPTIONAL paired
+`size` and `sha256` fields. Consumers MUST check both if they retrieve content
+with these fields. An external locator is not a metadata alias.
+
+An ordinary document-bearing Version MUST have local bytes or an explicit
```
marcschier: "document-bearing" -> Version referencing a document

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+external locator. A missing document is an invalid package, not empty content.
+A zero-byte local document MUST reference an actual zero-byte file, with
+`size: 0` and SHA-256
+`e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`.
+Binary and JSON bytes MUST NOT be decoded, reserialized, newline-converted,
+smudged or replaced by an empty JSON wrapper. The Version's `contenttype`,
```
marcschier: Remove "smuged"

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+Core document attributes (`<RESOURCE>`, `<RESOURCE>base64`, `<RESOURCE>url`)
+MUST be absent. A document request MUST produce `unsupported_operation`.
+This is distinct from a present zero-byte document.
+
+## Reads and Core Document View
+
+An exact read walks the Registry record, the necessary typed collection
+indexes and selected child records. A Version read MUST NOT require reading
+unrelated domain bytes. A consumer MAY scan a collection index in memory.
+This version defines one complete index per collection and no pagination.
```
marcschier: This -> The, unless another "version" is meant here?

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+A size or traversal limit MUST fail explicitly, not act as a partial index.
```
marcschier: What does this mean? "A size or traversal limit MUST fail explicitly, not act as a partial index."

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+contain external-only domain documents. Captured resolved model material MUST
+permit offline interpretation in both classes.
+
+Here `scope: "/"` denotes the modeled Registry, not every file physically
+present in the selected directory. Only files reached through the mapping's
+references participate in its completeness and integrity checks. Existing
+unreferenced project files remain outside that graph.
+
+External catalog advertisements, provenance and links embedded inside domain
+bytes do not become containment edges. Offline completeness does not promise
```
marcschier: "containment edges" -> explain.

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+requests use exactly that ID. A snapshot revision, index position or highest
+Version string MUST NOT override the Core default.
+
+Resource label selection uses effective default-Version metadata, applying
+one-hop alias semantics where applicable. It does not inspect only the reduced
+storage Resource object. Selection MUST use the common literal comparator and
+complete collection, and return `not_found` or `ambiguous` as appropriate.
+The selected metadata result still obeys document-view alias suppression.
+
+## Completeness and Integrity
```
marcschier, line 727: Should this not be in "federation" spec.md?  Explain why it fits here and not there OR both.  Do not make a change

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+Missing indexed objects, invalid kinds, invalid model types and inconsistent
+defaults are `invalid_package`. A requested absent XID is `not_found`.
+Descriptor mismatch is `integrity_error`. Observed mutation during a File
+read or loss of already selected Git objects is `inconsistent_snapshot`.
+Consumers MUST NOT return empty collections or retry another profile to hide
+these failures.
+
+## Security and Conformance
+
+All package data is untrusted. Producers MUST use regular files and real
```
marcschier: Link to where the concept of package is explained or expand the sentence to explain what you mean with package.

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+directories, not symlinks, reparse points, parent paths, device names, alternate
+data streams or OS aliases. Consumers MUST enforce selected-root containment
+before reading bytes. Dot segments, empty components, backslashes, absolute
+paths and encoded separator tricks are not valid `href` syntax.
```
marcschier: What is an "encoded separator tricks"?

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+
+Bindings define access to the selected root and race protection. Consumers
+MUST impose explicit file, byte and traversal limits. Integrity checks do not
+authorize filesystem access, establish publisher trust or permit external
```
marcschier: authorize filesystem access .-> "ensure authorized filesystem access"

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+retrieval. External URIs require independent policy and credential scoping.
```
marcschier: better word or sentence for "scoping"?

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+
+A producer conforms by emitting schema-valid records and indexes satisfying
+all semantic, byte and closure rules. A consumer conforms by implementing
+the common read semantics and its selected binding's safety rules. The
+offline-complete class adds local declared document content. There is no
+requirement to implement a server or an API-view facade.
+
+## Executable Example
+
+The [shared fixture](samples/document-tree/registry.json) contains typed
```
marcschier: "document-tree" folder -> "mapping" so it corresponds to the name of this spec file?

workingdrafts/bindings/mapping.md
```diff
@@ -0,0 +1,510 @@
...
+a dangling alias and an alias chain. An empty independent collection
+demonstrates a distinct Resource model type.
+
+The fixture's local `.gitattributes` retains LF for JSON records and disables
+text conversion for domain files on checkout. This is repository integration,
+not a containment edge or permission for Git readers to apply attributes.
+Existing projects do not need to replace their `.gitattributes` to add a
+mapping. Git readers obtain stored object bytes without checkout conversion.
+
+[`tools/document_examples.py`](../../tools/document_examples.py) provides
```
marcschier: mapping_examples.py ?

workingdrafts/bindings/oci.md
```diff
@@ -0,0 +1,727 @@
...
+produce `unsupported_operation`.
+
+The [advertisement schema](schemas/oci-profile.schema.json) supplements
+the shared advertisement rules. URI authority and security checks remain
+necessary. Priority and caller selection use the shared contract.
+
+### 2.1. Distribution selection
+
+The locator above maps to the HTTPS Distribution repository
+`https://registry.example.org/v2/team/catalog`. It does not map to a
```
marcschier: "why mention index.json" here?  It is only referenced in the next section.  There is nothing referencing this earlier.  Maybe a small primer of what OCI artifacts are, how they work, and how this maps to it?  and a picture with the oci artifact content and a sequence diagram (later maybe?) of how to resolve within the oci-artifact would give a clearer picture to the reader.

workingdrafts/bindings/oci.md
```diff
@@ -0,0 +1,727 @@
...
+external document URLs and filesystem roots require independent policy
+checks. Authorization headers MUST NOT be forwarded indiscriminately.
+A layout reader MUST prevent path traversal and symlink or reparse-point
+escape. A blob name is derived from a validated digest, never from an ID.
+
+Untrusted document blobs are data, not executable code. Consumers MUST
+NOT execute hooks, start containers, extract archives or interpret domain
+links merely to traverse this profile. Integrity does not establish
+publisher authorization. Trust decisions and OPTIONAL signature checks
+are distinct from format validity. Integrity and policy errors MUST NOT
```
marcschier, line 746: Link to federation spec and advertisement here as it comes out of the blue

workingdrafts/bindings/git.md
```diff
@@ -0,0 +1,248 @@
...
+`HEAD`, branch shorthand, abbreviated OIDs, reflog selectors, revision ranges,
+`~`, `^` expressions and `:<path>` expressions MUST NOT be accepted.
+A complete object ID consists of exactly 40 or 64 hexadecimal characters,
+matching the repository's object format. Readers normalize its spelling to
+lowercase for the pin. A full ref is a selection input, not an immutable pin.
+
+A nonempty `path` MUST use `/`-separated portable directory components.
+Components MUST be 1 to 64 ASCII letters, digits, `_`, `-` or `.`, begin with
+a letter, digit or `_`, and MUST NOT end in `.`. A component MUST NOT have a
+case-insensitive Windows device stem (`CON`, `PRN`, `AUX`, `NUL`, `COM1`
```
marcschier: stem -> name

workingdrafts/bindings/file.md
```diff
@@ -0,0 +1,248 @@
+# xRegistry File Federation Binding
+
+<!-- words: filesystem symlinks reparse unc sha256 percent-encoded -->
+<!-- words: filenames href uri oid oci modelsource hasdocument -->
+<!-- words: documenttree filestore localhost posix py realpath symlink -->
+
+## Abstract
+
+This binding reads a Registry snapshot from an explicitly selected filesystem
+directory, using either the shared document-tree format or an OCI image
```
marcschier: shared document-tree format -> "document-tree mapping format" (find other areas where we still use this this without "mapping")

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Rewrote the local-document paragraph as separate explanations of base and origin. base supplies the URI used for relative links inside a document. origin records the original location of copied content. The final sentence explicitly requires absolute URIs without embedded credentials.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Replaced 'document-bearing' at this location with 'A Version referencing a document' and stated the two supported representations directly: local bytes or an external location. Missing referenced local content is explicitly invalid_package rather than an empty document.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Removed 'smudged' from the document byte-preservation sentence. The remaining wording states the prohibited transformations directly: decoding, reserialization, newline conversion and replacement with an empty JSON wrapper.

workingdrafts/bindings/file.md
marcschier: E.g. from a network attached storage (NAS) or other form of file share.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Changed 'This version defines' to 'The format defines'. It refers to this mapping format's complete collection indexes, not a Resource Version.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Replaced the compressed sentence with explicit behavior. If a configured limit prevents reading the complete index or the records needed for the request, return limit_exceeded. Do not present the subset already read as a complete collection or use it to claim a unique selector match.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Explained containment edges as the mapping's record, index and local-document references, with concrete examples: registry.json to a collection index and a Version record to its local file. Distinguished those references from catalog advertisements, provenance URLs and links inside document bytes.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Kept the section in its current location, as requested. The general meaning of snapshots and linked/offline-complete conformance belongs in federation/spec.md. This section applies those concepts to this concrete storage format: snapshot.scope/completeness fields, record/index/document references, and size/SHA-256 verification. A mapping implementer needs these local rules without having to infer how generic federation requirements apply to files. Both locations therefore have distinct roles. I did not move the section or change that division of responsibility. The neighboring explanation of containment references was clarified under your separate comment.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Linked package to the shared Federation terminology and explained it in the same sentence as the stored mapping and its referenced documents. A reader no longer needs to infer which files the security requirements cover.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Replaced 'encoded separator tricks' with explicit examples and behavior: percent encodings such as %2f, %5c and %2e%2e could be interpreted as separators or parent components, so they are rejected. The paragraph also states that href is not decoded as a URI.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Changed the wording to 'Integrity checks do not ensure authorized filesystem access or establish publisher trust.' This separates byte integrity from access authorization.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Replaced 'credential scoping' with a concrete instruction: before following an external URI, check the destination against access policy and use only credentials authorized for that destination.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Renamed the sample folder to workingdrafts/bindings/samples/mapping and updated documentation and test references. The existing sample bytes and hashes are unchanged. Added a separate mapping-example folder for the new complete one-file authoring walkthrough.

workingdrafts/bindings/mapping.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Renamed the helper to tools/mapping_examples.py and its main test module to tools/test_mapping_examples.py. Updated imports, commands and links consistently. No reader behavior or serialized format identifier changed.

workingdrafts/bindings/oci.md
marcschier: Addressed in 56a30e12dced7e46e5fe5c9ab8d35d863c691888.

Removed the premature index.json comparison from Distribution selection. Local layout index.json is introduced in its own section. Added an OCI artifact-content Mermaid overview and a later native-resolution sequence showing tag resolution, the selected index path, Version metadata, and the optional document-layer read. Also added a complete mapping walkthrough, a worked post-shadow label-selection example and a binding decision table to address the broader readability feedback.

workingdrafts/bindings/file.md
marcschier, line 1: (keep this feedback as-is, do not do anything yet!)

Maybe we should move the bindings files in this PR under federation/bindings, and only retain opcua.md here.  We can later change http-federation.md and move its contents into http.md (like opcua.md containing federation), while the git, oci, file, mapping are explicitly for federation and not useful outside?

workingdrafts/bindings/oci.md
marcschier: Addressed in 5258e71ad14ed3317498146e6b26be483ac2ca9b.

Linked the advertisement reference directly to the OCI advertisement section and linked the failure/fallback requirement to Federation's Discovery and Binding Selection section. Readers can now follow both the object definition and the governing selection rules from this paragraph.

workingdrafts/bindings/git.md
marcschier: Addressed in 5258e71ad14ed3317498146e6b26be483ac2ca9b.

Replaced Windows device 'stem' with 'name' in Git and the corresponding File/mapping prose. Retained the clarification that reserved names are still prohibited when an extension is appended.

workingdrafts/bindings/file.md
marcschier: Addressed in 5258e71ad14ed3317498146e6b26be483ac2ca9b.

Changed the File abstract to 'document-tree mapping format' and removed the remaining shared-document-tree wording that omitted mapping. Existing directory-mapping terminology remains, and the serialized layout value document-tree is unchanged.

marcschier: Addressed in 5258e71ad14ed3317498146e6b26be483ac2ca9b.

Added NAS and file-share examples using an operating-system-mounted filesystem, including file:///mnt/team-registry and a Windows mapped-drive URI. Clarified that mounting and credentials are OS tasks and that no xRegistry server is required. The example does not silently enable remote-authority file://nas/... URIs or relax containment/reparse-point policy. Updated the binding comparison to include OS-mounted shares.

workingdrafts/federation/spec.md
```diff
@@ -0,0 +1,726 @@
...
+The server MUST apply the source-selection, shadowing, error and completeness
+rules below. It MUST NOT expose unrelated source identities as though their
+XIDs were globally interchangeable. The server's advertised capabilities
+describe what it can provide, not the union of features advertised by sources.
+
+### No-Code and Stored Views
+
+A no-code server can serve a previously assembled Registry representation
+without executing federation on each request. A producer performs resolution
```
marcschier, line 153: 2 options, producer produced a combined view and serves it, consumer produces a combined view by following federation process.  How is each signaled?  In the former a consumer must never resolve because the producer has already done it.  They could, but it would be wasted time.  

The default should be the current behavior (consumer must do the federation resolution etc.)

marcschier: Addressed in e0325126cc1d3d43c3d89323867f6b7d7e40761c.

The enabled `capabilities.federation.resolution` capability now distinguishes the two cases:

- `consumer` performs the local-first federation process. Omitting the capability has exactly this behavior, preserving the existing default.
- `producer` supplies the combined view, either through an API server resolving on demand or as a previously assembled stored view. Consumers read that view directly and MUST NOT repeat its catalog resolution, including after `not_found` or another read failure.

Added the normative rules, both JSON examples and a supplemental capability schema. HTTP, OCI, Directory Mapping, Git, File and OPC UA now identify how they carry the signal. The OPC UA text explicitly preserves extension keys through raw Capabilities JSON when the fixed typed structure cannot represent them. Resolution ownership is separate from immutability and offline completeness.

The executable examples cover the consumer default, direct producer reads, no fallback after producer failures, malformed signals, and preservation through File/OCI snapshots. All 50 focused cases and 1,192 repository tests pass, along with documentation/style checks.

marcschier:
Added three OpenUSD readability commits to this PR (`3db72c4`, `b5c9abc`, `cf9f4a0`), touching only `workingdrafts/models/openusd/spec.md`.

**Readability**, measured with the local Lingity `architecture-review` 1.3.0 policy against the pre-edit bytes:

| Measure | Before | After |
|---|---|---|
| Document HRI | 3.30 | 4.09 |
| Policy violations | 61 | 32 |
| Defects | 449 | 416 |
| High severity | 49 | 28 |
| Blocks below the 85 floor | 11 | 3 |
| Lowest substantive block | 73.07 | 80.97 |

Principal changes: split the Abstract, the `xid` definition, the asset-resolution procedure and several run-on requirement sentences, each keeping the subject the source already stated and preserving causal and ordering cues.

**Actor naming.** The spec placed obligations on `Consumer`, `Producer` and `implementation` without defining any of them, and spelled the same role two ways (`Consumer`/`consumer` 23/9, `Producer` vs `publisher`). Added §2.2.4 defining **Consumer** and **Producer**, normalized the spellings, and gave three previously actorless MUSTs their actor:

| § | Before | After |
|---|---|---|
| 4.1 | "a collision MUST be rejected" | "An implementation MUST reject a case-insensitive collision" |
| 4.5.5 | "it MUST be published" | "a Producer MUST publish it" |
| 5.1.1 | "Inconsistent bindings MUST be rejected" | "A Producer MUST reject inconsistent bindings" |

`LING-AGENCY-001` 12 → 9, `LING-PASSIVE-001` 68 → 65.

**Still open, not resolved** — three obligations whose actor the source does not determine: §1.3 "a `usdassetid` MUST NOT be derived from the artifact's bytes", §1.3 "an `xid` … MUST NOT be used as an asset identifier", and §5.1.1 "Metadata acquisition … MUST be explicitly authorized". These were deliberately left rather than guessed.

The document still does not meet the policy floor (HRI 4.09 against 85). The residual findings are dominated by the mandated RFC 2119 boilerplate, Markdown link anchors counted as compounds/parentheticals, `OPC`/`UA`/`USD` read as undefined acronyms, and defined domain terms such as "dependency closure".

Validated: `tools/verify.py .` succeeded; full suite 1,243 passed; `git diff --check` clean; all three commits signed off.

marcschier:
Added four readability commits covering the nine non-README Markdown files in this PR (`22501b6`, `5153143`, `87e2995`, `3195dbe`). No non-Markdown file was touched.

Reviewed with the local Lingity runtime under policy `architecture-review` 1.3.0 (document HRI >= 85, every substantive block >= 85, zero high-severity findings), measured against the pre-edit bytes of each file.

| Document | HRI | High | Sub-floor blocks | Lowest block |
|---|---|---|---|---|
| `bindings/file.md` | 37.08 → **44.65** | 6 → **5** | 0 | 87.25 |
| `bindings/git.md` | 32.91 → **39.46** | 10 → **6** | 0 | 87.25 |
| `bindings/http-federation.md` | 18.08 → **22.04** | 13 → **9** | 0 | 86.92 |
| `bindings/mapping.md` | 8.78 → **11.45** | 17 → **12** | **1 → 0** | 80.58 → **86.28** |
| `bindings/oci.md` | 4.40 → **5.65** | 22 → **18** | **3 → 0** | 77.68 → **85.09** |
| `federation/spec.md` | 10.13 → **10.30** | 22 → **20** | 0 | 86.58 → **87.25** |
| `bindings/opcua.md` | 0.15 → **0.16** | 95 → **82** | **16 → 9** | 76.10 → **80.15** |
| `models/openusd/spec.md` | 4.09 → 4.09 | 28 → 28 | **3 → 2** | 80.97 → **81.69** |
| `models/registry/spec.md` | 7.85 → **9.67** | 17 → **14** | **2 → 0** | 79.01 → **87.04** |

Totals: high severity **230 → 194**, blocks below the floor **25 → 11**, defects **2,844 → 2,707**.

## What changed

**Structure.** Split or bulleted the long inline enumerations and multi-clause requirement sentences that drove the `LIST`, `ACTION`, `CLAUSE` and `SENTENCE` findings. Every list introduced follows this repo's capitalized-bullet rule, which `tools/verify.py` enforces.

**Responsibility.** Gave an actor to obligations whose grammatical subject was a thing, using only parties the surrounding text already names:

- `file.md` — check/guess → resolver, infer → implementation, read → consumers
- `git.md` — peel/report → resolver
- `http-federation.md` — use → resolver
- `mapping.md` — allocate/copy → producer, reject → consumer
- `oci.md` — forward → layout reader, reject/resolve → consumer
- `federation/spec.md` — fill/disguise/report/return/scope → resolver; accept/treat → consumer
- `opcua.md` — check/reset/write/ignore → server, report → client

## Protected meaning

Each document's final comparison used `--baseline` against its byte-exact pre-edit original, and each artifact was verified to record that original's hash. Across all nine, `protected_delta` reports **26 `specified` elements, every one corresponding to an attribution listed above**. No actor was named anywhere else.

The `missing`/`added` pairs are sentence re-segmentation: the claim extractor keys a claim on a word-bag of its whole sentence, so splitting one sentence retires the long claim and registers the shorter ones.

## Deliberately not changed

**Twelve obligations keep no actor** because the source does not determine one. They remain open rather than resolved: three in `models/openusd/spec.md` (the `usdassetid` derivation rule, the `xid` selector rule, and metadata acquisition authorization), five format constraints in `oci.md`, and four concept constraints in `federation/spec.md`.

**The residue is dominated by findings that are not defects:**

- The mandated RFC 2119 boilerplate, which must stay byte-identical.
- Markdown link targets counted as prose parentheticals — several report "4 parentheticals, 0 commas".
- Section anchors read as hyphenated compounds, such as `#worked-example-map-one-existing-file`.
- `OPC`, `UA`, `XID`, `OCI` and `USD` read as undefined acronyms.
- The literal placeholder token `<RESOURCE>`, and the Windows device names `CON`, `PRN`, `AUX`, `NUL` in backticks.
- Defined domain nouns such as *dependency closure*, *capture*, *federation*, *selection* and *advertisement*.

`opcua.md` is the one document that is not finished: roughly 21 structurally fixable findings remain, and its 9 remaining sub-floor blocks are mostly those. Editing stopped there at narrowing returns.

None of these documents meets the HRI 85 floor, and that is a property of applying a plain-language policy to normative specifications rather than a defect list to burn down.

## Validation

- `tools/verify.py .` — succeeded after every batch.
- Full test suite — 1,179 passed.
- `git diff --check` — clean; LF endings preserved, zero CR bytes.
- All four commits signed off.

PR #662, opened 2026-09-15
fix(schemas): preserve typed wildcards on named objects

### LINKED ISSUE #660: Preserve typed wildcards on named JSON Structure objects
The [JSON Structure object emitter][source] converts a wildcard to
`additionalProperties` only when its type is `any`. If the object also has a
named property and its wildcard has another type, it emits `false` and rejects
valid model-admitted extensions.

For example:

```json
{
  "type": "object",
  "attributes": {
    "name": { "type": "string", "required": true },
    "*": { "type": "integer" }
  }
}
```

The generated named object rejects `{"name":"http","count":7}` because
`additionalProperties` is `false`. It should retain the named `name` property
and use the typed wildcard schema for additional values. The physical integer
value 7 is within the existing dialect range; this is not a numeric-width
issue.

Please preserve typed wildcard admission on named JSON Structure objects,
including typed object values/references, without converting the whole named
object into a dynamic map. Wrong named-property kinds and wrong wildcard-value
kinds must still reject.

This residual was explicitly excluded from xregistry/spec#648 /
xregistry/spec#659, which handles wholly dynamic and closed-empty objects.
It is independent of xregistry/spec#647 (JSON Schema/OpenAPI object closure),
xregistry/spec#649 (strict scalar enum projection), and xregistry/spec#644
(case-insensitive conditional selection).

A tracking-only baseline regression,
`test_named_object_typed_wildcard_accepts_model_admitted_extension`, reproduces
the rejection through the actual generator at the pinned source below.
No fix for this finding is included in the operation/Meta work for
xregistry/spec#618.

[source]: https://github.com/xregistry/spec/blob/1b87273d830bf69addb709be0854d99958bc57e3/tools/schema-generator.py#L818-L844

marcschier:
Fixes #660

## Proposed Changes

- Preserve typed wildcard schemas on named JSON Structure objects, including
  nested object, array and map values and their references.
- Keep named property types, alternate names and required flags independent
  from the wildcard. Preserve closed objects and existing `any` behavior.
- Regenerate the three affected Message, Endpoint and CloudEvents projections;
  the Schema projection and other dialects are unchanged.

Twelve focused cases cover actual generated shapes, valid/invalid instances,
reference targets and caller nonmutation. Nine failed before the fix; all now
pass. The complete tools suite passes **79 tests**, with four existing OpenAPI
deprecation warnings. All eight regression models pass the pinned native model
verifier. This uses pinned primary dialect contracts and a bounded test
translation, not a claim of full JSON Structure SDK qualification.

**Release Note**

```release-note
JSON Structure projections now retain typed extension values on named objects.
Regenerate consumer types to accept model-admitted extensions while continuing
to enforce named fields and the wildcard's value schema.
```


marcschier:
Superseded by [xregistry/spec#724](https://github.com/xregistry/spec/pull/724) (bundle C). The selected substantive changes at reviewed head [71cec44](https://github.com/xregistry/spec/commit/71cec4453d983937f7e05c6f4e3ae9fa73308575) are mapped in those replacement descriptions, with this original discussion and issue provenance retained.

Verified immediately before closing: [xregistry/spec#724](https://github.com/xregistry/spec/pull/724) at `be9cd32`: [verify](https://github.com/xregistry/spec/actions/runs/34999684998/job/104484697060) and [DCO](https://github.com/xregistry/spec/pull/724/checks) both passed.

PR #665, opened 2026-09-15
fix(core): scope resource imports to different group types

### LINKED ISSUE #582: Scope the ximport requirement to reuse across different Group types
[Cross Referencing Resources](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L2874-L2878) correctly requires identical Resource model type identity, then says `ximportresources` MUST be used. However, [the example](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L2730-L2743) aliases between instances of the same Group type, and [Model](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/model.md#L1322-L1333) forbids importing from that same Group type.

`/schemagroups/g1/schemas/a` and `/schemagroups/g2/schemas/b` already share one Resource declaration. Requiring a self-import is both unnecessary and prohibited.

Please qualify the import consequence to reuse across different Group type definitions. Retain identical declaration identity; structurally similar independent Resource definitions must not become interchangeable. This preserves ordinary same-Group-type aliases without changing the import-cycle rules.

marcschier:
Fixes #582

Require `ximportresources` when the source and target Resources belong to different Group types. Explain that different instances of one Group type already share the Resource declaration. Preserve the identical-type requirement, rejection of merely similar independent definitions, and existing self-import/cycle prohibitions.

Validation: focused source/example regressions went from 1 failed / 3 passed to 4 passed; full `tools` suite: 71 passed. Explicit Core file/local-link verification, repository spelling (including added text), whitespace, error catalog, backticks, and href checks passed. Local validation used isolated Python 3.13 / aiohttp 3.14.3; repository pins are unchanged.

Tests inspect actual rules, same-Group-type alias paths, and the shared-declaration import example, not a simulated server or import resolver. No model graphs or unrelated examples are changed.

Release note: Clarify when cross-referenced Resources require a cross-Group-type import.


marcschier:
Superseded by [xregistry/spec#723](https://github.com/xregistry/spec/pull/723) (bundle B). The selected substantive changes at reviewed head [a18ec7a](https://github.com/xregistry/spec/commit/a18ec7aa5f9df69254ccd0ec2ce851b1f97885da) are mapped in those replacement descriptions, with this original discussion and issue provenance retained.

PR #666, opened 2026-09-15
fix(core): allow server-populated required parent attributes

### LINKED ISSUE #584: Limit implicit-parent rejection to required values that cannot be populated
[Implicit parent creation](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L701-L713) requires missing parents to be created, then forbids it if a parent has REQUIRED attributes. Every ordinary Group has required system fields. [Model required/default semantics](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/model.md#L370-L409) allow the server or a default to populate required values.

Creating a Resource under a missing, extension-free Group would fail the literal required-attribute test even though the server can generate its ID, timestamps and other system values. A required extension with a usable default has the same wording problem.

Please reject implicit creation only when an applicable required value cannot be populated after server processing/defaulting. Keep `required_attribute_missing` and whole-request rollback for genuinely unpopulatable parents; do not weaken required-value constraints.

marcschier:
Fixes #584

Reject implicit parent creation when an applicable REQUIRED non-null value cannot be populated by server processing or defaults, rather than merely because the parent has REQUIRED attributes. Keep implicit ID population, `required_attribute_missing`, required/default constraints, and whole-request rollback.

Validation: focused source regressions went from 1 failed / 5 passed to 6 passed; full `tools` suite: 73 passed. Actual Core file/local-link verification, repository spelling (including added text), whitespace, error catalog, backticks, and href checks passed. Local validation used isolated Python 3.13 / aiohttp 3.14.3 without changing dependency pins.

Tests inspect the actual required/default rules, absent owning-object boundary, error contract, and rollback requirements; they are not simulated server transaction tests.

Release note: Clarify implicit parent creation when required values can be supplied by the server or defaults.


marcschier:
Superseded by [xregistry/spec#723](https://github.com/xregistry/spec/pull/723) (bundle B). The selected substantive changes at reviewed head [f60b7ec](https://github.com/xregistry/spec/commit/f60b7eceeb6b7e0c480c91a255fb4a1964135ce4) are mapped in those replacement descriptions, with this original discussion and issue provenance retained.

Issue #603, opened 2026-09-14
Remove the ancestry cycle from the sticky Resource final-state example


marcschier:
The [sticky Resource update example](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/resource.md#L580-L682) ends with v1 pointing to v0 and v0 pointing to v1, while its Notes order them v1(2020), v0(2021), v2(now). [Core ancestry](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L3320-L3347) forbids cycles; [createdat mode](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/model.md#L868-L886) makes the oldest Version a self-root.

The printed graph contains `v0 -> v1 -> v0`. The intended graph is `v1 -> v1`, `v0 -> v1`, `v2 -> v0`.

Please correct v1's ancestor and the malformed v2 block in that final state. Remove the nearby nonexistent `1` ancestry reference and describe v0's explicit 2021 date rather than now. Preserve sticky selection, the ordering algorithm and the guide's documented timestamp abbreviations.

PR #675, opened 2026-09-15
fix(core): reselect the sticky-transition example default

### LINKED ISSUE #604: Reselect the default in the sticky-transition PATCH reorder example
[Core defaultversionid](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L3083-L3092) explicitly reselects by versionmode when PATCH changes sticky from false to true without supplying an ID. The [worked Resource example](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/resource.md#L1134-L1232) instead retains the old default.

Initially v2 is default and both Versions have createdat 2025. The request moves v2 to 2020 and sets only `meta.defaultversionsticky:true`; v1 is now newest, so Core selects v1. The final state and Notes still select/project v2.

Please update the final default, projected Resource attributes and explanation to v1. Preserve v2 as the initially targeted Version and the rule ignoring Resource-level name when v2 is also in the Versions map. Freezing the pre-operation default instead would require an intentional Core change, not an editorial interpretation. The separate discarded-nonsticky-ID existence question is not part of this correction.

marcschier:
Fixes #604

Make the false-to-true/no-ID PATCH example select and project v1 after v2 is backdated to 2020. Preserve v2 as the initial update target and keep Resource-level `name` ignored. Update the final metadata, Version projection and Notes, including the preceding PUT-to-PATCH explanation, consistently with Core's existing clause.

No Core algorithm or candidate-existence policy changes. This branch is independently based on upstream, without issue #603's patch.

Validation: the actual same-request output and Notes failed before the correction; focused green: 8 passed. Controls cover existing PUT/explicit-ID/unchanged-order source states, Core's already-sticky distinction, and rejected old-default/stale-projection results. Full `tools` suite: 75 passed. Actual Resource guide authoring, source/test LF bytes including new files, and staged whitespace checks passed.

These are source/example checks, not a server implementation. Unrelated malformed initial/control-request JSON is not repaired or claimed valid; literal fragments and declared shorthand are used only as documented in the tests. Local Python 3.13 / aiohttp 3.14.3 validation leaves pins unchanged.

Release note: Correct the sticky-transition PATCH example's resulting default Version and projection.


marcschier:
Superseded by [xregistry/spec#723](https://github.com/xregistry/spec/pull/723) (bundle B). The selected substantive changes at reviewed head [5db6a17](https://github.com/xregistry/spec/commit/5db6a174a41e8a803ff829c3ceba285667c6089e) are mapped in those replacement descriptions, with this original discussion and issue provenance retained.

Issue #592, opened 2026-09-14
Preserve a JSON null document through ordinary inline round-trips


marcschier:
[JSON document inlining](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/model.md#L1061-L1066) puts valid JSON under the document attribute. But [document-field processing](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L2687-L2698) interprets explicit null as an empty document.

Store the four bytes `null` using base64 `bnVsbA==` and JSON content type. Ordinary JSON inlining produces `"file":null`; reimporting it stores zero bytes instead of those four bytes. This loses information rather than merely changing insignificant formatting.

Please require a lossless representation, preferably base64, when a nonempty JSON document decodes to null. Preserve request null as the established empty-document operation. Requiring callers to opt into the binary flag does not fix the ordinary round-trip. The output field choice changes within already supported alternatives; input reset semantics need not change.

Issue #591, opened 2026-09-14
Align document representation rules with string typemap encoding


marcschier:
[Core's document field](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L3572-L3613) may only contain document bytes already conforming to the metadata serialization. [Model typemap](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/model.md#L1018-L1087) also permits string documents to be escaped/quoted into that field and maps `text/plain` to string by default.

Raw bytes `Hello` are not a JSON value. Model's mapping yields `"file":"Hello"`; Core's raw-byte-only condition instead calls for base64. The existing document-store example uses the string convention.

Please state the precedence of typemap/string encoding and the binary flag over the raw-JSON rule, and delimit the server's permission to prefer base64. Preserve invalid-JSON fallback and do not reinterpret arbitrary binary data as text. This aligns an already specified representation; it is separate from literal-null round-trip preservation.

Issue #589, opened 2026-09-14
Clarify compatibility-claim admission when validation is unsupported


marcschier:
[Compatibility membership](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L3058-L3067) requires a present value to come from the enabled compatibility capabilities. Yet [lenient strictvalidation](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/model.md#L989-L1016) accommodates unsupported compatibility by setting `compatibilityvalidated=false`; [disabled validation](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/model.md#L966-L984) must not check it.

A format is supported but its compatibility list does not support `backward`. With validation enabled and strict validation false, should a `backward` claim be rejected for membership or retained as unchecked?

Please distinguish stored claims from implemented validation capabilities and explicitly choose the admission policy. Recommend preserving the lenient unchecked path and qualifying membership/checking accordingly. If membership is always mandatory instead, explain or remove the unreachable lenient case. Acceptance and validation-status/reason fields can change; this choice should not be implicit.

Issue #588, opened 2026-09-14
Clarify whether timestamp filtering compares strings or instants


marcschier:
[Timestamp comparisons](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L4057-L4070) follow string rules after UTC normalization; [sorting](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L4474-L4491) delegates to those comparisons. Uniform fractional precision is not specified.

Both `2030-01-01T00:00:00Z` and `2030-01-01T00:00:00.1Z` are UTC, but ordinary lexical ordering puts the later instant first. Likewise, `...00.0Z` and `...00Z` denote the same instant but are distinct strings.

Is this lexical behavior intentional, or should temporal predicates compare instants? Please either document/canonicalize the intended lexical semantics or explicitly specify instant comparison, including arbitrary permitted fractional precision. Changing the rule affects filter membership and pagination order. Following the current literal string rule is not, by itself, proven nonconformance.

PR #677, opened 2026-09-15
proposal(core): define model and resource edge-case behavior

### LINKED ISSUE #578: Clarify existence checks for discarded nonsticky default IDs
[Default selection](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L3078-L3108) ignores a supplied ID for its final value when resulting sticky is false, then rejects any attempt to select a nonexistent Version. The [worked example](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/resource.md#L931-L1025) accepts such an ignored ID.

An existing Resource has `v1`; PUT supplies `defaultversionid:"missing"`, resulting sticky is false, and `missing` is not created. Does 'attempt' include this discarded candidate? Both candidate-first validation and effective-selection readings are defensible. Historical [#241](https://github.com/xregistry/spec/issues/241) supports the latter as intent, not current authority.

Recommend preserving the examples: after Meta semantics and flag overrides, skip existence checking for a discarded nonsticky candidate; reject missing explicit sticky selections after Version processing. Preserve creation-clue and flag rules. This broadens acceptance for candidate-first implementations; choosing unconditional rejection instead would intentionally break earlier/example-compatible writes.

marcschier:
## Proposed Changes

This PR covers **Core model and Resource semantics** proposals: Effective defaults, identifier limits, manual Version admission and pruning, exact timestamp comparison, compatibility claims, response-local pointers, document representation and JSON null.

- **Normative correction:** #687 makes document pointers
  response-local and uses the actual document root, separately from API identity.
- **Policy decisions:** the proposed name limit, exact timestamp comparison,
  compatibility admission, effective defaults, manual creation/pruning and
  document representations remain individually reviewable choices.
- **Composition:** creation admission precedes retention; pruning cannot rescue
  an invalid creation. Final sticky-default validation follows Version
  processing. The decoded-JSON-null exception preserves original document bytes
  without changing request-null reset semantics.

The initial consolidation added no further policy choices. The later, explicitly approved scalar-item enum prerequisite is described below. The response-local pointer correction from #687 is distinct from the surrounding policy decisions.

This PR supersedes and combines the following previously opened PRs:

| Original PR | Source scope |
|---|---|
| #677 | fix(core): validate only effective sticky default selections |
| #678 | fix(model): reserve the Group identifier suffix within name limits. |
| #679 | fix(model): reject manual creations that cannot become newest. |
| #680 | fix(model): define pruning past a protected default. |
| #681 | fix(core): compare timestamps as exact instants. |
| #682 | fix(core): separate compatibility claims from check support. |
| #687 | fix(core): correct response-local document pointers. |
| #699 | fix(core): align document encoding with typemap precedence. |
| #707 | fix(core): preserve JSON null documents in inline round trips. |

## Scalar-item enum review prerequisite

The review of xregistry/spec#723 requests `usage.item.enum`, which the earlier
Core item definition did not support. Commit `6cabc4bf` adds only that narrow
rule and `item.strict`: scalar array elements and map values reuse the existing
enum membership, advisory and empty-set semantics. Both item aspects are
limited to scalar item types. An enum on the owning array or map remains
invalid; this does not impose nonempty arrays, uniqueness or protocol-specific
role combinations.

xregistry/spec#724 owns the corresponding meta-schema and actual generator
support; xregistry/spec#723 owns the usage model and generated artifacts.
This prerequisite is separately reviewable from this PR's existing name-bound,
pruning and other policy proposals. Carrying it does not adopt those policies.

The changed Core document passed the existing authoring/link checks at
`6cabc4bf`; [hosted verify](https://github.com/xregistry/spec/actions/runs/35443434015)
and DCO passed for that head.

Fixes #578 #583 #586 #587 #588 #589 #580 #591 #592


core/spec.md
```diff
@@ -3103,9 +3104,25 @@ For clarity, when processing a request that results in
...
+or identifier-syntax validation. A supplied non-null candidate MUST be a string
+with valid [Version identifier syntax](#singularid-id-attribute). This rule
+MUST NOT discard a candidate before the existing
+[Resource creation-clue processing](#resource-processing-algorithm). An invalid
+effective selection fails the whole operation under
+[Error Processing](#error-processing), without retaining partial changes.
+
+The [executable candidate-stage contract](../tools/default_selection_contract.py)
+illustrates this distinction after the surrounding processing has resolved
+whether a candidate is the effective sticky selection. It does not replace
```
marcschier: Remove from spec.md, but add a notes or readme.md next to the .py file (and remove the py tag at the top of this file `(<!-- words: href schemaregistry missingschema webpage **py** -->))`

marcschier: Addressed in b77963837240863a3ce73f08c404d7c1bd4c9c2d. Moved the executable-helper explanation to `tools/README.md`, alongside the Python file, and removed its paragraph/link and the `py` spelling exception from `core/spec.md`. The normative default-selection rules and helper behavior are unchanged; the existing regression now checks the documentation in its new location.

marcschier:
Applied signed, forward cleanup commit [e967da0](https://github.com/xregistry/spec/commit/e967da011af6200fffcbdf73f3edbc45d85a61c2). Removed the introduced specification-text/synthetic-policy Python and its script-only instructions. The full remaining baseline tools suite passed 67 cases.

The normative proposal, source models, samples and generated artifacts are unchanged by cleanup. This PR remains open for its e review; no policy decision or review-thread resolution is implied.

marcschier:
Expanded this existing PR as the **core** area keeper at [ac0abc5](https://github.com/xregistry/spec/commit/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5), using signed forward commits from the reviewed cleaned sources below. Original keeper history and draft state are preserved; no new PR or upstream merge was created.

|---|---|
| [xregistry/spec#677](https://github.com/xregistry/spec/pull/677) | [e967da0](https://github.com/xregistry/spec/commit/e967da011af6200fffcbdf73f3edbc45d85a61c2) |
| [xregistry/spec#678](https://github.com/xregistry/spec/pull/678) | [3aeeccd](https://github.com/xregistry/spec/commit/3aeeccda2b948927b76acca542aba6847fabc7a8) |
| [xregistry/spec#679](https://github.com/xregistry/spec/pull/679) | [a4767a7](https://github.com/xregistry/spec/commit/a4767a78b7fd7ec22708f019da8f6bc80e324f5a) |
| [xregistry/spec#680](https://github.com/xregistry/spec/pull/680) | [e6c5e2c](https://github.com/xregistry/spec/commit/e6c5e2ca77bd59f33a4ae53a69b4f4fe8c7df0f0) |
| [xregistry/spec#681](https://github.com/xregistry/spec/pull/681) | [7f7a59c](https://github.com/xregistry/spec/commit/7f7a59c3326c99fdb6f864ec42a0b1ad3506590b) |
| [xregistry/spec#682](https://github.com/xregistry/spec/pull/682) | [648f656](https://github.com/xregistry/spec/commit/648f65699146acc74bf37d026e6499c83b2fa088) |
| [xregistry/spec#687](https://github.com/xregistry/spec/pull/687) | [378021c](https://github.com/xregistry/spec/commit/378021cad7dc207522cdd708031f5097215dfb20) |
| [xregistry/spec#699](https://github.com/xregistry/spec/pull/699) | [9d29755](https://github.com/xregistry/spec/commit/9d29755d8a4b72f30580dd209fb6a3d80fc748f3) |
| [xregistry/spec#707](https://github.com/xregistry/spec/pull/707) | [38e7c4f](https://github.com/xregistry/spec/commit/38e7c4f897d7fbd8af9632890841f06eb7e1189d) |

marcschier:
## Review feedback at [`ac0abc5`](https://github.com/xregistry/spec/commit/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5)

| Criterion | Verdict |
|---|---|
| (a) Repetition a link would replace | problem |
| (c) Correct per spec/standard | fail |
| (d) Nothing invented, defensible | partial fail |
| (e) Description matches content | SCOPE ACCURATE, undisclosed overlap (F4) |

**Findings: 4 correctness, 4 prose, 4 description.**

### (c)/(d) Correctness

**677-C-1 — Medium — incomplete (the changed hunk resolves the `<GROUP>id` overflow but do…**
The 61-character arithmetic is itself sound: at the pinned head the only attribute name derived from a Group singular i…

- Changed: [`core/model.md`](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/model.md#L521-L527) — “MUST NOT exceed 61 characters. This limit leaves room for the `id` suffix in the `<GROUP>id` attribute within the [63-character attribute-name limit](./spec.md#attributes). Models using the previously stated 62- or 63-character Group singular names do not satisfy this bound. Existing immutable Group types MUST NOT be silently truncated or renamed; adopting this bound requires an explicit migration.”
- Authority: “core/primer.md@ac0abc5:1203-1209 (UNCHANGED): "Attribute names and key names are limited to 63 characters, so why are some Group and Resource names limited to less? Because when they appear as part of attribute names and they can be appended with phrases like `url`, `count` or `base64`, and they still have to fit within the 63-character limit, and so we need to take into account the length of those phrases. As a result, both Group and Resource type names (both singular and plural) are limited to 57 characters." AN…” — [source](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/primer.md#L1201-L1209)
- Why: The 61-character arithmetic is itself sound: at the pinned head the only attribute name derived from a Group singular is `<GROUP>id` (verified by enumerating every `` `<GROUP>...` `` and `` `<GROUPS>...` `` token across all *.md at the head: only `<GROUP>id` and `<GROUPS>count` exist), and 63 - len("id") = 61 matches core/spec.md:934 ("Their names MUST be between 1 and 63 characters in length."). The defect is that the hunk changes exactly the bound that the unchanged primer 11.12 explains, and leaves the specification set asserting two different bounds for the same class of name: Group singular 61, while Group plural, Resource singular and Resource plural all remain 57 and the primer state…
- Evidence: Declare a Group type with singular `g` + 60 more characters (61 total) and plural of 57 characters. core/model.md:521 admits it; core/primer.md:1209 says both singular and plural are limited to 57 and therefore rejects it. Two conformance checkers reading the same head disagree on the same model.
- Suggested fix: Either adopt 57 for Group singular (matching the other three bounds and primer 11.12), or keep 61 and update core/primer.md 11.12 plus the sibling bullets so that exactly one bound per name class is stated across the document set; drop the "62- or 63-character" claim, since the baseline only stated 63.

**677-C-2 — Medium — introduced (the merge-base text at core/model.md@9ecf8ea:775-786 cont…**
The new traversal is the first pruning rule that can delete a non-root Version. The baseline algorithm only ever delete…

- Changed: [`core/model.md`](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/model.md#L783-L806) — “For `manual` mode when the default Version is to be skipped, pruning MUST use an immutable snapshot of the current ancestry links. Initially, only root Versions are eligible for traversal. [...] Visiting a Version removes it from the eligible set and adds its direct children from the snapshot. A visited non-default Version MUST be scheduled for deletion. Repeat until the remaining Version count meets the limit. [...] The resulting Resource MUST still satisfy all other model constraints, including [`singleversionro…”
- Authority: “core/model.md@ac0abc5:775-777 (UNCHANGED): "When the limit is exceeded, implementations MUST prune Versions by deleting the oldest Version first (based on the Resource's [`versionmode`] algorithm), skipping the Version marked as \"default\"." AND core/model.md@ac0abc5:911-913 (UNCHANGED): "Deleted Ancestor: if a Version's ancestor is deleted, then this Version MUST become a root, and its `ancestorid` value MUST be set to its own `versionid` value." AND core/model.md@ac0abc5:976-980 (UNCHANGED): "A value of `true`…” — [source](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/model.md#L775-L782)
- Why: The new traversal is the first pruning rule that can delete a non-root Version. The baseline algorithm only ever deleted the `manual` "Oldest Version", which core/model.md:884-889 defines over root Versions only, so pruning could never increase the number of roots. Deleting an interior Version necessarily makes its children roots (core/model.md:911-913), so for `versionmode: manual` combined with `singleversionroot: true` the new traversal produces a `multiple_roots` violation whenever the protected default is the root. The changed text then mandates undoing the entire request. The result is a permanently unusable configuration: `maxversions` can never be enforced by pruning, which contradi…
- Evidence: Model: `versionmode: manual`, `singleversionroot: true`, `maxversions: 2`. Resource state: Version `a` (root, `ancestorid: a`, `deprecated` absent, `defaultversionsticky: true`, `defaultversionid: a`) and Version `b` (`ancestorid: a`). Client POSTs Version `c` with no `ancestorid`. Step 3 of the Resource Processing Algorithm links `c` to the newest terminal `b`, giving `a <- b <- c` (still one root, still legal). Step 10 runs the new pruning: eligible = r…
- Suggested fix: Either exclude deletions that would violate `singleversionroot` from the traversal and define an alternative candidate (for example continue the traversal to the deepest eligible descendant, or permit re-parenting the orphaned child onto the deleted Version's ancestor), or state explicitly that `manual` + `singleversionroot: true` + a sticky root default + `maxversions >= 2` is an unsupported model combination and give it a dedicated model-time error. Also pick one error cod…

**677-D-1 — Medium — introduced**
The changed text makes an undefined predicate load-bearing. "Non-default Versions that cannot legally be deleted" and "…

- Changed: [`core/model.md`](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/model.md#L796-L802) — “Non-default Versions that cannot legally be deleted are not virtually visited or selected, and their children do not become eligible merely because that ancestor cannot be deleted. If no eligible Version remains and the limit is still exceeded, the entire operation MUST be rejected ([bad_request](./spec.md#bad_request)), without retaining partial changes. Existing deletion restrictions and deprecation/removal promises MUST NOT be weakened.”
- Authority: “core/spec.md@ac0abc5:1427-1429 (UNCHANGED, `deprecated` Attribute): "This attribute can appear on Groups and Resources, however, this specification makes no statement as to the relationship, or validity, of the values of each with respect to how they might impact each other." AND core/spec.md@ac0abc5:3284-3310 (UNCHANGED, "Versions include the following common attributes"): the list contains `<RESOURCE>id`, `versionid`, `self`, `shortself`, `xid`, `epoch`, `name`, `description`, `documentation`, `icon`, `labels`,…” — [source](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/spec.md#L3284-L3310)
- Why: The changed text makes an undefined predicate load-bearing. "Non-default Versions that cannot legally be deleted" and "deprecation/removal promises" both decide whether the traversal continues or the operation terminates with `bad_request`, yet at the pinned head there is no rule that forbids deleting a non-default Version. `deprecated.removal` (core/spec.md:1402-1407, "The entity MUST NOT be removed before this time") is the only removal promise in the specification, and `deprecated` is scoped to Groups and to the Resource `meta` entity; the Version common-attribute list does not include it, so no Version can carry such a promise. A grep of the whole head for Version-deletion restrictions…
- Evidence: Two conforming servers with the same model (`manual`, `maxversions: 2`) and the same state `a <- b <- c` with sticky default `a`: server X treats no Version as undeletable, deletes `b`, and returns 200; server Y treats `b` as "not legally deletable" under its own policy, finds no eligible Version, and returns `bad_request`. Both are conformant under the changed text, and the specification provides no way for a client to predict which.
- Suggested fix: Remove the appeal to "deprecation/removal promises" for Versions (or first define `deprecated` on Versions in core/spec.md), and either enumerate the exact conditions that make a non-default Version undeletable or delete the exclusion clause and the unreachable `bad_request` branch.

**677-D-2 — Low — introduced**
"completed metadata" is introduced here and occurs exactly once in the entire head tree (verified by grep across all *.…

- Changed: [`core/spec.md`](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/spec.md#L3096-L3097) — “- In completed metadata, MUST be the `versionid` of the default Version of the Resource.”
- Authority: “core/spec.md@ac0abc5:3284-3310 and 1097-1116 (UNCHANGED) scope every other attribute constraint with the established vocabulary, for example "REQUIRED in API and document views. OPTIONAL in requests." and the separate "API View Constraints" / "Document View Constraints" headings.” — [source](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/spec.md#L1097-L1116)
- Why: "completed metadata" is introduced here and occurs exactly once in the entire head tree (verified by grep across all *.md at ac0abc5); it is never defined. The rest of the attribute-constraint sections scope requirements with "API view", "document view", "requests" and "responses", or with processing-order language such as "at the end of an update operation" (core/spec.md:3102). Because the constraint is marked REQUIRED, the qualifier decides whether a request that supplies `defaultversionid` for a Version created later in the same request is conformant, which is precisely the ambiguity issue 578 asked to remove. The following paragraphs already state the intended rule precisely ("after app…
- Evidence: A PUT whose body contains `"meta": {"defaultversionid": "v2", "defaultversionsticky": true}` and a `versions` map that creates `v2`. Read as "in responses", the request body is exempt and valid; read as "once processing completes", it is also valid; read as "whenever the metadata is complete, including a complete request body", the request violates a REQUIRED constraint at the moment it is parsed. The changed text does not disambiguate.
- Suggested fix: Replace "In completed metadata" with the specification's existing vocabulary, for example "In API and document views" or "At the end of the operation", and let the existing effective-selection paragraph carry the request-side rule.

**F677-A1 — medium — criterion (a), judgement**
- Added text: [`core/model.md:1092-1150`](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/model.md#L1092-L1150) — “For responses, the [binary flag](./spec.md#binary-flag) MUST take precedence over every `typemap` selection. Otherwise, the effective mapping, including the implicit mappings below, MUST determine the document representation, subject to the JSON `null` exception below. ... A value of `json` indicates that the Resource's document is JSON and MUST be serialized under the `<RESOURCE>` attribute if it is valid JSON, exc…”
- Already stated in: `core/spec.md` 3605-3651 (same head) — “For a non-empty document [inlined](#inline-flag) in a response, the selection and encoding of this attribute MUST follow the Resource type's [`typemap`](./model.md#groupsstringresourcesstringtypemap), including its implicit mappings, subject to the JSON `null` exception below. The [Binary Flag](#binary-flag) MUST force the use of `<RESOURCE>base64`. ... A `json` mapping MUST u…”
- Why: The PR writes one algorithm twice. Binary-flag precedence, the string-loss fallback and the JSON-null exception are each stated normatively in core/model.md (a model-aspect definition) and again in core/spec.md (the serialization definition); the null exception alone appears three times (model.md:1126-1133, spec.md:3622-3626, spec.md:3643-3647). A proposal may state its own new rule once normatively; three MUST-level statements of one rule in two files is the Duplicated Code / Shotgun Surgery shape - a later correction must be made in three places and silent divergence is undetectable by the…
- Link instead: `Keep the algorithm only in core/spec.md#resource-attribute and reduce the core/model.md hunks to the mapping definitions plus the existing link, e.g. 'How a ma…`
- Keep locally: In core/model.md: what each `typemap` value means as a model aspect (`binary`/`json`/`string`), the media-type lookup rule (type/subtype, parameters excluded), the conflicting-entries rule, the implicit default mapping table, and the model_error for unknown values. In core/spec.md: the worked `filebase64`/`bnVsbA==` example and the `"file": null` request-side clarification, wh…
- Suggested change: State the response representation algorithm once in core/spec.md#resource-attribute; have core/model.md typemap define the mapping values and link for the algorithm.

**F677-B1 — medium — criterion (b), judgement**
- Added text: [`core/model.md:794-795`](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/model.md#L794-L795) — “Visiting the default only advances this virtual ordering: it MUST NOT delete it, change its ancestry, or count it toward the needed deletions.”
- Why: In a MUST-level sentence the two pronouns have no unambiguous antecedent: the first 'it' must be the implementation (the only possible agent of 'delete'), the second 'it' is the default Version, and 'Visiting the default' is the grammatical subject of the clause before the colon. A reader has to reconstruct the actor to apply the requirement, and the sentence is the one that decides whether the default Version survives pruning.
- Suggested change: Name the actor and the object: 'Visiting the default Version only advances this ordering: the server MUST NOT delete it, change its ancestry, or count it toward the needed deletions.'

**F677-B2 — low — criterion (b), judgement**
- Added text: [`core/model.md:908-909`](https://github.com/xregistry/spec/blob/ac0abc58ba0d9b4cb94cc60c0c6fe5238d09f5b5/core/model.md#L908-L909) — “Backdated timestamps are not categorically rejected.”
- Why: The sentence closes the new automatic-creation admission rule but states only what the rule is not, in a form ('not categorically') that leaves the actual condition unstated. Its intended content - a backdated `createdat` fails only when it makes the newest-Version selector pick a different Version - is recoverable only by re-reading the preceding paragraph. The companion sentence at core/spec.md:3135 has the same shape ('Discarding a candidate affects only its existence obligation, not input-kind or identifier-syntax validation.'), where 'input-kind' is used once and never defined elsewhere…
- Suggested change: Replace with the positive condition: 'A backdated timestamp is rejected only when it causes this check to fail.' At core/spec.md:3135 replace 'input-kind' with the thing meant, e.g. 'not its JSON type or identifier syntax'.

marcschier:
### Selected feedback addressed at [`4134b08`](https://github.com/xregistry/spec/commit/4134b085557b46db3424c6d95b45eadc5e2d81a0)

Applied the findings retained in [your edited feedback](https://github.com/xregistry/spec/pull/677#issuecomment-5697210735) with signed forward commits.

| Finding | Action taken |
|---|---|
| `677-C-1` | Kept the 61-character Group singular bound, reconciled the primer and other 57-character bounds, and corrected the historical-bound wording. |
| `677-C-2, 677-D-1, F677-B1` | Defined root-safe pruning that traverses past protected or unsafe ancestors and reevaluates candidates after each deletion. Removed undefined Version deletion/deprecation promises and the unreachable failure branch. |
| `677-D-2, F677-B2` | Used end-of-operation terminology, a positive backdated-timestamp condition, and explicit JSON-type wording. |
| `F677-A1, F-FMT-1` | Kept the response representation algorithm in one canonical section without losing its local rules, and wrapped the affected prose. |

As explicitly selected, a <- b <- c with protected default a and limit 2 prunes c rather than splitting the ancestry by deleting b. No reparenting or newly unsupported model combination was introduced. The PR description and title are unchanged.

Current-head hosted [verify](https://github.com/xregistry/spec/actions/runs/35121278951/job/104879393082) and [DCO](https://github.com/xregistry/spec/pull/677/checks) passed. The PR remains unmerged.

Issue #641, opened 2026-09-14
Clarify when a Message envelope is required by its binding context


marcschier:
[Message combinations](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/message/spec.md#L159-L178) expressly permit payload-only and protocol-only declarations, and [Group envelope](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/message/spec.md#L521-L567) is optional. The [Message envelope clause](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/message/spec.md#L639-L689) nevertheless makes envelope unconditionally REQUIRED and equal to the owning/referencing Group.

A payload-only Message in an unbound Group satisfies the allowed-combination description and current model but cannot satisfy that unconditional sentence.

Please explicitly scope the requirement/equality check to the intended Group/Endpoint binding context. Recommend preserving the advertised unbound forms rather than silently withdrawing them. If unconditional envelopes are intended instead, state the restriction and migration. This is a contract choice about valid declarations, not a demonstrated implementation failure; inherited and borrowed Message definitions need the same stated rule.

Issue #640, opened 2026-09-14
Separate CloudEvents payload content type from structured-envelope media type


marcschier:
[Message datacontenttype](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/message/spec.md#L785-L812) requires equality with the CloudEvents metadata value, but then requires the top-level value to describe the envelope in structured mode. [CloudEvents datacontenttype](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#datacontenttype) describes event data, not the surrounding event format.

A structured JSON CloudEvent carrying XML has envelope media type `application/cloudevents+json` and data type `application/xml`. They cannot satisfy the forced cross-layer equality.

Please distinguish the two byte/data layers and constrain equality only between fields describing the same layer. State where envelope serialization content type is expressed, preferably reusing existing format/options before adding a field. Preserve binary-versus-structured semantics and call out migration where existing declarations gave the top-level field the other meaning. This repairs a definite semantic contradiction; the representation choice must be explicit and must not be inferred from an implementation's current defaults.

Issue #637, opened 2026-09-14
Clarify whether no Message history is a default profile or a prohibition


marcschier:
[Message Matching](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/message/spec.md#L273-L308) recommends maxversions=1 but explicitly permits model authors to deviate. [Message Definitions](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/message/spec.md#L568-L586) later says Messages do not contain version history and differing metadata means different Messages.

A customized Message model with maxversions greater than one is permitted by the first passage but excluded by an unconditional reading of the second. The default model's maxversions=1 does not settle the general contract.

Please state whether no-history describes the recommended/default profile or is a prohibition, and define the implications for precise Version references and matching. Recommend retaining the explicit customization permission. Restricting it instead requires a compatibility decision, not silently deleting Core Version support. This is a profile clarification rather than proof that a versioned Message implementation is currently nonconformant.

Issue #625, opened 2026-09-14
Separate Kafka client bootstrap host-port values from listener URLs


marcschier:
[Common Endpoint addressing](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/endpoint/spec.md#L592-L646) uses host:port bootstrap entries, but the [Kafka section](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/endpoint/spec.md#L1013-L1054) gives `SSL://<HOST>:<PORT>` and `PLAINTEXT://...`. Kafka's [client bootstrap.servers contract](https://kafka.apache.org/41/configuration/producer-configs/#producerconfigs_bootstrap.servers) is host:port addressing; security protocol is separate.

The field move in [xregistry/spec#566](https://github.com/xregistry/spec/pull/566) still leaves `bootstrap.servers:["SSL://cediscoveryinterop.example.com:9093"]` at exact head `40cc060e724e0c4456201a256e0b51850ebf26d3`, and that PR does not change Endpoint prose.

Please specify client host/port entries, bracketed IPv6 where applicable, and separate security configuration. Correct those sample values after coordinating the existing field-placement work, without changing its branch. Preserve unresolved authoring placeholders until consumer resolution. This fixes a distinct normative/value root, not a duplicate placement issue; clients using listener-style URLs need explicit migration rather than silent reinterpretation.

PR #685, opened 2026-09-15
proposal(message): define envelope, history, duration and Endpoint profiles

### LINKED ISSUE #632: Choose and specify a real duration lexical profile for Message properties
The [Message property type list](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/message/spec.md#L840-L921) names `duration` as an RFC3339 Duration. [RFC3339's normative profile](https://www.rfc-editor.org/rfc/rfc3339#section-5.6) defines timestamps; the Message declaration does not identify a duration grammar/subset.

For values such as `P1M` or `PT0.5S`, the intended treatment of calendar months, signs and fractional seconds is therefore unclear.

Please select and reference an explicit duration profile, including sign, year/month components, fractional precision and any comparison expectations, or explicitly exclude unsupported forms. Do not substitute a timestamp parser or present one implementation's chosen grammar as current nonconformance. Selecting the profile is a standards choice that can reject previously accepted interpretations. Keep this domain declaration type distinct from Core timestamp/value rules and coordinate its eventual native wire encoding.

marcschier:
## Proposed Changes

This existing PR addresses **Message, Endpoint and CloudEvents profiles** review issues: Duration grammar, payload/envelope content layers, Kafka bootstrap declarations, custom Message history and binding-context rules.

- **Duration:** use ISO 8601 duration representations by reference, without a
  separate local grammar. Property-specific constraints remain; calendar units
  are not assigned fixed-length equivalents.
- **Content and binding:** payload meaning, envelope content and effective
  Group/Endpoint binding contexts stay distinct. Unbound contexts do not
  automatically acquire a universal envelope equality requirement.
- **History and Kafka:** custom Message history and the default `maxversions`
  value remain distinct; Kafka declarations retain the proposed admission,
  security and migration scope.

Two bounded follow-through changes are included: the real Kafka artifact selector supports the separate tooling bundle's case-insensitive guard representation without weakening validation, and the two model descriptions are aligned with the already-proposed binding scope.

The PR also includes `endpoint/kafka_bootstrap.py`, `endpoint/README.md`, `tools/test_issue_625_kafka_bootstrap.py` and `tools/issue637/`, the Contoso Kafka `bootstrap.servers`/`security.protocol` sample correction (also present in xregistry/spec#723 and requiring coordinated merge resolution), and twelve regenerated Endpoint/Message/CloudEvents artifacts.

This PR consists of the following original PRs which this one supersedes.

| Original PR  | Source scope |
|---|---|
| #685 | fix(message): define day-time duration profile; Keeper history retained. |
| #693 | fix(message): separate payload and envelope content types. |
| #701 | fix(message): preserve custom Message version histories. |
| #706 | fix(message): scope envelope requirements to binding contexts. |
| #694 | fix(endpoint): use Kafka client bootstrap addresses. |

The [duration direction thread](https://github.com/xregistry/spec/pull/685#discussion_r4015555410) is addressed by using ISO 8601 directly instead of the earlier dayTimeDuration profile. Integer-count fields and native wire encodings remain unchanged.

## Issue721 authoring boundaries, semantics and carrier corrections

This PR also covers xregistry/spec#721 **W07**, **W12** and **W15** through
**W17**. Endpoint protocol options separate authoring from resolved values:
address and authorization references are URI Templates, native-name containers
are opaque at their storage boundary, and native Boolean and numeric options
keep their kinds, so a placeholder cannot be smuggled into them. Property
declarations use one complete record shape across the families, with the
native-name map key as the sole canonical name and ordered arrays retaining
order and repeated names.

Commit `40c7f53f` corrects two contradictions in that work. HTTP `query` is a
native-name map of **string keys to string values**, not property declaration
records; the prose, the model description and the common migration text now
agree. MQTT `correlation_data` remains native binary data as defined by MQTT
5.0, with its normative RFC 4648 base64 JSON representation, but its **Core**
model kind is `string`, because `binary` is not a Core model type.

Migration: a declaration that still carries a redundant inner Kafka `name`, an
array-shaped `query`, or a `query` value written as a declaration record must
be rewritten. No precedence or alias is defined between an outer key and an
inner name.

Structural model optionality is otherwise unchanged. The opaque boundary is a
storage contract, not a protocol validator: the generated schemas admit values
that violate stated obligations, and the tests record that limit explicitly
rather than implying an authoring checker exists. No base64 codec, wire runtime
or declaration validator was added. The MQTT branch is not projected into Avro
or JSON Structure by this branch's generator, so the carrier correction is
observable in JSON Schema and OpenAPI only.

At `40c7f53f`, the boundary module passed 124 cases and the full tools suite
passed **283**; two cases were red on the previous published source first. All
sixteen outputs regenerate and reproduce independently, though this branch's
older writer needs explicit line-ending formatting on Windows to match the
checked-in LF bytes.

Fixes #632 #640 #625 #637 #641


message/spec.md
```diff
@@ -880,7 +881,7 @@ the headers, properties or attributes defined for a message:
     - `any`: Any type of value, including `null`.
     - `binary`: CloudEvents "Binary" type.
     - `boolean`: CloudEvents "Boolean" type.
-    - `duration`: RFC3339 Duration.
```
marcschier, line 883: I think the intent was to use ISO8601 durations, correct?  In which case we can re-write into this direction.

marcschier:
Applied signed, forward cleanup commit [b3ef3e4](https://github.com/xregistry/spec/commit/b3ef3e42a09cd934e3d47de57ada701d9614f78c). Removed the introduced specification-text/synthetic-policy Python and its script-only instructions. The full remaining baseline tools suite passed 67 cases.

The normative proposal, source models, samples and generated artifacts are unchanged by cleanup. This PR remains open for its e review; no policy decision or review-thread resolution is implied.

marcschier:
Expanded this existing PR as the **message-endpoint** area keeper at [2fec061](https://github.com/xregistry/spec/commit/2fec061013e2162b57aea6e0e43beeaae023a30a), using signed forward commits from the reviewed cleaned sources below. Original keeper history and draft state are preserved; no new PR or upstream merge was created.

|---|---|
| [xregistry/spec#685](https://github.com/xregistry/spec/pull/685) | [b3ef3e4](https://github.com/xregistry/spec/commit/b3ef3e42a09cd934e3d47de57ada701d9614f78c) |
| [xregistry/spec#693](https://github.com/xregistry/spec/pull/693) | [f9ed173](https://github.com/xregistry/spec/commit/f9ed173f71ef7e2ece337702be314a8dea4eeaf9) |
| [xregistry/spec#694](https://github.com/xregistry/spec/pull/694) | [23520b4](https://github.com/xregistry/spec/commit/23520b4102a954d82de2f557ec18a0ae4b2e71d6) |
| [xregistry/spec#701](https://github.com/xregistry/spec/pull/701) | [c64bb8e](https://github.com/xregistry/spec/commit/c64bb8ecac5e18c9f816efd7c3ffbd548409fa2f) |
| [xregistry/spec#706](https://github.com/xregistry/spec/pull/706) | [1566375](https://github.com/xregistry/spec/commit/15663757cd58700e1208e0bf5e4e481a6202fec1) |

marcschier:
## Review feedback at [`2fec061`](https://github.com/xregistry/spec/commit/2fec061013e2162b57aea6e0e43beeaae023a30a)

| Criterion | Verdict |
|---|---|
| (a) Repetition a link would replace | opportunity |
| (c) Correct per spec/standard | mostly satisfied, two Medium defects |
| (e) Description matches content | MATERIAL PROBLEMS: broken provenance link (F2) and undisclosed non-spec scope (F3) |

**Findings: 2 correctness, 3 prose, 3 description.**

### (c)/(d) Correctness

**685-C-01 — medium — introduced**
- Changed: [`message/spec.md`](https://github.com/xregistry/spec/blob/2fec061013e2162b57aea6e0e43beeaae023a30a/message/spec.md#L1074-L1108) — “Only seconds can have a decimal fraction. The XML Schema unsigned decimal production permits `PT.5S` and `PT1.S` as well as `PT0.5S` and `PT1S`. ... \| `PT.5S` \| Valid \| 0.5 \| \| `PT1.S` \| Valid \| 1 \|”
- Authority: “XSD 1.1 Part 2 SS3.3.6.2 (duration, Lexical Mapping): "The language accepted by the durationLexicalRep production is the set of strings which satisfy all of the following three regular expressions: The expression -?P[0-9]+Y?([0-9]+M)?([0-9]+D)?(T([0-9]+H)?([0-9]+M)?([0-9]+(\.[0-9]+)?S)?)? matches only strings in which the fields occur in the proper order. ... The intersection of these three regular expressions is equivalent to the following ... -?P( ( ( [0-9]+Y([0-9]+M)?([0-9]+D)? \| ([0-9]+M)([0-9]+D)? \| ([0-9]+D)…” — [source](https://www.w3.org/TR/xmlschema11-2/#duration)
- Why: The changed text elevates one side of a genuine internal inconsistency in the cited authority to a normative admission table without saying so. Read via the EBNF chain duSecondFrag ::= (unsignedNoDecimalPtNumeral \| unsignedDecimalPtNumeral) 'S' and unsignedDecimalPtNumeral ::= (unsignedNoDecimalPtNumeral '.' fracFrag?) \| ('.' fracFrag), 'PT.5S' and 'PT1.S' are accepted. But the same section states that the language accepted by that production IS the intersection of three regular expressions whose seconds field is [0-9]+(\.[0-9]+)?S, which requires at least one digit before the '.' and at least one digit after it; neither 'PT.5S' nor 'PT1.S' matches. dayTimeDuration's lexical space is then d…
- Evidence: 1. git -C <integration> show 2fec061013e2162b57aea6e0e43beeaae023a30a:message/spec.md \| sed -n '1074,1108p' -> the quoted sentence and the PT.5S / PT1.S 'Valid' rows. 2. Fetch https://www.w3.org/TR/xmlschema11-2/ and read SS3.3.6.2: the three-regex characterization and its stated intersection both spell the seconds field [0-9]+(\.[0-9]+)?S. 3. Implementations diverge: a .NET probe ([System.Xml.XmlConvert]::ToTimeSpan and an XmlReader validating <xs:elemen…
- Suggested fix: Either drop the 'unsigned decimal production permits' sentence and the PT.5S / PT1.S rows and restrict fractional seconds to [0-9]+('.' [0-9]+)? 'S', or keep them and state plainly in the profile that XSD 1.1 SS3.3.6.2's EBNF and its own regex characterization disagree, and that this specification normatively chooses the permissive reading. Do not present the contested forms as settled.

**685-C-02 — medium — incomplete**
- Changed: [`endpoint/spec.md`](https://github.com/xregistry/spec/blob/2fec061013e2162b57aea6e0e43beeaae023a30a/endpoint/spec.md#L835-L846) — “Changed in this PR, endpoint/spec.md L541-L550: "- `format` : indicates the format of the CloudEvent when sent in `structured` mode. This attribute MUST NOT be specified when `mode` is `binary`. The value is the serialized envelope's media type, for example `application/cloudevents+json`, and MUST match the content type of that envelope in the protocol body (e.g. the HTTP `Content-Type` header)." Left unchanged in the same file, endpoint/spec.md L835-L844: "However, there are times when this is not possible. For e…”
- Authority: “CloudEvents JSON Event Format 1.0.2, SS3 Envelope: "Each CloudEvents event can be wholly represented as a JSON object. Such a representation MUST use the media type `application/cloudevents+json`." -- CloudEvents 1.0.2 core, datacontenttype: "Content type of `data` value. This attribute enables `data` to carry any type of content, whereby format and encoding might differ from that of the chosen event format. For example, an event rendered using the JSON envelope format might carry an XML payload in `data`, and the…” — [source](https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/formats/json-format.md#3-envelope)
- Why: This PR redefines `envelopeoptions.format` in endpoint/spec.md L541-L550 (and points message/spec.md L816-L818 at that definition) so that `format` is the serialized envelope's media type and MUST NOT appear at all in binary mode. The unchanged paragraph at L835-L846 still instructs readers to separate a JSON-serialized definition from an XML-serialized definition of the same event using `envelopemetadata.type` plus `envelopeoptions.format`. After the change that is impossible in both modes: in structured mode both definitions carry the single JSON envelope media type `application/cloudevents+json`, and in binary mode neither definition may carry a `format`. The correct post-change discrimi…
- Evidence: 1. git -C <integration> show 2fec061013e2162b57aea6e0e43beeaae023a30a:endpoint/spec.md \| sed -n '541,550p;835,846p'. 2. Construct the two definitions the paragraph describes with mode=structured: both require format=application/cloudevents+json per the new L543-L545 and per CloudEvents JSON format SS3, so `envelopeoptions.format` is identical for both. 3. Repeat with mode=binary: L542 forbids `format` on both, so the named metadata does not exist. 4. Cont…
- Suggested fix: Update endpoint/spec.md L835-L846 in this PR to name the payload data media type (Message `datacontenttype`, or `envelopemetadata.datacontenttype.value`) as the discriminating metadata, and stop citing `envelopeoptions.format` for that purpose.

**F685-A1 — medium — criterion (a), judgement**
- Added text: [`message/spec.md:1066-1123`](https://github.com/xregistry/spec/blob/2fec061013e2162b57aea6e0e43beeaae023a30a/message/spec.md#L1066-L1123) — “Values MUST contain one or more day, hour, minute, or second components, in that order, with `P` and, before time components, `T`. A leading `-` is allowed and applies to the entire value; a leading `+` or a sign on an individual component is not allowed. Year and calendar-month components are not allowed, even when zero. `M` after `T` denotes minutes, not months.\n\nOnly seconds can have a decimal fraction. The XML…”
- Already stated in: `https://www.w3.org/TR/xmlschema11-2#dayTimeDuration (linked from message/spec.md:1650 as [dayTimeDuration])` XML Schema 1.1 Part 2, dayTimeDuration… — “The specification's own opening sentence states the dependency: 'The `duration` type uses the XML Schema 1.1 `dayTimeDuration` lexical and value spaces, not an RFC3339 timestamp or an unrestricted ISO 8601 duration.' (message/spec.md:1066-1068)”
- Why: Having declared conformance to an external standard and linked it, the text then re-derives that standard's lexical rules in prose - component set and order, sign placement, the unsigned-decimal production, the minimum-one-component rule, whitespace collapse - plus a 19-row table whose Invalid rows (`P1M`, `P0Y`, `P1W`, `+PT1S`, `P-1D`, `PT1.5H`, `PT1e3S`, `P`, `PT`, `P1DT`) are all consequences of the XSD grammar rather than xRegistry decisions. A restatement of an external normative grammar can only add a second, unversioned copy that may drift from the referent; the link already carries it…
- Link instead: `Keep the conformance sentence with its existing [dayTimeDuration] link and cut the derived grammar prose and the Invalid rows of the table, e.g. 'The `duration…`
- Keep locally: The xRegistry-specific decisions the external standard does not make: comparison by exact signed decimal seconds with 60/3600/86400 factors, the no-silent-rounding/no-substitution requirement when an implementation cannot represent a value, the statement that field-specific range constraints still apply (with the `-PT0.5S` example), the exclusion of integer-count fields such a…
- Suggested change: Reference dayTimeDuration for the grammar; keep only the comparison semantics, the precision/rounding prohibition, the field-constraint and integer-count carve-outs, and the migration note.

**F685-A2 — low — criterion (a), judgement**
- Added text: [`endpoint/spec.md:543-551`](https://github.com/xregistry/spec/blob/2fec061013e2162b57aea6e0e43beeaae023a30a/endpoint/spec.md#L543-L551) — “The Message [`datacontenttype`](../message/spec.md#datacontenttype) and the CloudEvents `datacontenttype` instead describe the event data. They need not equal `format`: a structured JSON envelope can carry XML or binary data. In binary mode the HTTP `Content-Type` describes the event data and there is no separate structured-envelope `format`.”
- Already stated in: `message/spec.md` 908-920 (same head, added by this same… — “In CloudEvents structured mode, `envelopeoptions.format` identifies the media type of the serialized envelope, for example `application/cloudevents+json`. It does not describe the nested event data and MUST NOT be substituted for the Message `datacontenttype`. ... A protocol's content type describes the bytes in that protocol's body. Under the CloudEvents [HTTP binding](...),…”
- Why: The same three-layer explanation (envelope media type vs event-data media type vs protocol Content-Type, and how binary and structured mode differ) is written out in both specifications, and each already links the other for exactly this topic - message/spec.md:816-818 sends the reader to endpoint/spec.md#cloudevents10 for `mode` and `format`, while endpoint/spec.md:547 links message/spec.md#datacontenttype. message/spec.md is the side that also carries the combination table and the worked example, so it is the natural single home.
- Link instead: `Trim endpoint/spec.md:546-551 to the endpoint-side rule plus the existing pointer, e.g. 'It MUST match the content type of that envelope in the protocol body (…`
- Keep locally: In endpoint/spec.md: that `format` MUST NOT be specified when `mode` is `binary`, and that the value is the serialized envelope's media type matching the protocol body - those are the endpoint attribute's own constraints.
- Suggested change: Keep the layering explanation (and its table/example) in message/spec.md#datacontenttype; leave endpoint/spec.md with the attribute constraint and the existing link.

### (e) PR description

**E-685-01 — Material (High) — cross-references and provenance / broken link**
- Description says: “The [ISO 8601/dayTimeDuration direction thread](https://github.com//pull/685#discussion_r4015555410) remains unresolved on this keeper. Grouping does not select a new duration grammar or resolve that feedback.”
- Actual: `curl.exe -s -o NUL -w "status=%{http_code}" -L "https://github.com//pull/685"` returns `status=404` with no redirect (final url unchanged). The referenced review comment does exist, at a different URL: `gh api repos/xregistry/spec/pulls/comments/4015555410 --jq '.id, .html_url, .path'` returns `4015555410`, `https://github.com/xregistry/spec/pull/685#discussion_r4015555410`, `message/spec.md`.
- Why it misleads: The URL omits the `xregistry/spec` owner/repository segment, leaving a doubled slash (`https://github.com//pull/685`), so it 404s. This is the body's single pointer to the design question it says a reviewer must weigh, and the stakes are real: the pinned diff does NOT leave the grammar open -- `git diff 9ecf8ea...2fec061 -- message/spec.md` replaces `- \`duration\`: RFC3339 Duration.` with `+ \`duration\`: [XML Schema dayTimeDuration][dayTimeDuration].` and adds normative text ('The `duration` type uses the XML Sc…
- Suggested fix: Replace the URL with the canonical one returned by the API: `https://github.com/xregistry/spec/pull/685#discussion_r4015555410`. While editing, consider making the caveat unambiguous, e.g. 'this PR proposes dayTimeDuration normatively; the ISO 8601 alternative raised in <thread> is not yet resolved.'

**E-685-02 — Material (Medium) — scope accuracy -- substantial actual changes the body does not mention**
- Description says: “Two bounded follow-through changes are included: the real Kafka artifact selector supports the separate tooling bundle's case-insensitive guard representation without weakening validation, and the two model descriptions are aligned with the already-proposed binding scope. Structural model optionality and constraints are unchanged.”
- Actual: `git -C "D:\git\marcschier\xregistry-consolidation-8ee5d61f-integration" --no-pager diff --stat 9ecf8ea67bc9a6128c154b4545a21164493cc71e...2fec061013e2162b57aea6e0e43beeaae023a30a` shows the PR ships far more than two bounded changes. Added files never named in the body: `endpoint/kafka_bootstrap.py` (+32) -- an executable Python validator (`validate_bootstrap_servers`, `_HOST_PORT`/`_IPV6_PORT` regexes, `ipaddress.IPv6Address`) placed INSIDE the `endpoint/` specification directory rather than under `tools/`; `endpoint/README.md` (+29) -- a new guidance document; `tools/test_issue_625_kafka_bootstrap.py` (+317);…
- Why it misleads: A reviewer reading the body expects a specification-prose proposal plus two model description strings, and will size and route the review accordingly. In fact the PR adds ~407 lines of Python and README across four new files, puts a runnable Python module into a specification directory (a placement decision that deserves explicit review), edits a checked-in sample, and regenerates eight artifacts. The body even carries commit b3ef3e4 'chore(tools): remove ungrounded Python checks from proposal 685', which reinforc…
- Suggested fix: Add one sentence to the Proposed Changes section, for example: 'Also adds `endpoint/kafka_bootstrap.py` and `endpoint/README.md`, `tools/test_issue_625_kafka_bootstrap.py` and `tools/issue637/`, applies the Contoso Kafka `bootstrap.servers`/`security.protocol` sample correction (the same hunk also appears in #723 and will need merge resolution), and regenerates the eight Endpoint/Message/CloudEvents artifacts affect…

marcschier:
### Selected feedback addressed at [`15ee5c4`](https://github.com/xregistry/spec/commit/15ee5c4a7e61a4c6d3816915fee084cbbdb70d83)

Applied the findings retained in [your edited feedback](https://github.com/xregistry/spec/pull/685#issuecomment-5697211347) with signed forward commits.

| Finding | Action taken |
|---|---|
| `685-C-01, F685-A1` | Applied your selected ISO 8601-1 direction by reference. Removed the dayTimeDuration-only grammar, disputed admission table and universal fixed-seconds comparison without defining a new local grammar. |
| `685-C-02` | Used payload datacontenttype rather than envelope format to distinguish JSON/XML payload definitions in either CloudEvents mode. |
| `F685-A2, F-FMT-1` | Kept the content-layer explanation in the Message specification, linked it from Endpoint, and wrapped the affected prose. |
| `E-685-01` | Repaired the discussion URL and the duration-direction caveat. |
| `E-685-02` | Disclosed the helper, README, test and Contoso sample scope, plus the actual twelve generated artifacts. The review's count of eight was incorrect. |

Field-specific constraints, integer-count fields and native wire encodings remain unchanged. No local duration validator or fixed-length calendar conversion was added. Historical consolidation provenance and the PR title are preserved.

Current-head hosted [verify](https://github.com/xregistry/spec/actions/runs/35121298180/job/104879456085) and [DCO](https://github.com/xregistry/spec/pull/685/checks) passed. The PR remains unmerged.

message/spec.md
marcschier: Addressed in [`15ee5c4`](https://github.com/xregistry/spec/commit/15ee5c4a7e61a4c6d3816915fee084cbbdb70d83). The duration type now references ISO 8601-1:2019 directly, following the selected direction. The dayTimeDuration-only grammar and admission table are removed; property constraints and integer-count fields remain unchanged, and calendar units are not assigned invented fixed-length equivalents.

Issue #619, opened 2026-09-14
Clarify committed-state notification publication without implying delivery guarantees


marcschier:
[HTTP atomicity](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L341-L342) and [Core rollback](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L4816-L4821) include response-generation failures. [Events](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/events.md#L67-L84) define per-interaction generation/coalescing but explicitly leave transport and subscriptions outside scope.

If a Group update queues a notification and response inlining then fails, publishing tentative state before rollback could expose a change that never committed. Staging until success is already an implementable reading; the specification does not force premature publication.

Please clarify that successful-state notifications describe committed interactions, and separately decide whether durable enqueue is part of commitment. Do not infer exactly-once delivery, durable replay, parent/child delivery order or a transaction with every consumer from one event generated. This is a profile/standards proposal, not demonstrated implementation nonconformance. It may tighten systems that currently publish tentative state.

PR #574, opened 2026-09-10
fix(schemas): correct Core metadata and generated API conformance

marcschier:
Fixes xregistry/spec#573

## Proposed Changes

Correct the released schema generator's Core metadata and generated API
contracts, independently of the federation work in xregistry/spec#571.
The existing structured-array, metadata, timestamp, navigation and Version
addressing fixes remain.

The review follow-up fixes all nine reported defects:

- **574-S01/S02:** separate metadata-only Resource POST inputs, Version
  outputs and raw document content. Ignore permitted Resource read-only
  input fields without admitting them as Version output fields.
- **574-S03/S04/S05/S06:** preserve owning namespaces and unique Avro record
  names, define structured wildcard records before referencing them, apply
  Core overlays once, and generate Registry conditionals in the root namespace.
- **574-S07/S08/S09:** retain native JSON Structure URI-reference types,
  keep Meta controls out of unmodeled Version fields, and make inlined
  model/capability slots objects rather than unconstrained values.

Head covered by the verification below:
[`2b23699`](https://github.com/xregistry/spec/commit/2b23699fc5cb754d8fec6de2abbf38fadf9d6369).
All changes are signed forward commits; the original history is retained.

## Issue721 document-contract follow-up

Commit `f76f4f24` additionally addresses the functional document/API contracts
in xregistry/spec#721 **W04-c through W04-i**. The real generator and OpenAPI
template now distinguish document metadata headers from legacy query
parameters; use `xRegistry-<effective singular>id` and the other Core scalar
headers; expose map headers per key rather than as an aggregate; and keep native
`Location`, `Content-Location` and `Content-Disposition` grammars separate from
already-encoded private header values.

The generated operations distinguish 200 updates, 201 creation and body-free
303 redirects. Metadata-only routes do not gain document redirects. Document
PUT responses retain binary as well as domain JSON content, and Resource and
Version header contexts remain distinct.

The 54 added cases in `tools/test_schema_generator_regressions.py` exercise
the real generator, validate whole OpenAPI descriptions and cover imported
Resource types. They add no mock server, header-policy simulator or prose
assertions. At `f76f4f24`, the complete tools suite passed **320** cases and
all sixteen outputs independently reproduced. `docs/RELEASE_NOTES.md` records
the related wire-contract changes. Structural schemas cannot prove native
header serialization or server concurrency behavior.

The separate Primer typo, **W04-k**, belongs to marcschier/xregistry-drafts#1.
Complementary ProblemDetails, flag and metadata-input work remains in
[...]

PR #686, opened 2026-09-15
proposal(events): define lifecycle notifications and delivery semantics

### LINKED ISSUE #605: Distinguish locally owned Versions from xref projections in event rules
[Resource creation events](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/events.md#L377-L387) require a Version-created event, and [Resource event data](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/events.md#L188-L203) requires projected epoch information when data is supplied. [Core xrefs](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L2822-L2906) permit aliases with no local Versions, including dangling targets.

Creating `alias` with `meta.xref:"/dirs/d1/files/missing"` creates no Version subject or target epoch. Omitting optional event data avoids only the epoch problem, not the unconditional Version-created obligation.

Please scope Version lifecycle notifications to locally owned entities and define Resource event data when projected target state is unavailable. Recommend omitting unavailable projected epochs rather than inventing state. Cover normal-to-xref, xref-to-normal and alias deletion consistently; deleting an alias must not announce deletion of the target's Versions. Consumers must not assume every Resource-created event implies a newly created local Version.

marcschier:
## Proposed Changes

This PR contains the **Events and notifications** updates: Owned lifecycle versus xref projection events, successful-state publication and delivery guarantees.

- **Normative correction:** Version lifecycle notifications concern actual
  locally created/deleted Versions, not invented Versions for dangling aliases.
- **Projection policy:** per-field projected-data omission and source
  notifications for target-only changes remain distinct decisions.
- **Commitment/delivery policy:** successful committed-state publication is
  distinguished from transport delivery. No durable-enqueue transaction,
  additional delivery ordering or exactly-once guarantee is inferred.

This supersedes the previous 2 PRs:

| Original PR | Source scope |
|---|---|
| #686 | fix(events): distinguish owned lifecycles from xref projections. |
| #697 | fix(events): separate committed notifications from delivery guarantees. |

Fixes #605 #619


core/events.md
```diff
@@ -187,21 +187,35 @@ When present, and serialized in JSON, the `data` MUST be of the form:
...
+  `created` or `updated` event that includes `data`, `epoch` and `meta.epoch`
+  are projected target values. Each MUST be included only if that value is
+  available through the Core-defined projection at the end of the interaction;
+  each unavailable value MUST be omitted. A value
+  MUST NOT be invented or replaced with zero or a previously stored local
+  epoch. This applies, for example, when the target is missing, inaccessible,
+  or itself a cross-reference Resource. Generating events does not require
+  additional target acquisition, recursive cross-reference resolution, or
+  rejection of a cross-reference that Core permits. Other available event
+  data MAY still be included.
```
marcschier: Edit for easier reading

marcschier:
Applied signed, forward cleanup commit [f662470](https://github.com/xregistry/spec/commit/f66247054c69a1c47a906c689f345827c06d9bc2). Removed the introduced specification-text/synthetic-policy Python and its script-only instructions. The full remaining baseline tools suite passed 67 cases.

The normative proposal, source models, samples and generated artifacts are unchanged by cleanup. This PR remains open for its d/e review; no policy decision or review-thread resolution is implied.

marcschier:
Expanded this existing PR as the **events** area keeper at [581cd31](https://github.com/xregistry/spec/commit/581cd3166a3abbb93b1d64fb88012a86abc4f331), using signed forward commits from the reviewed cleaned sources below. Original keeper history and draft state are preserved; no new PR or upstream merge was created.

|---|---|
| [xregistry/spec#686](https://github.com/xregistry/spec/pull/686) | [f662470](https://github.com/xregistry/spec/commit/f66247054c69a1c47a906c689f345827c06d9bc2) |
| [xregistry/spec#697](https://github.com/xregistry/spec/pull/697) | [75f6cbe](https://github.com/xregistry/spec/commit/75f6cbe902d9ed934bf73903deca9edd7894780e) |

marcschier:
## Review feedback at [`581cd31`](https://github.com/xregistry/spec/commit/581cd3166a3abbb93b1d64fb88012a86abc4f331)

| Criterion | Verdict |
|---|---|
| (a) Repetition a link would replace | opportunity |
| (c) Correct per spec/standard | pass with one Medium gap |
| (e) Description matches content | ACCURATE, two low findings (F5, F9) |

**Findings: 1 correctness, 2 prose, 3 description.**

### (c)/(d) Correctness

**F-686-1 — Medium — incomplete**
Cross-reference epoch exception treats `meta.epoch` as a conditionally-available projected value, but unchanged Core excludes it from a cross-reference Resource unconditionally

- Changed: [`core/events.md`](https://github.com/xregistry/spec/blob/581cd3166a3abbb93b1d64fb88012a86abc4f331/core/events.md#L256-L266) — “- For a [cross-reference Resource](spec.md#cross-referencing-resources) `created` or `updated` event that includes `data`, `epoch` and `meta.epoch` are projected target values. Each MUST be included only if that value is available through the Core-defined projection at the end of the interaction; each unavailable value MUST be omitted. A value MUST NOT be invented or replaced with zero or a previously stored local epoch. This applies, for example, when the target is missing, inaccessible, or itself a cross-referen…”
- Authority: “Note that the `meta` entity's attributes `epoch` through `defaultversionsticky` MUST be excluded if `xref` is set because those would appear as part of the target Resource's `meta` entity.” — [source](https://github.com/xregistry/spec/blob/9ecf8ea67bc9a6128c154b4545a21164493cc71e/core/spec.md#L3949-L3951)
- Why: The two projected values are not symmetric in Core. Resource-level `epoch` is projected from the target and is therefore genuinely target-availability-dependent. `meta.epoch` is not projected onto the source at all: Core removes `epoch` (through `defaultversionsticky`) from a source Resource's `meta` whenever `xref` is set, in both the API-view example and the document-view rule, precisely because those values belong to the target's own `meta` entity. The changed bullet conditions both on target availability and lists only availability causes ('the target is missing, inaccessible, or itself a cross-reference Resource'), so a resolvable target reads as the case where BOTH values are included…
- Evidence: Registry has `/dirs/d1/files/real` with `epoch` 3 and `meta.epoch` 5. A client creates alias `/dirs/d1/files/alias` with `meta.xref: "/dirs/d1/files/real"` (resolvable). Reading the changed bullet, an implementation emits: { "type": "io.xregistry.resource.created", "subject": "/dirs/d1/files/alias", "data": { "epoch": 3, "meta.epoch": 5 } } The `epoch` 3 is correct (Core projects it). The `meta.epoch` 5 is not the subject's value: GET /dirs/d1/files/alias…
- Suggested fix: State the two cases separately: Resource-level `epoch` is a projected target value included only when the projection yields it; `meta.epoch` MUST be omitted from every cross-reference Resource `created`/`updated` event, because Core excludes the `meta` epoch from a source Resource whenever `xref` is set, regardless of target availability. Optionally add a resolvable-target example alongside the existing dangling-target one.

**F686-A1 — low — criterion (a), judgement**
- Added text: [`core/events.md:332-334`](https://github.com/xregistry/spec/blob/581cd3166a3abbb93b1d64fb88012a86abc4f331/core/events.md#L332-L334) — “The normal per-interaction event selection and coalescing rules still apply, including during Group or Registry deletion.”
- Already stated in: `core/events.md` 74-84 (same head, unchanged by this PR) — “A single interaction with the Registry MAY result in multiple events; however, within the scope of one interaction (see [`xregcorrelationid`](#xregcorrelationid-extension-context-attribute)), the following constraints apply: - Only one `created`, `updated` or `deleted` event MUST be generated per `subject`. - If the interaction involved more than one of those actions, then the…”
- Why: The PR twice reminds the reader that the existing per-interaction constraints still apply - at 114-116 ('Per-interaction event selection and coalescing still apply to implicit parent creation, cascading changes, and multiple changes to the same entity') and again at 332-334 - without pointing at the constraints themselves, which sit unchanged at 74-84. Reminders that carry no rule are the cheapest thing to replace with a reference, and here the reference also tells the reader which constraints are meant.
- Link instead: `Make the second occurrence a pointer, e.g. 'The per-interaction event constraints in [Registry Events](#registry-events) still apply, including during Group or…`
- Keep locally: 'including during Group or Registry deletion' at 334, and the clause at 115-116 naming implicit parent creation, cascading changes and repeated changes to one entity - both extend the existing constraints to cases they do not mention.
- Suggested change: Replace the second reminder with a reference to the existing constraint list.

**F686-B1 — low — criterion (b), judgement**
- Added text: [`core/events.md:127-133`](https://github.com/xregistry/spec/blob/581cd3166a3abbb93b1d64fb88012a86abc4f331/core/events.md#L127-L133) — “This specification does not define transport delivery guarantees, including exactly-once delivery, durable replay, ordering between parent and child events, or a transaction with every consumer. Any such guarantees MUST be defined by an explicit profile; they MUST NOT be inferred from the number of events generated. Neither client receipt of a response nor event delivery defines the commit point here. A later enqueu…”
- Why: The new section's last two paragraphs (119-133) are built almost entirely from statements about what the specification does not require. The third sentence restates the section's opening rule (commitment, not delivery, is the boundary - already stated at 88-92 and 94-101) and the fourth is its contrapositive; the reader has to hold four negations to extract one requirement, which is that a profile must define any delivery guarantee.
- Suggested change: Compress to two sentences: 'This specification does not define transport delivery guarantees such as exactly-once delivery, durable replay, or parent/child ordering; an explicit profile MUST define any such guarantee, and it MUST NOT be inferred from the number of events generated. Delivery, client receipt and later transport failures do not determine whether an interaction committed.'

marcschier:
### Selected feedback addressed at [`76b27f7`](https://github.com/xregistry/spec/commit/76b27f773b50b6f64fad24789b5e795e0ffc492e)

Applied the findings retained in [your edited feedback](https://github.com/xregistry/spec/pull/686#issuecomment-5697212190) with signed forward commits.

| Finding | Action taken |
|---|---|
| `F-686-1` | Separated conditional projected Resource epoch from meta.epoch, which is always omitted on a cross-reference source, including when its target resolves. |
| `F686-A1` | Linked the reminder to the existing per-interaction constraints in Event Definition; the suggested registry-events anchor did not exist. |
| `F686-B1` | Simplified delivery-guarantee prose while retaining profile-defined guarantees and the commitment boundary. |

A resolvable target with Resource epoch 3 and Meta epoch 5 yields source event epoch 3, not meta.epoch 5. No extra target acquisition, recursive resolution or target Version events were introduced. PR title and description are unchanged.

Current-head hosted [verify](https://github.com/xregistry/spec/actions/runs/35121318603/job/104879527429) and [DCO](https://github.com/xregistry/spec/pull/686/checks) passed. The PR remains unmerged.

core/events.md
marcschier: Addressed in [`76b27f7`](https://github.com/xregistry/spec/commit/76b27f773b50b6f64fad24789b5e795e0ffc492e). Rewrote this passage as separate Resource epoch and Meta epoch rules. A cross-reference event includes the projected Resource epoch only when available and always omits source meta.epoch, even for a resolvable target. The neighboring ordinary-Resource rule and delivery explanation are now shorter and explicit.

Issue #620, opened 2026-09-14
Specify how header writes distinguish the string null from deletion


marcschier:
[Document metadata headers](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L1602-L1611) interpret value `null` as deletion. The [header decoding rules](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L3234-L3252) unquote and percent-decode before obtaining the attribute value.

A stored four-character string `"null"` is read as header value `null`; copying that header into an update deletes it. Ordinary quoting or percent-encoding does not establish a distinct specified literal-string spelling after decoding.

Please either document this header-mode limitation and require `$details` JSON for the string, or define an unambiguous literal representation while preserving bare-null deletion. Lossless header transport for every scalar is not explicitly guaranteed, so this is a codec/fallback choice rather than proof that an implementation violates the current rule. Any escape profile needs collision/decode-order and compatibility guidance. Core string/null and document-null semantics must remain unchanged.

Issue #617, opened 2026-09-14
Clarify optional pagination relations and count without a traversal Link


marcschier:
[Generic pagination links](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/pagination/spec.md#L48-L134) require a link when more records exist for its relation. The [HTTP binding](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/pagination/spec.md#L156-L198) explicitly makes prev/first/last optional while requiring next away from the end. Count is strongly recommended, but examples carry it only as a Link parameter.

An empty or terminal response may have no traversal Link. The text does not specify how such a response carries count, or whether omission is intended.

Please distinguish supported/offered relations from mandatory ones and choose the no-Link count behavior: an appropriate self/first link, an explicitly defined carrier, or permitted omission. Do not invent a count header by implication or label optional-link implementations nonconformant. The chosen profile can affect client termination/count handling; preserve opaque continuations, count consistency and the existing mandatory-next rule.

Issue #601, opened 2026-09-14
Define an HTTP-safe representation for metadata map-key suffixes


marcschier:
[Core map keys](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/spec.md#L853-L861) permit `:`, while [HTTP map headers](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L419-L428) append the key directly to the field name. The [encoding rules](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L3228-L3296) cover values, not names.

Valid metadata `{"labels":{"a:b":"value"}}` becomes `xRegistry-labels.a:b: value`; an HTTP parser instead reads field name `xRegistry-labels.a` and value `b: value`.

Please choose a reversible field-name-safe suffix encoding, including escaping its escape character and specifying decode order, or explicitly require a JSON metadata fallback for unrepresentable keys. Recommend reversible encoding while retaining token-safe spellings. Do not ban valid Core keys. The new codec/fallback is an interoperability policy choice, not evidence that a particular implementation is nonconformant; avoid collisions with existing literal escapes and model-extended name characters.

PR #689, opened 2026-09-15
proposal(http): define header, metadata encoding and pagination rules

### LINKED ISSUE #608: Preserve native HTTP header grammars when serializing document metadata
[contenttype](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L430-L440) maps to native Content-Type, but [generic header encoding](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L3234-L3282) does not exempt native fields. [Content-Disposition](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L1664-L1703) is separately specified as a bare Resource ID.

Encoding `multipart/mixed; boundary="a b"` as `multipart/mixed;%20boundary=%22a%20b%22` loses the MIME boundary. Likewise, `Content-Disposition: doc.txt` declares a disposition type, not a filename.

Please scope private percent-encoding/unquoting to xRegistry metadata headers. Use each native field's grammar for Content-Type, Location, Content-Location and Content-Disposition; put the Resource ID in a correctly quoted filename parameter, for example `attachment; filename="doc.txt"`. Correct the repeated rules and examples together. Standard HTTP clients must not need a private decoder, and xRegistry clients should obtain identity from the existing ID metadata header rather than the entire disposition value.

marcschier:
## Proposed Changes

This PR covers the **HTTP binding and pagination** fixes: Native header grammars, private metadata/null encoding, map-key transport and pagination links/counts.

- **Normative correction:** valid native MIME/URI field values retain their native grammar rather than undergoing private xRegistry value escaping.
- **Policy decisions:** disposition/quoting and identity conventions, metadata name encoding, literal-null fallback and pagination count/link rules remain explicit proposals.
- **Composition:** native grammar, private name decoding and private value decoding stay separate. One-pass decoding is followed by null interpretation
  and existing attribute validation. Existing mixed commits are preserved.

The pre-existing #696 table inconsistency is corrected: an empty Core `name` is rejected, not stored. This preserves the existing non-empty-name and invalid-value rules; it does not change null deletion policy.

This PR combines the original 4 PR:s

| Original PR  | Source scope |
|---|---|
| #689 | fix(http): preserve native document header grammars; history retained. |
| #696 | fix(http): require JSON metadata for literal null strings. |
| #708 | fix(http): encode metadata map keys for HTTP field names. |
| #719 | fix(pagination): clarify optional links and count omission. |

#689's native-grammar correction and its additional disposition, quoting, identity and framing choices remain separate review sections. 

Fixes #608 #620 #601 #617


marcschier:
Applied signed, forward cleanup commit [4877914](https://github.com/xregistry/spec/commit/48779146b72bf6d90febcb4cad2996e2ef12d803). Removed the introduced specification-text/synthetic-policy Python and its script-only instructions. The full remaining baseline tools suite passed 67 cases.

The normative proposal, source models, samples and generated artifacts are unchanged by cleanup. This PR remains open for its d/e review; no policy decision or review-thread resolution is implied.

marcschier:
Expanded this existing PR as the **http** area keeper at [9d5fb0b](https://github.com/xregistry/spec/commit/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386), using signed forward commits from the reviewed cleaned sources below. Original keeper history and draft state are preserved; no new PR or upstream merge was created.

|---|---|
| [xregistry/spec#689](https://github.com/xregistry/spec/pull/689) | [4877914](https://github.com/xregistry/spec/commit/48779146b72bf6d90febcb4cad2996e2ef12d803) |
| [xregistry/spec#696](https://github.com/xregistry/spec/pull/696) | [b7bbb53](https://github.com/xregistry/spec/commit/b7bbb5376295cec73a2a3d4d5928d80850bb97fe) |
| [xregistry/spec#708](https://github.com/xregistry/spec/pull/708) | [64821d8](https://github.com/xregistry/spec/commit/64821d8e7f966f00b5aa0b8b4461931ca859c593) |
| [xregistry/spec#719](https://github.com/xregistry/spec/pull/719) | [3984528](https://github.com/xregistry/spec/commit/39845280df2f416478c7749f89bca7da7d56da7c) |

marcschier:
## Review feedback at [`9d5fb0b`](https://github.com/xregistry/spec/commit/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386)

| Criterion | Verdict |
|---|---|
| (a) Repetition a link would replace | problem |
| (c) Correct per spec/standard | pass with two Medium defects |
| (e) Description matches content | ACCURATE -- best-calibrated proposal body of the five |

**Findings: 4 correctness, 4 prose, 1 description.**

### (c)/(d) Correctness

**F-689-1 — Medium — introduced (the split is new; the underlying double-encoding of URL-v…**
The native/private encoding boundary exempts only four standard fields; `<URL>`-typed `xRegistry-` headers still require a second round of percent-encoding, which breaks the unchanged 'Location MUST be the same as the self URL' and '<RESOURCE>url MUST have the same value as Location' rules

- Changed: [`core/http.md`](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/core/http.md#L3392-L3406) — “The percent-encoding and quoted-string unescaping rules below apply only to `xRegistry-` metadata header values. They MUST NOT be applied to standard HTTP fields, even when those fields carry xRegistry metadata. Standard HTTP fields MUST use their own grammars: [...] - `Location` and `Content-Location` use the URI grammars in [RFC9110, section 10.2.2](...) and [section 8.7](...), respectively. Percent escapes already present in a URI MUST NOT undergo an additional round of xRegistry encoding or decoding.”
- Authority: “- The `Location` header only appears in response to a create operation and MUST be the same as the `self` URL.” — [source](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/core/http.md#L1717-L1718)
- Why: The exemption list is closed and enumerates only `Content-Type`, `Location`, `Content-Location` and `Content-Disposition`. Every other `xRegistry-`-prefixed field - including the `<URL>`-typed attributes that the same templates declare (`xRegistry-self`, `xRegistry-metaurl`, `xRegistry-versionsurl`, `xRegistry-defaultversionurl`, `xRegistry-<RESOURCE>url`, `xRegistry-documentation`, `xRegistry-icon`) - remains under the private encoding, which still lists Percent (U+0025) as MUST-encode. Before this PR both sides of each equality were ambiguous in the same direction, so the equalities were at least self-consistent. After it, one side is defined as a native URI and the other as a privately p…
- Evidence: Registry base URL contains an escape, e.g. self = https://example.com/my%20reg/dirs/d1/files/f1, and the Resource has documentation = https://example.com/docs/my%20guide.html. A conforming document-mode 201 response is then: Location: https://example.com/my%20reg/dirs/d1/files/f1 xRegistry-self: https://example.com/my%2520reg/dirs/d1/files/f1 xRegistry-documentation: https://example.com/docs/my%2520guide.html Content-Location: https://example.com/my%20reg…
- Suggested fix: Extend the boundary to `<URL>`- and `<XID>`-typed xRegistry metadata headers: state that their values are serialized using the URI grammar of RFC3986 with existing percent escapes preserved and no additional xRegistry round, OR state explicitly that they ARE privately encoded and that the 'same value' rules compare decoded values, not field values. Either way, say which, and add one URL-bearing line to the worked example block at head L3424-L3431.

**F-689-2 — Medium — pre-existing; incomplete relative to this PR's own modernization of t…**
`Expires` bullet still mandates RFC 3339 formatting behind a link to obsoleted RFC 7234, contradicting RFC 9111 and the spec's own unchanged Example 2

- Changed: [`pagination/spec.md`](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/pagination/spec.md#L185-L187) — “- The response MAY include the `expires` attribute in any response as an HTTP "Expires" header. If present, it MUST adhere to the format specified in [RFC3339](https://tools.ietf.org/html/rfc7234#section-5.3).”
- Authority: “The Expires field value is an HTTP-date timestamp, as defined in Section 5.6.7 of [HTTP]. Expires = HTTP-date For example Expires: Thu, 01 Dec 1994 16:00:00 GMT A cache recipient MUST interpret invalid date formats, especially the value "0", as representing a time in the past (i.e., "already expired").” — [source](https://www.rfc-editor.org/rfc/rfc9111#section-5.3)
- Why: RFC 9111 section 5.3 defines `Expires = HTTP-date`, i.e. the IMF-fixdate / obsolete formats of RFC 9110 section 5.6.7 - not RFC 3339. The bullet is doubly wrong: the link text names RFC 3339 while the href points at RFC 7234 section 5.3, which RFC 9111 obsoletes. The spec's own unchanged Example 2 emits an HTTP-date, so the normative bullet contradicts the adjacent example. This survives in a bullet list the PR otherwise rewrote, one line above the RFC5988->RFC8288 modernization the PR did perform.
- Evidence: An implementation that follows the bullet literally emits `Expires: 2021-12-01T16:00:00Z`. Per RFC 9111 section 5.3 that is an invalid date format, so every cache recipient MUST treat the page continuation as already expired - the opposite of the attribute's stated purpose ('Indicates when the complete set of records referenced by the `link` will no longer be available'). The unchanged Example 2 at head pagination/spec.md L190 shows `Expires: Thu, 01 Dec…
- Suggested fix: Replace with: 'If present, it MUST be an HTTP-date as defined in [RFC9111, section 5.3](https://www.rfc-editor.org/rfc/rfc9111#section-5.3) / [RFC9110, section 5.6.7](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.7).' Note the generic `expires` attribute is typed `Timestamp` (RFC3339-ish) in the protocol-agnostic section, so the HTTP binding should state the format conversion explicitly rather than reuse the generic type's format.

**F-689-3 — Low — introduced**
Serialization templates pin the disposition type to `attachment` inside 'MUST adhere to this form' blocks while the normative prose only requires 'a disposition type'

- Changed: [`core/http.md`](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/core/http.md#L1704-L1722) — “Content-Disposition: attachment; filename="<RID>" ? [...] - `Content-Disposition` SHOULD be present and, if present, MUST carry the `<RESOURCE>id` in its `filename` parameter as specified in [HTTP Header Values](#http-header-values).”
- Authority: “It MUST contain a disposition type and a `filename` parameter whose value is the Resource's `<RESOURCE>id`, serialized as a quoted-string.” — [source](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/core/http.md#L3409-L3411)
- Why: Seven templates were changed from `Content-Disposition: <STRING> ?` to the literal `Content-Disposition: attachment; filename="<RID>" ?`, and each sits under a preamble of the form 'it MUST adhere to this form' / 'MUST be of the form'. That normatively pins the disposition type, while the section the templates defer to only requires 'a disposition type'. The two normative statements disagree about whether `inline` is permitted. RFC 6266 section 4.2 gives the choice real consequences ('If the disposition type matches "attachment" ... this indicates that the recipient should prompt the user to save the response locally, rather than process it normally (as per its media type)').
- Evidence: A server emits `Content-Disposition: inline; filename="msg1"` on GET /endpoints/ep1/messages/msg1. It satisfies head L3409-L3411 (a disposition type plus a quoted `filename` carrying the RID) and RFC 6266 section 4.1, but does not match the template at head L2183, which a validator reading 'MUST adhere to this form' would flag.
- Suggested fix: Either write the templates as `Content-Disposition: <DISPOSITION-TYPE>; filename="<RID>" ?` and keep `attachment` only in the concrete examples, or state in the HTTP Header Values bullet that the disposition type MUST be `attachment` and say why.

**F-689-4 — Low — pre-existing; incomplete relative to this PR's modernization of the s…**
`rel` section still normatively cites obsoleted RFC 5988 sections while the HTTP binding one section later now cites RFC 8288

- Changed: [`pagination/spec.md`](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/pagination/spec.md#L95-L116) — “This attribute adheres to the Relation Type as defined in section 5.3 of [RFC5988](https://tools.ietf.org/html/rfc5988). [...] - MUST be a string as defined by `relation-types` in [RFC5988](https://tools.ietf.org/html/rfc5988).”
- Authority: “Additionally, Links MUST appear in the HTTP response as HTTP headers using the format described in [RFC8288](https://www.rfc-editor.org/rfc/rfc8288#section-3).” — [source](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/pagination/spec.md#L193-L194)
- Why: The document now cites two different, mutually obsoleting specifications for the same Link serialization: RFC 5988 sections 5.3 and 6.2.2 for relation types, and RFC 8288 section 3 for the header format. A reader resolving the `rel` grammar lands on an obsoleted RFC whose section numbers no longer map.
- Evidence: A reader following head L98-L99 to RFC 5988 section 5.3 to check whether `count` may sit beside `rel` on the same link-value gets the obsoleted text, while head L194 sends them to RFC 8288 section 3, whose `link-param = token BWS [ "=" BWS ( token / quoted-string ) ]` is the rule the PR's own new `count` paragraph depends on.
- Suggested fix: Retarget the `rel` citations to RFC 8288 section 3.3 (relation-type ABNF) and to the IANA Link Relations registry, matching the binding's RFC 8288 reference.

**F689-A1 — medium — criterion (a), judgement**
- Added text: [`pagination/spec.md:196-199`](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/pagination/spec.md#L196-L199) — “In this binding, `next` is the only relation REQUIRED when more records remain. The `prev`, `first` and `last` relations remain OPTIONAL subject to the boundary rules above. No `self`, `first` or `last` Link is REQUIRED solely to carry a count, and no separate HTTP count header is defined.”
- Already stated in: `pagination/spec.md` 173-184 (same head; bullets pre-existin… — “- If the response refers to the start of the set of records, then the `prev` Link MUST NOT be included in the response. - If the response does not include the start of the set of records, then the `prev` Link MAY be included in the response. - The response MAY include the `first` Link in any response. ... - If no more records remain after the current subset, then the `next` Li…”
- Why: The first two sentences restate the bullet list 12 lines above - which this PR just reworded - and that list in turn restates the protocol-independent constraints the PR added at lines 77-84. The same presence matrix is therefore carried three times in one short specification; the third copy even says 'subject to the boundary rules above', acknowledging that it adds no rule. Only the third sentence (no Link is required solely to carry a count; no count header) states something the earlier text does not.
- Link instead: `Delete the first two sentences and keep the paragraph as the count-transport clarification, e.g. 'No `self`, `first` or `last` Link is REQUIRED solely to carry…`
- Keep locally: The third sentence: that no Link relation exists purely as a count carrier, and that this binding defines no count header. That is binding-specific and is not implied by the generic constraints.
- Suggested change: Drop the two sentences that repeat the presence matrix; keep only the count-carrier clarification plus a pointer to #server-attributes.

**F689-B1 — medium — criterion (b), judgement**
- Added text: [`core/http.md:3378-3388`](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/core/http.md#L3378-L3388) — “A server unable to perform this check MUST reject a header-mode map update rather than silently accept ambiguous data. Clients can use JSON metadata instead.\n\nThis profile replaces literal map-key suffix construction. Deployments using older peers MUST coordinate support for the encoding before exchanging names that require it, or use the JSON metadata view via `$details`. Receivers MUST NOT guess between literal…”
- Already stated in: `core/http.md` 1625 and 1639-1641 (same head) — “There is no separate header spelling for assigning that literal string. ... To assign the four-character string `null` to a string attribute, clients MUST use the JSON metadata view, via the `$details` URL suffix, rather than a metadata header.”
- Why: The same escape hatch - 'when headers cannot express it, use the JSON metadata view' - is stated three times within eleven lines (3379, 3383, 3386-3387) and twice more earlier in the file. The third occurrence carries no new condition, so the paragraph's actual new content (no guessing between literal and encoded forms; no lossy truncation) is diluted. The earlier null-header passage shows the same pattern: 1625, 1639-1641 and 1656-1657 each tell the reader that assigning the literal string requires JSON metadata.
- Keep locally: The MUST NOT-guess rule, the MUST NOT-truncate rule, the older-peer coordination requirement and the closing statement that standard HTTP field names and field-value grammars are unaffected.
- Suggested change: Keep one JSON-metadata fallback sentence per passage. In the migration paragraph delete 'JSON metadata is the fallback when the header transport cannot carry the complete names.' (already said at 3383) and in the duplicate-key paragraph delete 'Clients can use JSON metadata instead.'; in the null passage keep only the explicit instruction at 1639-1641.

**F689-A2 — low — criterion (a), judgement**
- Added text: [`pagination/spec.md:274-274`](https://github.com/xregistry/spec/blob/9d5fb0b2f4aac5eb633b7900ad6c810d0a126386/pagination/spec.md#L274-L274) — “Omitting a count does not change any count already supplied for the same set.”
- Already stated in: `pagination/spec.md` 140-142 and 203-204 (same head) — “- When present, MUST be consistent with every other supplied `count` in `link` attributes and messages for that particular set of records. Omission in an intervening message does not reset an earlier count. ... All supplied counts for the same record set MUST agree, whether they occur on different Link values, different header lines or later pages.”
- Why: One rule, three statements: the `count` constraint (140-142), the HTTP binding paragraph (203-204) and the iteration guidance (274). The third adds nothing that the first two do not already say and is not binding-specific; the second is the binding-specific one (Link parameter syntax, header lines) and is the natural canonical home for the HTTP restatement.
- Link instead: `Delete line 274 and rely on [`count`](#count) plus the binding paragraph; if a reminder is wanted in the iteration section, use 'see [`count`](#count)'.`
- Keep locally: Lines 272-273 ('Absence of `count`, or an empty current page, MUST NOT be interpreted as the end when a `next` Link is present') state a client-side inference rule that the `count` constraints do not state, and should stay.
- Suggested change: Remove the third statement of the count-consistency rule at line 274.

marcschier:
### Selected feedback addressed at [`b0a080c`](https://github.com/xregistry/spec/commit/b0a080c80737eecfb317070abca7a9dd4588f445)

Applied the findings retained in [your edited feedback](https://github.com/xregistry/spec/pull/689#issuecomment-5697213108) with signed forward commits.

| Finding | Action taken |
|---|---|
| `F-689-1` | Kept private encoding for prefixed URL/XID headers and made equality rules compare decoded metadata values. Added the escaped-URI example without decoding URI escapes twice. |
| `F-689-2, F-689-4` | Specified Timestamp-to-HTTP-date conversion for Expires and updated the HTTP-date/Link relation references. |
| `F-689-3` | Used a disposition-type placeholder in normative templates; attachment remains an example, not a new mandatory policy. |
| `F689-A1, F689-B1, F689-A2, F-FMT-1` | Removed the selected pagination/count/fallback repetitions while preserving their constraints, and wrapped the affected prose. |

The private wire encoding is unchanged: its decoding turns %2520 into the URI escape %20, not a space. The PR title and description are unchanged.

Current-head hosted [verify](https://github.com/xregistry/spec/actions/runs/35121339850/job/104879597908) and [DCO](https://github.com/xregistry/spec/pull/689/checks) passed. The PR remains unmerged.

PR #723, opened 2026-09-15
fix(spec): align examples and models with existing contracts

marcschier:
## Proposed Changes

This PR reconciles examples, descriptions, the Endpoint model and checked-in artifacts where existing requirements already settle the answer. It does not choose between competing new policies.

Corrections cover model admission and diagnostics, Resource/Meta/Version examples, copied HTTP behavior, event fixtures, format/envelope metadata, MQTT names, Endpoint usage and a real Contoso Kafka sample. The Endpoint
`usage` attribute becomes required as already specified, and its role values move
from the invalid array-level `enum` onto `usage.item.enum`. Endpoint and
CloudEvents artifacts are regenerated from this branch's own generator.

The combined PR is sourced from the individual PRs which were closed. 

| Source PR / discussion | Transferred scope |
|---|---|
| #663 | Transferred both existing-catalog diagnostic corrections: known non-inlineable attributes retain inline_noninlineable; malformed paths and unknown attributes use bad_inline. Omitted the source-text-only Python suite. |
| #664 | Qualified unknown-extension storage by existing model admission and applicable validation, preserving the existing server rejection boundary. Omitted the prose-only test. |
| #665 | Scoped ximportresources necessity to different Group types; instances of the same Group type already share Resource type identity. Omitted the source-text test and left independent copied literals untouched. |
| #666 | Aligned implicit parent creation with existing required/default rules that allow server-populated non-null required values. No publication or transaction policy was added; prose-only tests omitted. |
| #667 | Transferred existing unavailable-versus-failed validation diagnostic references, including format_unknown, compatibility_unknown and external-document validation guidance. Omitted the source-text suite. |
| #670 | Aligned MQTT user_properties and payload_format_indicator spellings with the formal model. Retained only the existing model field/enum/item-kind checks; removed all Markdown, table, prose and fenced-example dependencies. |
| #673 | Transferred all model-backed Schema format guidance: optional Group format and equals constraint, required Version format, matchversions and absence of an implied default. Group sketch remains optional; only the Version sketch loses its optional marker. Retained model-aspect tests only. |
| #674 | Transferred only the worked-state v1 ancestorid correction from v0 to v1 (source default-context hunk at old line 639). Excluded all three A syntax/copied-note corrections and the entire source-derived ancestry test. |
| #675 | Transferred all existing-rule non-sticky-to-sticky PATCH example corrections: recompute selection without an explicit ID, project v1 and retain the correct ancestry. No discarded-ID admission policy or synthetic transition engine was imported. |
[...]

Issue #721, opened 2026-09-15
Spec items that could use fixes/corrections from clean room implementation effort


marcschier:
# Remaining specification work

This issue consolidates the remaining specification work previously tracked in separate issues. Every checkbox represents unfinished work; closing the original tickets means superseded by this tracker, not implemented or merged.

Each item retains its original description, source link, and applicable follow-up or recorded decision so it can be worked on independently or made into a separate sub-issue later.

xregistry/spec#688, covering the relocation of specification tooling, remains separate and is not part of this checklist.

The original descriptions contain pinned baseline evidence. Check current upstream and related PRs before implementing, and do not duplicate already proposed fixes. A closed source issue does not imply its linked PR has merged.

- [ ] **W01: Clarify Core primer federation identity once federation draft specs are signed off**

  <details>
  <summary>Full description and retained context</summary>

  **Original issue:** [xregistry/spec#572](https://github.com/xregistry/spec/issues/572)

  **Original description**

  ## Deferred documentation follow-up

  Review of xregistry/spec#571 requested keeping released/non-working-draft
  documentation unchanged while the federation specification family is still
  being developed.

  Once the federation documents are ready for inclusion in the main
  specification, update the Core primer's federation/identifier discussion to:

  - Distinguish stable identity agreed by participating domains from Core Group
    IDs and XIDs, whose uniqueness is scoped to their Registry context.
  - Explain that a cross-registry selection needs a Registry context and the
    typed Group/Resource path, not merely an equal XID.
  - Preserve Core `meta.xref` as a same-Registry, same-Resource-model-type,
    one-hop reference rather than a remote endpoint URL.
  - Link the stabilized federation specification and its catalog/binding
    documentation from the primer.

  This is a future documentation task, not a proposal to change Core `xref`
  semantics or to add working-draft links to the current primer. The premature
  primer changes are being removed from xregistry/spec#571.

  **Retained context and decisions**

  - Keep this documentation follow-up deferred until the federation specification family is signed off. It does not authorize changing released Core xref semantics or adding premature working-draft links.

  </details>

- [ ] **W02: Align the published full-model reference types and their producer**

  <details>
  <summary>Full description and retained context</summary>

  **Original issue:** [xregistry/spec#593](https://github.com/xregistry/spec/issues/593)

  **Original description**

[...]

PR #709, opened 2026-09-15
fix(openapi): align flags and epoch guard carriers

marcschier:
Fixes #615

Replace the remaining legacy noepoch/nodefaultversion/noreadonly and attribute query controls with the supported `ignore` flag on write operations. Remove epoch query guards from updates and Meta reads; single-entity DELETE guards are optional and use equality, with Resource deletion checking the Meta epoch rather than the default Version epoch.

Document-view writes advertise `xRegistry-epoch` only for document-bearing Resource types, including imported types. Metadata writes carry their guards in JSON; no invented Meta-object header is added. Preserve null/absent bypass, create-time ignoring, and unsigned values. Describe `doc` as a bare presence flag, not true/false, and correct the Primer's metadata selector guidance.

OpenAPI 3.0 empty-value parameters cannot themselves mandate omission of `=`; the parameter description explicitly requires the binding's bare `?doc` wire form. Server concurrency enforcement is governed by Core; these tests verify generated carriers/schema values and source rules, not a live concurrent server.

Only parameter contracts and scoped descriptions change in the four regenerated OpenAPI files. Paths/methods, bodies, response headers and entity schemas are unchanged. This does not duplicate #574's route/meta-selector corrections, #661's operation-aware metadata schemas, or the separate administrative/document-header proposals.

**Evidence:** 26 focused cases failed on the baseline; all 26 and all 93 full tools tests pass after the fix. A mixed document/metadata-only model with imported Resource types proves correct header applicability. Actual Primer authoring checks cover 113 local links, spelling and tabs. Eight files, opened as draft.

PR #703, opened 2026-09-15
fix(generator): match ifvalues selectors case-insensitively

marcschier:
Fixes #644

Known selectors and their complementary unknown-selector fallback now use the
same case-insensitive exact guards in the JSON Schema/OpenAPI emitter.
The supplied selector spelling and unrelated values are never normalized.
Literal punctuation remains literal, final newlines do not turn exact matching
into prefix matching, native Boolean/integer selector kinds remain intact, and
multiple/nested selector conditions do not overwrite one another.

The original OpenAPI emitter only produced discriminator mappings and disconnected
components. A discriminator is case-sensitive and cannot express case variants
plus arbitrary extension selectors. This PR replaces those components/mappings
with connected inline validation branches at the actual containing schemas.
It does not globally open objects or change Core matching to a prefix rule.

This starts directly from main commit 9ecf8ea67bc9a6128c154b4545a21164493cc71e.
The same defective conditional logic was inspected at #574's published head;
none of that PR or the lossless/closure/other pending PRs is copied or stacked.
Source alias admission (#655) and strict scalar enum projection (#649) remain
separate concerns.

| Requirement | Evidence |
| --- | --- |
| Known-selector case variants receive identical valid/invalid checks | `test_all_ascii_case_variants_activate_the_same_branch` |
| Genuine unknowns remain extensions, without prefix matching | `test_unknown_and_nonexact_selectors_keep_the_extension_fallback` |
| Types and original spelling are preserved | `test_unknown_fallback_preserves_the_selector_type`, `test_native_scalar_selectors_keep_their_types` |
| Independent and nested conditions remain active | `test_multiple_independent_selectors_all_remain_effective`, `test_nested_ifvalues_apply_without_normalizing_other_values` |
| Actual published Message/Endpoint/CloudEvents behavior | `test_published_case_variant_options` |
| Connected OpenAPI semantics | `test_openapi_conditionals_are_connected_without_case_sensitive_discriminators` |

Validation used the isolated Python environment with bytecode/plugin autoload
disabled. Original production emitter: 36 failed / 15 passed; original published
validators: 8 failed. Final focused suite: 63 passed. One final full
`python -B -m pytest tools -q --tb=short -p no:cacheprovider
-p pytest_asyncio.plugin`: 130 passed (existing validator deprecation warnings).

Eight Makefile-equivalent JSON Schema/OpenAPI candidates were generated and
checked; only the six semantically changed Message, Endpoint, and CloudEvents
[...]

PR #724, opened 2026-09-15
fix(tools): correct schema generation and OpenAPI contracts

### LINKED ISSUE #616: Align generated OpenAPI administrative operations with current HTTP APIs
At [generator comparison head 36af874](https://github.com/xregistry/spec/blob/36af874b748665262fcbbfb0fed82a0866fcf235/core/templates/xregistry_openapi_template.json#L157-L397), effective OpenAPI still advertises PUT/PATCH `/model`, offered capabilities through `?offered`, and export using `model`. [Current HTTP](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L843-L1090) instead uses `/capabilitiesoffered`, read-only `/model`, and writable `/modelsource`.

A generated `putRegistryModel` client therefore targets the wrong administrative API. The capability component also lacks the current available-map declaration; [export](https://github.com/xregistry/spec/blob/16483bb564586423de9c128db89f3b61d360f4e3/core/http.md#L755-L773) expands to `*,capabilities,modelsource`.

Please correct administrative paths/methods, offered/current capability schemas and the export alias, then regenerate only affected OpenAPI projections. Do not require optional APIs merely because the template describes them. This changes stale generated client contracts to the current binding and is independent of the Group/Version route work in xregistry/spec#573 and xregistry/spec#574.

marcschier:
## Proposed Changes

This composes fixes in the real schema generator and OpenAPI template, with
actual model-to-artifact checks and regenerated outputs. Source specification
Markdown and source models are unchanged.

The combined implementation fixes closed-object admission, dynamic and typed
wildcard maps, case-insensitive conditional selectors, local includes,
request/completed-response and Meta schemas, administration operations,
ProblemDetails, flag/guard carriers, and independent decimal/epoch mappings.
All 16 checked-in schema outputs are generated from these sources.

Changes are combined from the individual PR's now closed:

| Source PR / discussion | Transferred scope |
|---|---|
| #656 | Only independent decimal JSON-number mapping and nonnegative integer epoch constraints transferred, with actual-generator regressions; the lossless scalar profile stays in its original decision PR. |
| #658 | Model naming, nonmutation, root attributes and closed-object validation with conditional wildcard admission. Added group-less root and late-request-member composition regressions. |
| #659 | JSON Structure dynamic primitive/object/nested maps and closed-empty attribute maps with typed wildcard handling. |
| #661 | Operation-specific input/completed-response schemas, custom Meta identities, aliases, required/default/readonly obligations and dynamic modelsource requests. Full properties precede request closure. |
| #662 | Typed JSON Structure wildcard extras, including nested values and closed-empty exceptions, without importing a scalar profile. |
| #683 | Restored the required member list in the OpenAPI ProblemDetails schema; kept structural artifact validation and enabled format-nongpl validation dependencies. |
| #690 | Local includes resolution before all output dialects, preserving caller models and root attributes; no profile-only exact-number parser imported. |
| #700 | modelsource/capabilities administration paths and export parameters with operation-specific request schemas. Prose assertions removed. |
| #703 | Case-insensitive conditional selector guards with wildcard closure and request/response roles. Fixture controls reflect model-admitted extensions and existing required/default response fields. |
| #709 | Structural flag carriers, ignore-on-write operations and modeled deletion guards. Primer content routed only to B; all prose/Primer assertions removed. |

#656 remains open as the expanded Schema/projection keeper, including its still-unapproved scalar profile, native-root and selector proposals. Required prerequisite overlap is preserved until this bundle
actually lands. The Primer changes from #709 are in the primer PR, not this one. #574 remains separate with overlapping Core metadata and generated-API work; none of its unmerged code is adopted here.

Fixes #616 #618 #621 #624 #644 #647 #648 #660
Refs #615, #645.

## Issue721 and scalar-item follow-up scope

This PR also addresses xregistry/spec#721 **W03** and **W14**:
`core/model.schema.json` and `tools/validate-models.py` distinguish source-model
admission from expanded-model completion; the generator handles scalar enum
constraints, defaults and conditional selectors without conflating enum value
equality with case-insensitive selector activation.

Commits `543d279` and `7e67818` implement the additional scalar `item.enum` and
`item.strict` rule proposed in xregistry/spec#677. JSON Schema, OpenAPI and
JSON Structure constrain array elements and map values, including nested
containers. Avro emits a real named enum for legal string symbols, shares
identical definitions, deduplicates equal symbols and gives differing sets
distinct names. A non-string enum or a string outside Avro's symbol grammar
has no equivalent Avro restriction and retains the plain mapped type; the
other dialects retain the restriction. This adds no codec or server runtime.
See `tools/item-enum-projection.md` for the precise limits.

The source and emitter checks reject item enum/strict aspects on non-scalars.
Advisory enum values must still have the declared item kind. Array-level enums
are **not Core-valid**: this standalone branch still retains its old array
projection and deferred source applicability check until the illegal source
declaration is replaced in xregistry/spec#723. That compatibility debt is not
an endorsement of array-level enum semantics.

Other prerequisites remain explicit: actual CloudEvents composite expansion
needs the corrected `#/groups` pointers in marcschier/xregistry-drafts#2.
The input schema retains current Core's 63-character Group singular limit;
the separate 61-character proposal in xregistry/spec#677 is not adopted.

At `7e678180`, the full tools suite passed **1,096** cases, including 142 new
item-enum cases. All sixteen exact generation commands reproduced the checked-in
LF bytes independently without newline normalization; no source-model or
generated-artifact change was needed on this branch for the new aspect.
[Hosted verify](https://github.com/xregistry/spec/actions/runs/35444999697)
and DCO passed for that head. These are standalone results, not a claim that
the later multi-PR composition is conflict-free or complete.

## Earlier review follow-up (historical evidence)

At [`f1f033e`](https://github.com/xregistry/spec/commit/f1f033ed0f6a5686c8e49249188243ad2f49d57d),
all seven functional findings and the guidance finding are addressed:

- **R724-C01/C02/C03:** omit empty `required` arrays, include applicable Core
  envelopes and ignored navigation fields, and use input rather than completed
  response schemas on the affected metadata writes.
- **R724-C04:** reject null collection entities while preserving legal
  attribute resets; remove the unsupported entity-deletion interpretation.
- **R724-C05/C06/C07:** retain writable conditional siblings, apply inactive-name
  wildcard fallback, and permit legitimate identity-only alias responses
  without weakening normal response requirements.
- **STD-724-01:** replace outdated typed-wildcard limitations with links to the
  implemented feature's guidance, keeping qualification limits explicit.

The new focused cases passed 65 checks; the full tools suite passed 523.
All sixteen outputs independently reproduced, and model/documentation checks
passed. For that historical head, [verify](https://github.com/xregistry/spec/actions/runs/35086398590)
and [DCO](https://github.com/xregistry/spec/pull/724/checks) passed.
The earlier unpublished four-PR composition passed 696 cases with source-level merge
resolutions and explicit fixture adaptations; standalone passing results do
not imply conflict-free integration.


marcschier:
Addressed all **8** findings for this PR at [f1f033e](https://github.com/xregistry/spec/commit/f1f033ed0f6a5686c8e49249188243ad2f49d57d) using signed forward commits.

| Finding | Verification |
|---|---|
| R724-C01 | `test_filtered_input_requirements_produce_valid_openapi_documents`, `test_version_identity_filter_drops_empty_required_without_losing_client_fields` |
| R724-C02 | `test_published_registry_get_to_write_roundtrip_keeps_core_navigation`, `test_closed_version_contenttype_is_core_input_not_an_extension`, `test_core_request_inventory_keeps_nested_fields_and_rejects_unknowns`, `test_core_readonly_navigation_is_ignored_regardless_of_supplied_kind`, `test_published_root_schema_hint_is_a_message_member`, `test_schema_hints_follow_single_entity_message_boundaries_not_nested_objects`, `test_cli_generates_the_selected_model_in_all_formats` |
| R724-C03 | `test_existing_metadata_write_routes_use_inputs_not_defaulted_responses`, `test_document_bearing_raw_requests_are_not_rewired_as_metadata_inputs` |
| R724-C04 | `test_patch_collection_null_entries_are_errors_but_attribute_resets_remain_valid`, `test_published_patch_rejects_null_group_entries_without_rejecting_objects` |
| R724-C05 | `test_readonly_selectors_keep_writable_conditional_siblings`, `test_unknown_server_selectors_admit_conditional_and_wildcard_alternatives` |
| R724-C06 | `test_request_conditional_names_select_active_or_inactive_wildcard_contract`, `test_unknown_server_selectors_admit_conditional_and_wildcard_alternatives`, `test_request_conditional_wildcards_remain_guarded_for_known_selectors`, `test_colliding_request_conditions_preserve_known_constraints_and_unknown_alternatives`, `test_known_conditional_wildcards_are_not_widened_by_unknown_or_colliding_selectors` |
| R724-C07 | `test_identity_only_resource_aliases_do_not_require_target_defaults`, `test_alias_response_does_not_weaken_incomplete_expanded_response_validation`, `test_alias_meta_uses_the_same_model_specific_definition_nested_and_standalone` |
| STD-724-01 | Current guidance cross-references and existing authoring checks; no prose assertion |

Current-head [verify](https://github.com/xregistry/spec/actions/runs/35086398590/job/104762116076) and [DCO](https://github.com/xregistry/spec/pull/724/checks) passed. The corrected four-PR integration passed 696 cases and independently reproduced all sixteen outputs; integration-specific source/fixture resolutions remain explicit, not a claim of conflict-free merging.

Updated the description with current corrections and evidence. Concurrent description simplifications and ready-for-review transitions were preserved; no PR was merged or closed and no source branch was deleted.

marcschier:
## Review feedback at [`f1f033e`](https://github.com/xregistry/spec/commit/f1f033ed0f6a5686c8e49249188243ad2f49d57d)

| Criterion | Verdict |
|---|---|
| (a) Repetition a link would replace | pass |
| (c) Correct per spec/standard | changes requested |
| (e) Description matches content | ACCURATE on all measurable claims; one low internal contradiction (F6) |

**Findings: 1 correctness, 2 prose, 2 description.**

### (c)/(d) Correctness

**R9-724-C01 — Medium — incomplete**
Document-Resource PUT/POST JSON bodies still use the completed-response schema, rejecting Core-valid metadata-view writes

- Changed: [`tools/schema-generator.py`](https://github.com/xregistry/spec/blob/f1f033ed0f6a5686c8e49249188243ad2f49d57d/tools/schema-generator.py#L440-L445) — “resource_path = f"{group_path}/{plural}/{{resourceid}}" if not resource.get("hasdocument", True): for method in ("put", "post"): openapi["paths"][resource_path][method]["requestBody"]["content"]["application/json"]["schema"] = { "$ref": f"#/components/schemas/{name}WriteInput" }”
- Authority: “core/http.md (pinned baseline, lines 2041-2053): "The `$details` suffix controls whether the request is directed to the Resource metadata or to the Resource domain-specific document. ... When directed to the Resource metadata, the request MUST be of the form:\n\n```yaml\nPATCH /<GROUPS>/<GID>/<RESOURCES>/<RID>[$details]\nContent-Type: application/json; charset=utf-8\nor\nPUT /<GROUPS>/<GID>/<RESOURCES>/<RID>[$details]\nContent-Type: application/json; charset=utf-8\n\n{ ... Resource entity excluded for brevity ...…” — [source](https://github.com/xregistry/spec/blob/9ecf8ea67bc9a6128c154b4545a21164493cc71e/core/http.md#L2041-L2053)
- Why: This PR introduces the whole request/response role separation (`<name>WriteInput`, `<name>PatchInput`, `<name>Meta{Write,Patch}Input`, `Registry{Write,Patch}Input`) and rewires every write body it touches: `PUT\|PATCH /`, `PUT /<GROUPS>/{groupid}`, `PUT\|PATCH /.../{resourceid}/meta`, `POST /.../{resourceid}/versions`, and the nested collection maps. The rewire for the direct Resource path is gated on `if not resource.get("hasdocument", True)`, so Resource types that keep the Core default `hasdocument: true` keep the *completed response* schema as their JSON request body. At this head that response schema is strictly stronger than before: it now carries `anyOf[0] = {"oneOf": [{"required": ["v…
- Evidence: {'environment': 'D:\\git\\marcschier\\xregistry-consolidation-8ee5d61f-venv\\Scripts\\python.exe, PYTHONDONTWRITEBYTECODE=1, openapi-schema-validator OAS30WriteValidator, temporary sandbox containing only pinned head blobs', 'validatedAgainst': 'schema/schemas/openapi.json at head, $ref taken from /paths/~1schemagroups~1{groupid}~1schemas~1{resourceid}/put/requestBody/content/application~1json/schema (resolves to #/components/schemas/schema)', 'invalidlyR…
- Evidence: Not a runtime regression: at 9ecf8ea the same two operations were also bound to `#/components/schemas/schema`, and the four probe payloads above were already rejected there. What changed is that this PR (a) built the correct input schemas, (b) rewired every other write body, and (c) added `hasdocument`-aware handling at exactly this call site, while leaving the `hasdocument: true` case on the response schema. Scope check across all four published artifact…
- Suggested fix: Wire the `application/json` request body of the direct Resource `PUT`/`POST` to `{name}WriteInput` for all Resource types, independent of `hasdocument`. Keep the existing `application/octet-stream` (and any document-view) content entry for `hasdocument: true` types, since Core routes the domain-specific document through that view with `xRegistry-*` headers and the `xRegistry-epoch` guard this PR added. Consider whether the read-only `$details` path should gain the same metad…

**F724-B1 — low — criterion (b), judgement**
- Added text: [`README.md:71-87`](https://github.com/xregistry/spec/blob/f1f033ed0f6a5686c8e49249188243ad2f49d57d/README.md#L71-L87) — “JSON Schema and OpenAPI projections preserve model-defined object closure. Known Core metadata and active conditional members remain admitted; explicit wildcards and `any` retain their openness. Collection and map keys remain open while their modeled object values are validated independently. ... JSON Structure uses maps for wildcard-only dynamic objects and zero-entry maps with explicit validation-feature support f…”
- Already stated in: `tools/json-structure-dynamic-maps.md` tools/json-structure-dynamic-maps.md:7-… — “The generator projects a Core object with only a wildcard definition as a JSON Structure `map`, not an `object` with empty `properties`. The dialect requires at least one named property for a structured object. ... A closed object with no named properties is not an open dynamic map. Its precise representation is an empty map”
- Why: The first paragraph states generator-internal behaviour in generator-internal terms and, unlike the two paragraphs that follow it, offers no link to the document that explains it. 'Zero-entry maps with explicit validation-feature support for closed-empty objects' compresses a rule that tools/json-structure-dynamic-maps.md needs a section and two JSON examples to convey. The README's function here is to tell a reader that generated types changed and where to read about it; the second and third paragraphs already do that correctly by ending in 'See the ... guidance'.
- Link instead: `tools/json-structure-dynamic-maps.md (already linked from the third paragraph; the first paragraph should link tools/openapi-request-meta-schemas.md or be fold…`
- Keep locally: The fact that generated request/response and JSON Structure types changed and that consumers should regenerate - stated plainly.
- Suggested change: Reduce the first paragraph to one plain sentence about what changed for consumers and let the existing 'See the ... guidance' links carry the detail, as the other two paragraphs already do.

marcschier:
### Selected feedback addressed at [`cb6ea8e`](https://github.com/xregistry/spec/commit/cb6ea8e95b294199c2ef903f84b06bca1a6d397f)

Applied the findings retained in [your edited feedback](https://github.com/xregistry/spec/pull/724#issuecomment-5697216706) with signed forward commits.

| Finding | Action taken |
|---|---|
| `R9-724-C01` | Applied the approved route-aware correction: metadata input schemas on the appropriate routes, Resource $details PUT/POST operations, and distinct Resource PUT versus Version-creating POST semantics. |
| `F724-B1, F-FMT-1` | Simplified the README explanation, clarified the route/media guidance and wrapped the two explicitly identified guide lines. |

The original recommendation misread [$details]: for a document-bearing Resource, the suffix selects metadata, not application/json. Applying that recommendation literally would undo the raw-document fix. Bare JSON/binary content, response constraints and ignored read-only input semantics are preserved. No unrelated route expansion or PR metadata edits.

**Behavioral evidence:** `test_details_routes_expose_metadata_write_operations`, `test_details_put_uses_resource_metadata_input`, `test_details_post_creates_a_version_with_core_ignored_resource_values`, `test_details_post_returns_version_fields_not_resource_navigation`, `test_bare_document_routes_carry_domain_documents_not_metadata`.

Current-head hosted [verify](https://github.com/xregistry/spec/actions/runs/35121394358/job/104879789904) and [DCO](https://github.com/xregistry/spec/pull/724/checks) passed. The PR remains unmerged.

tools/README-spec-644-ifvalues.md
marcschier, line 1: Combine all markdown files that were added in tools into 1, add TOC at the beginning.  Combine all test python scripts added in tools into 1.

tools/requirements.txt
```diff
@@ -9,5 +9,5 @@ jsonpointer
 pytest
 pyyaml
 avro
-jsonschema
+jsonschema[format-nongpl]
```
marcschier, line 15: Explain why?

tools/README-spec-644-ifvalues.md
marcschier: Done in 62fb982.

**Markdown.** The seven notes this PR added are now one file,
`tools/schema-generator.md`, with a contents list at the top:

- Running the tests
- Model source schema: what each validation stage proves
- Local model includes
- Conditional selector matching
- Scalar item value sets
- JSON Structure dynamic object projection
- Typed wildcards on named JSON Structure objects
- OpenAPI request roles and Resource Meta schemas

The three per-file "run these tests" blocks collapsed into one section, the
links between the notes became anchors, and the four `README.md` links now point
at the corresponding sections.

Combining them exposed a contradiction worth flagging: the OpenAPI notes still
said "Array-level `enum` keeps its existing item projection, and no `item.enum`
vocabulary is introduced", while the item notes describe exactly that vocabulary
and the generator now rejects an `enum` on the owning array or map. Corrected to
match the code.

**Tests.** The twelve added modules are now
`tools/test_schema_generator_projections.py`.

This one needed care rather than concatenation: the modules shared 25 top-level
names, and several of them were fixtures with the same name and different bodies
(`document`, `model`, `accepts`, `rejects`, `generator`, `entry_meta`,
`base_document`, `generated_validator` and more). Appending the files would have
let a later fixture silently rebind an earlier one, and tests would have kept
passing against the wrong fixture. So the merge:

- emits the repository/generator loader preamble once instead of twelve times,
- shares the definitions that were byte-identical everywhere,
- prefixes the remaining colliding names per section, including the fixture
  parameters that reference them, and demotes a shared definition back to a
  renamed one if anything it reads is renamed,
- rewrites names by token position, so comments and formatting are unchanged.

Verified by count rather than by inspection: the twelve files collected
31 + 12 + 194 + 76 + 21 + 144 + 36 + 68 + 79 + 39 + 170 + 234 = 1104 tests, the
merged module collects 1104, and `pytest tools/` is 1171 passed before and after.
Section headers record which file each part came from.

tools/requirements.txt
marcschier: `jsonschema` treats `format` as an annotation, not an assertion, unless the
matching validator packages are installed. Without the extra, `FormatChecker()`
has no `uri` checker at all, so a `format: uri` field accepts anything:

```python
>>> c = jsonschema.FormatChecker(); "uri" in c.checkers
False
>>> v = jsonschema.Draft7Validator({"type": "string", "format": "uri"}, format_checker=c)
>>> list(v.iter_errors("#/categories/x"))
[]
```

That matters because the Problem Details tests assert that the RFC 9457 `type`
and `instance` members reject a non-URI, and several OpenAPI checks run through
`OAS30Validator.FORMAT_CHECKER`. Without the extra those assertions could never
fail. It is not only a quality point: installing plain `jsonschema` and running
the suite gives 32 failed, 1139 passed, all of them the URI and Problem Details
cases.

`format-nongpl` rather than `format` is a licensing choice. The two extras are
otherwise the same, but plain `format` pulls in `rfc3987`, which is GPL-licensed:

```
rfc3987; extra == 'format'
rfc3986-validator>0.1.0; extra == 'format-nongpl'
rfc3987-syntax>=1.1.0; extra == 'format-nongpl'
```

I have recorded both reasons as a comment next to the requirement in 62fb982, so
the next reader does not have to ask.

PR #725, opened 2026-09-16
clarify maxversions text

duglin:


core/model.md
```diff
@@ -767,11 +767,12 @@ Note that this feature has similar results to setting the Resource attribute's
...
 - When the limit is exceeded, implementations MUST prune Versions by
-  deleting the oldest Version first (based on the Resource's
-  [`versionmode`](#groupsstringresourcesstringversionmode)
-  algorithm), skipping the Version marked as "default".
-  Once the single oldest Version is determined, delete it.
-  A special case for the pruning rules is that if `maxversions` is set to
+  deleting the [oldest Version](#groupsstringresourcesstringversionmode) first.
+  If the oldest Version is the current default Version, then for the purposes
+  of determining the "next oldest Version", the default Version is removed from
+  the collection of Versions, then the list of "root" ancestors is recalculated
```
jskeet: I'm not sure what the difference is between this and just "skipping the Version marked as default". If there's some subtlety here about what the list of root ancestors is, etc, might it be worth giving a concrete example?

duglin: It wasn't 100% sure to me that everyone knew what to do in all cases. It could mean "skip that default-version/root and only look at the other roots" or it could mean "pretend that root is already gone, then what would you do?".  This tries to make it clear... pretend it's gone.

duglin:
Sample tests from our chats to help explain the proper behavior:
image

core/model.md
jskeet: I think this is another case where not being sufficiently into xReg means I'm missing the importance of roots (and ancestor trees, it seems). Will try to understand your sample tests... but regardless, I'm sure this change is fine to take.

duglin:
Conditionally approved pending minor wording tweak suggest by @jskeet

core/model.md
```diff
@@ -843,16 +844,16 @@ Note that this feature has similar results to setting the Resource attribute's
...
-    are not referenced as an `ancestor` of another Version and choosing the
-    one with the newest `createdat` timestamp. If there is more than one, then
-    the one with the highest alphabetically case-insensitive `versionid` value
-    MUST be chosen.
-  - The "oldest" Version MUST be determined by finding all root Versions (ones
-    that have an `ancestorid` value that points to itself), and then choosing
-    the one with the oldest `createdat` timestamp. If there is more than one,
-    then the one with the lowest alphabetically case-insensitive `versionid`
-    value MUST be chosen.
+  - The "newest" Version MUST be determined by finding all "leaf" Versions
```
jskeet, line 847: I'd personally just make "newest leaf Version" and "oldest root Version" terms that are defined here used throughout - but even without any further changes, this is definitely clearer.

duglin: I think the problem with using "newest leaf Version", at least here, is that "leaf" becomes a (sort of) qualifier and someone could then ask "ok, but how do I find the newest non-leaf Version? or do I ever need to?". By leaving "leaf" out of it I think it makes it clear that there's only one "newest" concept not (potentially) several - even though "leaf" is part of the algorithm. But I'll think more...

lailabougria (approved):

